PR #128, opened 2014-11-11
Fix errors introduced by adding runtime_Semacquire.

andybalholm:
I fixed the panics by removing unnecessary locks (unnecessary since JS is single-threaded).


neelance:
I finally found the time to take care of it. Your other pull request is now (manually) merged and this one is not required any more.

PR #153, opened 2015-01-11
use top level this as global object

tenntenn:
I want to use gopherjs for [JavaFX](http://docs.oracle.com/javase/8/javase-clienttechnologies.htm) with [Nashorn](http://www.oracle.com/technetwork/articles/java/jf14-nashorn-2126515.html) such as a following code.
[My article](http://qiita.com/tenntenn/items/e61822b265bc2b8f6c88)  explains this code.
(Sorry! This article is written in Japanese.)

* https://gist.github.com/tenntenn/f8c89022f7139a3c15bf

In Nashorn, we can get the global object from only `this` in top level.
But gopherjs cannot get top level `this`.
So I fix it.

Please review and merge my pull request.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -105,7 +105,7 @@ func WriteProgramCode(pkgs []*Archive, w *SourceMapFilter) error {
 		}
 	}
 
-	if _, err := w.Write([]byte("\"use strict\";\n(function() {\n\n")); err != nil {
+	if _, err := w.Write([]byte("\"use strict\";\n(function($top_level_this) {\n\n")); err != nil {
 		return err
 	}
 	if _, err := w.Write(removeWhitespace([]byte(prelude.Prelude), minify)); err != nil {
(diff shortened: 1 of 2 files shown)


neelance:
Please try if 635a153 works for your case.

tenntenn:
Thank you. I'll try it.
But `$module` become a global variable.

tenntenn:
Sorry. Your code does not work on Nashorn.

on Nashorn:
```shell
$ cat nashorn.js
print(this !== (function(){ return this; })());
$ jjs nashorn.js
false
```

on Node.js:
```shell
$cat node.js
console.log(this !== (function(){ return this; })());
$ node node.js
true
```

tenntenn:
Sorry. My example code is wrong.

tenntenn:
`this` in local function  does not equal global object with strict mode on Nashorn.

```shell
$ cat nashorn.js
"use strict";
print(this === (function(){ return this; })());
$ jjs nashorn.js
false
```

tenntenn:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#.22Securing.22_JavaScript

neelance:
Note to self: Don't push a bugfix quickly before leaving the house. Fixed `$module` issue by d5cd629.

neelance:
Damn, I only tested without strict mode...

According to http://stackoverflow.com/questions/3277182/how-to-get-the-global-object-in-javascript it seems to be really hard to get the global object in some homogenous way, so let's just go with your pull request.

PR #166, opened 2015-01-27
candidate for improving the source maps

iansmith:
@neelance I'm not sure if you prefer this method of determining the gopath.    There are a couple of cases this avoids: 1) if your /tmp/ is actually something /private/tmp (which it is on OS X) and 2) if you have relative (..) parts to your GOPATH.  

I also added a warning if this detects the problematic case, I don't know if you care about that.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -8,6 +8,7 @@ import (
 	"go/scanner"
 	"go/token"
 	"io/ioutil"
+	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -536,6 +537,7 @@ func (s *Session) WriteCommandPackage(pkg *PackageData, pkgObj string) error {
 			}
 			pos := fileSet.Position(originalPos)
 			file := pos.Filename
+
 			switch hasGopathPrefix, prefixLen := hasGopathPrefix(file, s.options.GOPATH); {
 			case hasGopathPrefix:
 				file = filepath.ToSlash(filepath.Join("/gopath", file[prefixLen:]))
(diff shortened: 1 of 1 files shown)


shurcooL:
Hey @iansmith, thanks for your PR.

I suggest for you to carefully look over the original commit that introduced the `hasGopathPrefix` helper func, if you haven't already.

https://github.com/gopherjs/gopherjs/commit/849fd25ac7bd6817e627b82aa4b60f77e7dda5fc

I wrote that commit so I can give you more context about it. The main task of `hasGopathPrefix` was to solve one problem, that GOPATH was not being parsed as a list of 1 or more workspaces; my goal was to keep the rest of the behavior completely unchanged. That isn't to say the rest of the behavior (using `strings.HasPrefix` on paths, etc.) is completely accurate. But that explains why `hasGopathPrefix` possibly doesn't do the best thing.

iansmith:
Well, I think that my commit above is busted anyway (doesn't generate correct source maps) so maybe we should just generate a warning message if the user has a ".." in their GOPATH?

shurcooL:
Is it even valid for a user to have a `..` element in their GOPATH? That doesn't seem right.

iansmith:
Well, I'm not sure I'd _recommend_ it, but it seems like it should be legit.  I was actually using a script that was setting my GOPATH to something that had ../.. in it and gopherjs "worked" but it generated the jsmaps without full /gopath/... entries, as hasGopathPrefix was not "matching" the paths.  I'd say a warning if any GOPATH entry != that same GOPATH entry after applying "filepath.Clean" might be in order?

shurcooL:
Fair enough, I suppose `..` is a valid path elment, and GOPATH is a list of paths.

neelance:
How do we resolve this? Isn't a simple `filepath.Clean` on `gopathWorkspace` enough?

iansmith:
I think a filepath.Clean on each element in GOPATH would probably solve the problem. I tried to do it after the fact (at the point of generating jsmaps) and that did not work.

neelance:
Please reopen if this change is not enough.

jskDr:
Is there no support for pythonjs anymore?

On Tue, Feb 3, 2015 at 9:09 AM, Richard Musiol <notifications@github.com>
wrote:

> Please reopen if this change is not enough.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gopherjs/gopherjs/pull/166#issuecomment-72566410>.
>



-- 
Best regards,
James Sungjin Kim, Ph.D.
Samsung Advanced Institute of Technology (SAIT)
Samsung. Electronics Co., LTD

shurcooL:
I guess [this](https://github.com/PythonJS/PythonJS#go-backend) is what you're referring to. I've never heard of PythonJS until now.

I would imagine their use of GopherJS is independent and GopherJS does not need to "support" it. It just compiles Go to JavaScript. PythonJS looks like it outputs Go, which GopherJS can then compile.

If you're having problems with PythonJS, I think you should [open an issue](https://github.com/PythonJS/PythonJS/issues) in their repo, not this one.

(I could be mistaken about this, but that's my understanding.)

PR #191, opened 2015-03-04
Fix false positives when using watch flag.

albrow:
The watch flag is an awesome feature! However, I noticed one small issue when I watched what was going on closely. There were some false positives triggering a rebuild when it wasn't necessary.

Here's a demo in which I save the file once but two rebuilds are triggered. One immediately, then another about a second later:

False positive demo

I have used fsnotify before to watch for changes, so I had a hunch for where to look. I found two cases which were causing false positives:

1. Temporary files being created or edited, e.g. those used by editors like Sublime Text for atomic saves. I fixed this by ignoring changes to temporary files.
2. Mac OS Finder indexing the file and making it look like a change. I fixed this by keeping a hash of each file and comparing the hashes to see if anything actually changed. It's possible other operating systems have similar problems, but I'm not sure.

I can understand if you have some hesitation about keeping a large number of hashes of files, but in practice, I haven't noticed any impact on performance. I used [xxhash](https://github.com/vova616/xxhash) which is pretty fast. That being said, it's possible the code could be optimized a bit more. For example, [here](https://github.com/albrow/gopherjs/blob/fix-watch-false-positives/build/build.go#L592) I'm using ioutil to read the entire source file so I can calculate the hash. If you are already reading the source file somewhere, it might make sense to calculate the hash at that point instead of reading the whole file twice. However, I couldn't find a more appropriate place to put it at first glance.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -2,6 +2,7 @@ package build
 
 import (
 	"fmt"
+	"github.com/vova616/xxhash"
 	"go/ast"
 	"go/build"
 	"go/parser"
@@ -20,6 +21,12 @@ import (
 	"github.com/neelance/sourcemap"
 	"gopkg.in/fsnotify.v1"
 )
+
+// a map of known file hashes. This is required to
+// determine whether a file actually changed. It is a workaround
+// to fix false positives that can be triggered by e.g., an
+// operating system indexing a file
+var fileHashes = map[string][]byte{}
 
 type ImportCError struct{}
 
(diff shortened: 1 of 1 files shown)


neelance:
Hey. Thanks for wanting to contribute, much appreciated! The check for temporary files was a good one, but the Finder issue could be more easily resolved by checking `ev.Op`, because in that case `ev.Op == fsnotify.Chmod` only.

albrow:
Gotcha. Thanks!

PR #195, opened 2015-03-14
natives/net/http: set response ContentLength from headers on HEAD requests

encryptio:
The `(*http.Response).ContentLength` field was incorrectly being set to `0` during `HEAD` requests. This patch uses the `Content-Length` response header if it exists, otherwise leaves it at `-1`, documented in `net/http` to mean "don't know"; this matches the behavior of Go's `net/http` package.

### DIFF
--- a/compiler/natives/net/http/http.go
+++ b/compiler/natives/net/http/http.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"io/ioutil"
 	"net/textproto"
+	"strconv"
 
 	"github.com/gopherjs/gopherjs/js"
 )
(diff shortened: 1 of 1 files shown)


compiler/natives/net/http/http.go
```diff
@@ -38,11 +39,22 @@ func (t *XHRTransport) RoundTrip(req *Request) (*Response, error) {
 	xhr.Set("onload", func() {
 		header, _ := textproto.NewReader(bufio.NewReader(bytes.NewReader([]byte(xhr.Call("getAllResponseHeaders").String() + "\n")))).ReadMIMEHeader()
 		body := js.Global.Get("Uint8Array").New(xhr.Get("response")).Interface().([]byte)
+
+		contentLength := int64(-1)
+		if req.Method == "HEAD" {
+			i, err := strconv.ParseInt(header.Get("content-length"), 10, 64)
```
shurcooL: Is it correct that `"content-length"` is being used instead of `"Content-Length"`? It might very well be, I just wanted to confirm.

encryptio:
I tested both and both worked, that happens to be the last capitalization I checked; Content-Length is more canonical though. I'll update the patch.

shurcooL:
LGTM.

neelance:
Thx!

PR #205, opened 2015-04-09
handle non-boolean case clauses in switch statement with empty tag

### LINKED ISSUE #204: bug in switch statements with empty switch expression
Given the file `min.go`:

```go
package main

import "fmt"

func main() {
	var zero interface{} = 0
	switch {
	case zero:
		fmt.Println("zero")
	default:
		fmt.Println("default")
	}
}
```

`go run` outputs "default":

```
$ go run min.go
default
```

but `gopherjs run` outputs "zero":

```
$ gopherjs run min.go
zero
```

If I switch on the value `true` instead:

```go
package main

import "fmt"

func main() {
	var zero interface{} = 0
	switch true {
	case zero:
		fmt.Println("zero")
	default:
		fmt.Println("default")
	}
}
```

then gopherjs outputs "default" as expected:

```
$ gopherjs run min.go
default
```

jeremyschlatter:
Fixes #204

For switch statements with no tag, we effectively just cast each case
expression to a bool. But a case expression can have type interface{},
and casting the javascript equivalent to a bool always returns true. To
fix that, this commit converts such an expression X to the expression
"true == x" before translation to javascript.

### DIFF
--- a/compiler/statements.go
+++ b/compiler/statements.go
@@ -78,6 +78,17 @@ func (c *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 					Y:  cond,
 				})
 			}
+		} else {
+			translateCond = func(cond ast.Expr) *expression {
+				if c.p.Info.TypeOf(cond).Underlying() == types.Typ[types.Bool] {
+					return c.translateExpr(cond)
+				}
+				return c.translateExpr(&ast.BinaryExpr{
+					X:  c.newIdent("true", types.Typ[types.Bool]),
+					Op: token.EQL,
+					Y:  cond,
+				})
+			}
 		}
 		c.translateBranchingStmt(s.Body.List, true, translateCond, nil, label, c.Flattened[s])
 
--- a/tests/misc_test.go
+++ b/tests/misc_test.go
@@ -283,3 +283,12 @@ func zero() int {
 	}()
 	panic("")
 }
+
+func TestSwitchEmptyTag(t *testing.T) {
+	var i interface{} = 0
+	switch {
+	case i:
+		t.Fail()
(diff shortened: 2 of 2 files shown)


shurcooL:
This looks reasonable.

The [Go spec](https://golang.org/ref/spec#Switch_statements) says that if the switch condition is not present, it's effectively `true`:

```Go
switch x := f(); {  // missing switch expression means "true"
```

neelance:
The idea of your fix looks good. However, I would like to not just add code, but clean up the implementation. The LHS of the equality should really just default to `true` without having two branches with `&ast.BinaryExpr{...}` and to not check the type there, but instead at a later stage reduce the JS output `a === true` / `true === a` to just `a`. Also, better use `types.Identical` instead of comparing `types.Type` directly.

jeremyschlatter:
That sounds good. I'll aim to send out a cleaner version next week.

neelance:
Feel free to contact me via Google Hangout or Skype if you need some advice.

neelance:
@jeremyschlatter Hey, could you give me some update on your status? I'm quite busy myself at the moment, so no worries. Please just say if I should also dig into this issue if I find some spare time. :-)

jeremyschlatter:
@neelance I haven't started the cleaner version yet. Go ahead and dig into it if you find some time before I do. I'll update this thread if I make progress.

neelance:
I've implemented d5efe76dcb1b11c082e60945b936d620bfd78566 now, which fixes the bug and avoids additional `true` constants in the output. Thanks for your pull request anyways!

jeremyschlatter:
Awesome. Thanks!

PR #207, opened 2015-04-12
Use filepath.Join to manipulate a filepath.

shurcooL:
`http.Dir` expects a native filepath, so it's correct to use `filepath.Join` instead of hardcoding the filepath separator as '/'.

See https://godoc.org/net/http#Dir for details:

> While the FileSystem.Open method takes '/'-separated paths, a Dir's string value is a filename on the native file system, not a URL, so it is separated by filepath.Separator, which isn't necessarily '/'.


neelance:
Completely right. Thanks!

PR #209, opened 2015-04-13
fix for goroutine blocking bookkeeping

jeremyschlatter:
I imagine this problem shows up elsewhere, but the clearest
demonstration I can find is in the behavior of runtime.NumGoroutine():
The second time a goroutine blocks, and every time thereafter, it
invokes the goroutine exiting logic. This causes runtime.NumGoroutine to
underreport the number of goroutines, because the exiting logic includes
decrementing that count. In my testing, runtime.NumGoroutine was often
negative.

The fix is pretty simple. When calling the `r.$blk` closure, return its
output, which may itself be another closure.

### DIFF
--- a/tests/misc_test.go
+++ b/tests/misc_test.go
@@ -297,8 +297,8 @@ func TestNumGoroutine(t *testing.T) {
 	c <- true
 	c <- true
 	c <- true
-	if g, w := runtime.NumGoroutine(), n+1; g != w {
-		t.Errorf("runtime.NumGoroutine(): Got %d, want %d.", g, w)
+	if got, want := runtime.NumGoroutine(), n+1; got != want {
+		t.Errorf("runtime.NumGoroutine(): Got %d, want %d.", got, want)
 	}
 	c <- true
 }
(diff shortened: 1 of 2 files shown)


tests/misc_test.go
```diff
@@ -283,3 +284,21 @@ func zero() int {
...
+	go func() {
+		<-c
+		<-c
+		<-c
+		<-c
+	}()
+	c <- true
+	c <- true
+	c <- true
+	if g, w := runtime.NumGoroutine(), n+1; g != w {
```
shurcooL: I think it'd be nicer to fully spell out the variables `g`, `w` as `got` and `want`. That seems to be more common, and IMO, more clear.

The rest of this test looks good to me.

jeremyschlatter: You're right, `got` and `want` are more common:

```
$ cd `go env GOROOT`
$ grep -r 'if g, w := ' | wc -l
      16
$ grep -r 'if got, want := ' | wc -l
      48
```

```
$ cd `go env GOPATH`
$ grep -r 'if g, w := ' | wc -l
      76
$ grep -r 'if got, want := ' | wc -l
     362

neelance:
Good find!

PR #218, opened 2015-04-29
crypto/rand: avoid QuotaExceededError in window.crypto

dchest:


neelance:
LGTM

shurcooL:
I have 2 comments here:

1. `crypto/rand.Read` is documented:

    > Read is a helper function that calls Reader.Read using io.ReadFull. On return, n == len(b) if and only if err == nil.

    I think that contract is broken here, isn't it? If I call `rand.Read` with a slice of size greater than 65536, it will return n != len(b) and err == nil.

2. Are there any timing requirements on `rand.Read`? If not, then can't this situation be handled differently by using a loop and calling `crypto.Call("getRandomValues", array.Call("subarray", offset, offset+n))` multiple times as long as the remaining bytes to be populated are greater than 65536?

dchest:
1. Correct, the contract is broken.

2. No timing requirements: the length of requested array is not considered secret, so it's okay fetch bytes in loop until we get them all.

neelance:
You are confusing `rand.Read` (package function) and `rngReader.Read` (reader function). For the reader it is the intended behavior to return less if not enough is available.

dchest:
Ah, @neelance is right, the contract is not broken. The reader can return fewer bytes (e.g. see https://golang.org/src/crypto/rand/rand_unix.go#L64). `crypto.Read` though uses `ReadFull` internally https://golang.org/src/crypto/rand/rand.go.

shurcooL:
Indeed, I made a mistake and missed that it's `func (r *rngReader) Read(b []byte)` and not `func Read(b []byte) (n int, err error)`, sorry about that.

The current code is indeed correct.

PR #223, opened 2015-05-06
Implement runtime.Callers

nlacasse:
The implementation returns 0, and leaves `pc` unchanged.


shurcooL:
Just out of curiosity, what code are you writing or running @nlacasse that prompted this addition?

nlacasse:
I have an error library that uses `runtime.Callers` to capture stack trace information when errors are created.

These stack traces won't work correctly with gophersjs's mocked `runtime.Callers` function, but at least the library can compile.

shurcooL:
Yep, that makes perfect sense. Thanks.

PR #231, opened 2015-05-31
gopherjs: Minor improvements.

shurcooL:
Print potential errors to stderr rather than stdout.

Use `syscall.Stderr` instead of hardcoding a magic number. It increases readability.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -35,12 +35,12 @@ func init() {
 	var err error
 	currentDirectory, err = os.Getwd()
 	if err != nil {
-		fmt.Println(err)
+		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 	currentDirectory, err = filepath.EvalSymlinks(currentDirectory)
 	if err != nil {
-		fmt.Println(err)
+		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 }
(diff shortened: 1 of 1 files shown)


tool.go
```diff
@@ -55,7 +55,7 @@ func main() {
 	flagWatch := pflag.Lookup("watch")
 	pflag.BoolVarP(&options.Minify, "minify", "m", false, "minify generated code")
 	flagMinify := pflag.Lookup("minify")
-	pflag.BoolVar(&options.Color, "color", terminal.IsTerminal(2) && os.Getenv("TERM") != "dumb", "colored output")
+	pflag.BoolVar(&options.Color, "color", terminal.IsTerminal(syscall.Stderr) && os.Getenv("TERM") != "dumb", "colored output")
```
neelance: This does not compile for Windows. I had to change it a bit: 979c6e69961b8f41759a6c44f9bf7a6404cdf50b

shurcooL: I see, sorry about that.

I was a little surprised that was the case because `os.Stderr` is defined exactly the same on _all_ platforms:

http://gotools.org/os#Stderr

But I see that the type of `syscall.Stderr` is not `int` on Windows, because it's defined as:

```
Stderr = getStdHandle(STD_ERROR_HANDLE)
```

And `getStdHandle` returns type `Handle` which is defined as `uintptr`. So I should've done `terminal.IsTerminal(int(syscall.Stderr))` to make it work everywhere. But I guess `os.Stderr.Fd()` is also valid (and maybe simpler/easier to understand, since more people are familiar with `os` package than `syscall` package).

Thanks!

Issue #121, opened 2014-11-04
Add "gopherjs serve" command


neelance:

albrow:
@neelance, do you want some help with this? I'm very comfortable using the http package in go and built something similar to what you're talking about in [a static blog generator](https://github.com/albrow/scribble) I'm working on. I understand this might be low priority, but it also seems relatively independent and a great candidate for something I could help with. Let me know and I'll work up a PR!

neelance:
Truth is, I'm undecided. I am very happy that you want to contribute. On the other hand I already have quite specific plans for this and it was one of the next things on my todo list, because I need it before starting to promote https://github.com/neelance/dom/tree/master/examples/todomvc . Speaking of promotion, publishing helper libraries, blog posts, etc. would be tremendously useful, because we still need to raise awareness.

albrow:
Totally understand if you think it's easier to just do it yourself. I would love to see gopherjs succeed, so I thought I'd offer to help here and see if it made sense. For now, I'll try and do what I can to help you with promotion. For me that means writing blog posts and building libraries on top of gopherjs.

neelance:
Initial implementation is done. You can try it on the example of https://github.com/neelance/dom . Feedback very welcome. :-)

shurcooL:
Just tried it out. My goal is to deprecate my personal tool and rely on `gopherjs serve`, because it's more general and other people can use it too.

So far it works well. It took me a minute or two to figure out. I've only tried it on my very simple, single `main.go` file GopherJS project.

I like that it serves all of your GOPATH so can use same command to test multiple GopherJS projects (my personal tool only served current directory). But I wonder how that will handle projects that have static assets in working directory of the Go package.

One minor thought/suggestion, I wish it would use port 8080 by default rather than 6060. 6060 is typically used by godoc, and I've always used 8080 for testing personal projects. I know it can be set, which is great, but I just think 8080 is a better default than 6060.

I'll post more feedback as I have it, but so far I think this will work really well!

shurcooL:
It works for Go packages with multiple `.go` files too, so that's great. It looks like one just needs to include `<script src="main.js" type="text/javascript"></script>` in the `index.html` file, that's all (if the default `index.html` doesn't suit you).

shurcooL:
It works for static assets too, but only if I use relative paths.

neelance:
Changed default port to 8080. Yes, assets need relative paths.

slimsag:
[...]

PR #232, opened 2015-05-31
gopherjs serve: Display compilation errors in browser console.

shurcooL:
This is nice when using with gopherjs serve, because it already recompiles on browser refresh. This allows one to switch between code editor and browser for a faster iteration cycle. No need to also switch to the terminal to see errors, as now both compilation and runtime errors are visible in browser console.

I'm backporting some enhancements from my `gopherjs_serve_html` tool, so I can stop using it and rely on `gopherjs serve` full time.

This resolves a future enhancement that was mentioned in https://github.com/gopherjs/gopherjs/issues/121#issuecomment-87495493.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -125,7 +125,7 @@ func main() {
 					}
 				}
 				return nil
-			}, options)
+			}, options, nil)
 
 			if s.Watcher == nil {
 				os.Exit(exitCode)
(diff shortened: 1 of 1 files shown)


shurcooL:
A possible future enhancement here is to make use of source mapping to have the browser display compilation errors in the actual Go code.

I'm not very familiar with the details of how source mapping works, but I think it can be done.

PR #233, opened 2015-05-31
gopherjs serve: Allow specifying host and port, print all addresses served.

shurcooL:
Previously, gopherjs serve would print "serving at http://localhost:port" but actually bind at ":port". That means it was serving content at "localhost:port" and also any other interfaces.

This changes the "port" flag from an int (default 8080), to "http" parameter (default ":8080") that allows one to specify either just the port, or port and host to bind at. This "http" flag is common in many Go tools, including godoc.

Additionally, print all addresses where content is being served. This is helpful in case you want to share a non-localhost IP with someone else or visit the page on another computer or mobile device. For example, the output may be:

	$ gopherjs serve
	serving at:
	http://localhost:8080
	http://10.0.0.7:8080

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -10,6 +10,7 @@ import (
 	"go/token"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
@@ -433,13 +434,13 @@ func main() {
 	cmdServe.Flags().AddFlag(flagMinify)
 	cmdServe.Flags().AddFlag(flagColor)
 	cmdServe.Flags().AddFlag(flagTags)
-	var port int
-	cmdServe.Flags().IntVarP(&port, "port", "p", 8080, "HTTP port")
+	var addr string
+	cmdServe.Flags().StringVarP(&addr, "http", "", ":8080", "HTTP bind address to serve")
 	cmdServe.Run = func(cmd *cobra.Command, args []string) {
 		dirs := append(filepath.SplitList(build.Default.GOPATH), build.Default.GOROOT)
 		sourceFiles := http.FileServer(serveCommandFileSystem{options: options, dirs: dirs, sourceMaps: make(map[string][]byte)})
-		fmt.Printf("serving at http://localhost:%d\n", port)
-		fmt.Println(http.ListenAndServe(fmt.Sprintf(":%d", port), sourceFiles))
+		printServingAt(addr)
+		fmt.Fprintln(os.Stderr, http.ListenAndServe(addr, sourceFiles))
 	}
 
 	rootCmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


neelance:
I'm okay with changing the flag. However, I'm not convinced of printing all addresses. In the most cases the user will not need it. How about just printing `http://{http flag}` and use `localhost` as host if not specified?

shurcooL:
It'd be great to get other people's thoughts, but I think printing the actual IPs being served is really helpful.

For one, it would avoid the problem that the URL is not copy-pasteable if you print:

```
serving at http://:8080
```

That is not a valid URL and makes no sense. Some browsers will interpret it anyway, but others won't, and some people will think it's a bug in their browser. Let's print correct urls.

Additionally, if you use default flag, it may be helpful to realize your content is available to other people on the network. So if you want to test something sensitive/private, you should explicitly set `--http=localhost:8080"`.

Finally, if you want to visit the served content on a mobile device in order to test, it's convenient that the non-loopback URL is visible and no need to investigate to find out your IP.

I [have this feature](https://github.com/shurcooL/cmd/blob/a9a2f166c984437b35d436060cf657fca9057854/gopherjs_serve_html/main.go#L32-L51) in my `gopherjs_serve_html` command and I'm personally very glad I added it.

dominikh:
I'm not a fan of listing all addresses. That list would only be an approximation and it has the potential to be too noisy in complex network configurations with a lot of interfaces. 

Personally, I can't think of any tool that prints all addresses.

P.S. This is unrelated, but one issue I found with the current implementation is that listening on port 0 will actually print port 0, instead of the random ephemeral port that the OS picked.

shurcooL:
> P.S. This is unrelated, but one issue I found with the current implementation is that listening on port 0 will actually print port 0, instead of the random ephemeral port that the OS picked.

That sounds like a valid problem, can you please open a separate issue since it is independent of this PR?

> P.P.S I'm not sure if the default really should be ":8080" – wouldn't it be much safer to listen on localhost only by default?

Yes it would, and one of my original thoughts (back when it was just a `port` flag) was to fix the inconsistency by forcing `http.ListenAndServe(fmt.Sprintf("localhost:%d", port))`. But then I remembered I often want to visit a GopherJS project on my mobile device for testing, and localhost would make that impossible.

I think ":8080" is a better default (rather than "localhost:8080"). It's more user friendly. It's less common you are serving really sensitive stuff, so in those infrequent cases, you can explicitly specify `--http="localhost:8080"` rather than assume it's the default.

neelance:
I'd also prefer the user friendly way of serving all interfaces by default.

shurcooL:
I'd like to make progress on this. Let's identify what we agree on, and what still needs to be resolved for this PR to be merged.

I think we agree that the current default of serving on ":8080" (aka 0.0.0.0:8080 aka "bind on all addresses") is good and we should keep it. It's a better default because it's more user friendly and convenient, and the most commonly used case.

In rare cases the user wants to serve something sensitive, they should **explicitly** opt in to serving on "localhost:8080" only - that is safer than relying on a default value for limited access.

Now, my question is what should be printed when you specify `-addr=:8080` flag?

An original suggestion was:

> How about just printing `http://{http flag}` and use `localhost` as host if not specified?

I feel somewhat strongly against doing that because it's untruthful (rather, it's not the full truth). You're telling the user it's serving on http://localhost:8080, while in reality it is serving there **but also** on http://10.0.0.7:8080 and any other public interfaces. This is potentially misleading in terms of security, the user may think it really is serving on localhost only, while in reality they are serving on public IPs also and sensitive information may be leaked.

So, if not the list of all public IPs the way I've done in this PR currently, what should be printed instead?

> P.S. This is unrelated, but one issue I found with the current implementation is that listening on port 0 will actually print port 0, instead of the random ephemeral port that the OS picked.

That should also be addressed, as saying it's listening on port 0 is also untruthful.

neelance:
What about `Serving on :8080, e.g. http://localhost:8080` when no host is given, else just `Serving on http://192.168.0.1:8080`? My goal is to have a single easy link that the user can just click on. But you're right that it is important to not hide the fact that all interfaces are used.

shurcooL:
> My goal is to have a single easy link that the user can just click on.

If we print all addresses, the idea is that _all_ of them would be clickable and just work. Some would work from the current machine only (e.g., localhost), while others would be available on local network as well. None of the links printed would be unusable.

> In the most cases the user will not need it.

I think that depends on the use case. If you're testing something on a mobile device rather than local computer, you will need the non-localhost address every time. If you're doing testing on all devices (computers, iOS browser, Android browser, etc.), you will also need non-localhost address.

I admit that I test/try things on my mobile device rather often, which is why I went out of my way to write the code that prints all addresses. If I weren't testing on a mobile device at all, I wouldn't care for this functionality.

> Personally, I can't think of any tool that prints all addresses.

Is that because it's a bad experience, or because other tools don't use Go where it's this easy to find out all public interfaces that will be bound to? If I wrote this tool in C++, for example, I would also not have this feature just because it'd be too hard to do.

I wonder what those tools/frameworks that are meant to assist with developing and testing on mobile devices do. It's hard to imagine they don't try to make it easier to test on a mobile device by printing a url that is usable.

It's been 2 months and I'm not completely convinced printing all public IPs would be a worse user experience. However, no one else is convinced that it would be better either. So in interest to make progress, I agree on a simple solution that is okay with Richard and is an improvement over current state. The suggestion was to print something (truthful) like this:

```
$ gopherjs serve
serving on port 8080 on all interfaces, e.g. http://localhost:8080
```

I'll implement that soon.

shurcooL:
> **shurcooL** wants to merge 1 commit into `gopherjs:master` from `unknown repository`

It seems I've deleted my gopherjs fork since making the PR 2 months ago, so there's no branch than I can push (or force push) to with further commits. I'll close this and make a new PR.

PR #240, opened 2015-06-07
Make nil == null, js.Undefined == undefined a bijection.

### LINKED ISSUE #237: Cannot read property 'constructor' of undefined
I ran into this one when trying to use Go functions that are callable from JavaScript code.

Playground link:
http://www.gopherjs.org/playground/#/A_bP7IWgxi

Basically, $internalize has no code path for 'undefined' values that are mapped to interfaces.  

I would like to recommend the following fix, for line 245 in compiler/prelude/jsmapping.go like so:
```
    if (v === null || v === undefined) {
      return $ifaceNil;
    }
```
This changes $internalize such that undefined is converted to nil, just like null, for interfaces.

shurcooL:
This PR is a work in progress.

Fixes #237.

- Add failing test for nil, undefined equality.
- Run gopherjs tests with both GopherJS and Go.


shurcooL:
Ok, good, the tests I've added are indeed failing now, which is expected:

https://circleci.com/gh/gopherjs/gopherjs/453

shurcooL:
Woohoo, everything works! Tests pass, including old tests, and newly added tests.

Now that's the case, next step would be to see if the code can be refactored to be simpler/more direct/better. @neelance, please take a look and review.

compiler/prelude/prelude.go
```diff
@@ -372,7 +372,7 @@ var $internalAppend = function(slice, array, offset, length) {
 
 var $equal = function(a, b, type) {
   if (type === $jsObjectPtr) {
-    return a === b;
+    return a === b || (a.hasOwnProperty("object") && a.object === undefined && b.hasOwnProperty("object") && b.object === undefined);
```
shurcooL: It might make sense to factor out this "if both a and b are undefined, return true" check out into `$interfaceIsEqual`, since that's where "if both a and b are nil interfaces, return true" happens. I wasn't sure which is better.

Also, I am unsure about the way "if both a and b are undefined" is being checked. It seems to work, but maybe there's a much better JavaScript way?

circle.yml
```diff
@@ -12,3 +12,4 @@ dependencies:
...
+    - go test -v -race ./...
```
shurcooL: The beauty of doing this is that it verifies that my "expected test output" for GopherJS tests match what the actual Go compiler produces. So doing this is a great idea.

Note that the newly added `TestNilUndefinedEquality` test was actually passing using Go compiler without me making any changes (clearly, this PR only makes changes to the GopherJS compiler, not Go compiler), and I used this to verify.

tests/nil_undefined_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	if got, want := (interface{})(nil) == js.Undefined, false; got != want {
+		t.Errorf("\n(interface{})(nil) == js.Undefined: got %v, want %v", got, want)
+	}
+	if got, want := (interface{})(nil) == ui, false; got != want {
+		t.Errorf("\n(interface{})(nil) == ui: got %v, want %v", got, want)
+	}
+	if got, want := (interface{})(nil) == u, false; got != want {
+		t.Errorf("\n(interface{})(nil) == u: got %v, want %v", got, want)
+	}
```
shurcooL: Some of these above may be already present everywhere, if so, they should be removed.

neelance:
The tests in `github.com/gopherjs/gopherjs/tests` only use Go features, but not the `js` package. That's why they can be run by both, normal Go and GopherJS. Tests for the `js` package are at https://github.com/gopherjs/gopherjs/blob/master/js/js_test.go and they are guarded by `// +build js` because the `js` package is not intended to work with pure Go (it could provide features to create mocks for testing, but that functionality is not implemented yet).

I am grateful that you wanted to tackle this bug, but it seems to me like you are not aware about how `*js.Object` works internally (and it is not really documented, sorry). The basic idea is that whenever something has the type `*js.Object` (with pointer), then it will directly contain a native JS object at runtime. This is why `a === b` works if the type which is being checked for equality is `*js.Object`. No need to change that. But this only applies to `*js.Object`. For `js.Object` (without pointer) and interfaces, the normal `js.Object` Go struct is used, which is recursive: It has a `*js.Object` field which, according to the rule above, can hold the JS object. Voila! For example take a look at https://github.com/gopherjs/gopherjs/blob/master/compiler/expressions.go#L1092

shurcooL:
> I am grateful that you wanted to tackle this bug, but it seems to me like you are not aware about how *js.Object works internally (and it is not really documented, sorry).

No problem, it's true that I am poorly familiar with internal details of GopherJS compiler, in fact this was my first time trying to make a PR for it, so I expected that even if I got it to work and pass tests, there's still a high chance my entire approach is wrong or there's a simpler, better, more direct way to accomplish it.

> the js package is not intended to work with pure Go

Is this documented somewhere? I thought the `js` package did not provide functionality beyond the current mocks, but I didn't realize it was completely not intended to be used by normal Go compiler. If that's the case, what are mocks for? Are they just for docs?

Note that I initially wrote the if guard like this:

```Go
if runtime.GOARCH == "js" { ... }
```

But then changed it. Anyway, I thought importing `js` package was okay to do without hiding it behind js build tag.

neelance:
The mocks are there for type checking, `go/types` needs something to work on. It is also quite good that you can compile it with pure Go without any errors. Just when using it at runtime, don't expect any meaningful thing to happen. That's why I'd say that printing `js.Undefined` right in pure Go is not a requirement. All other stuff of the `js` package doesn't work either with pure Go.

neelance:
In short: The behavior of the `js` package with pure Go is currently not defined at all.

shurcooL:
Ok. How would you like to proceed? Do you want to give me actionable feedback to change this PR so it can be accepted, or do you want to do this yourself?

neelance:
I would love to help you. I'll post some comments on the code.

circle.yml
```diff
@@ -4,7 +4,7 @@ machine:
 
 dependencies:
   pre:
-    - cd /usr/local && sudo rm -rf go && curl https://storage.googleapis.com/golang/go1.4.1.linux-amd64.tar.gz | sudo tar -xz && sudo chmod a+w go/src/path/filepath
+    - cd /usr/local && sudo rm -rf go && curl https://storage.googleapis.com/golang/go1.4.2.linux-amd64.tar.gz | sudo tar -xz && sudo chmod a+w go/src/path/filepath
```
neelance: Separate commit pls.

compiler/prelude/jsmapping.go
```diff
@@ -245,6 +245,9 @@ var $internalize = function(v, t, recv) {
     if (v === null) {
       return $ifaceNil;
     }
+    if (v === undefined) {
+      return new $jsObjectPtr(undefined);
+    }
```
neelance: This looks good. The JS `undefined` is stored in an interface by wrapping it into `*js.Object`.

compiler/prelude/prelude.go
neelance: As I've written in the other comments, the `*js.Object` type holds the JS value itself. So when comparing two `undefined`, the JS value in `a` would be `undefined` and `b` would also be `undefined`. So `a === b` holds.

js/js.go
```diff
@@ -59,7 +59,12 @@ func (o *Object) New(args ...interface{}) *Object { return o.object.New(args...)
 func (o *Object) Bool() bool { return o.object.Bool() }
 
 // String returns the object converted to string according to JavaScript type conversions.
-func (o *Object) String() string { return o.object.String() }
+func (o *Object) String() string {
+	if o == Undefined {
+		return "undefined"
+	}
+	return o.object.String()
+}
```
neelance: This should not be changed, because it is just a proxy when the JS value is wrapped by the real `*js.Object` struct (in case of being stored in an interface or `js.Object` without pointer).

compiler/prelude/prelude.go
shurcooL: Unless I change something else elsewhere, `a === b` did not hold. That's why I needed to add the or block. If you remove it, tests will not pass.

tests/nil_undefined_test.go
```diff
@@ -0,0 +1,72 @@
...
+		t.Errorf("\nui == js.Undefined: got %v, want %v", got, want)
+	}
+	var u = js.Undefined
+	if got, want := u == js.Undefined, true; got != want {
+		t.Errorf("\nu == js.Undefined: got %v, want %v", got, want)
+	}
+	if got, want := u == ui, true; got != want {
+		t.Errorf("\nu == ui: got %v, want %v", got, want)
+	}
+	if got, want := (interface{})(nil) == js.Undefined, false; got != want {
```
neelance: In pure Go `js.Undefined` is just a `nil` stub. So this comparison is comparing the interface values `(nil, nil)` and `(*js.Object, nil)`, which is `false` (the tricky `nil` situation with interfaces...). In GopherJS on the other hand, `(interface{})(nil)` gives JS's `null` and `js.Undefined` gives `undefined`, which are also unequal.

compiler/prelude/prelude.go
neelance: Why was this necessary? Which test was failing?

neelance: This change is really not the correct way. If something fails, then we'll have to look for the bug somewhere else. ;-)

shurcooL: Have you reviewed #237? This PR is to fix the issue described there, and it uses the solution I proposed. I should make sure you agree with that proposal first.

> Why was this necessary? Which test was failing?

The test for #237 that I added.

Namely, calling `js.Global.Call("eval", "(foo(null))"))` would print expected output, but calling `js.Global.Call("eval", "(foo(undefined))")` would print:

```Go
	// x: undefined
	// x == nil: false
	// x == js.Undefined: false
```

Instead of:

```Go
	// x: undefined
	// x == nil: false
	// x == js.Undefined: true
```

compiler/prelude/jsmapping.go
shurcooL: I think the reason `a === b` doesn't hold below when comparing two instances of `undefined` may be because I'm calling `new` every time here. Maybe it should be done outside, and this can return a single/the same instance of `new $jsObjectPtr(undefined)"` every time.

compiler/prelude/prelude.go
neelance: In the `x == js.Undefined` case, `a` and `b` on this line should end up being the JS value `undefined`. But this does not happen. The issue is in `$interfaceIsEqual`, which does not correctly unwrap the JS value. It already shows with this simple test:
```go
var i interface{} = js.Undefined
println(i == js.Undefined)
```
So `$interfaceIsEqual` needs to be fixed. I think it does not even need to call `$equal` in this case, because this just boils down to `a === b` which can be inlined.

shurcooL: That sounds like the way to go. How should I unwrap the 'undefined` value in there?

neelance: Just get the `object` field. Check for the type first, then you don't need `hasOwnProperty`.

tests/nil_undefined_test.go
shurcooL: You're right. That is far too much of a coincidence to rely on for tests to pass. I will remove that.

shurcooL:
> I am grateful that you wanted to tackle this bug, but it seems to me like you are not aware about how *js.Object works internally (and it is not really documented, sorry).

Just to prove how little I knew when I began working on this PR, you can see my [WIP commit here](https://github.com/shurcooL/gopherjs/commit/4aa3f0031833a68af730682eee7ee49cf3e9d038) from when I got it to work, before I cleaned up all the debugging/test code. :P

circle.yml
shurcooL: Done in a55786ee761f03ca5dd9e555b992c4a15c3188e0.

shurcooL:
I've addressed all your feedback, PTAL.

I've moved the Go 1.4.1 -> 1.4.2 update into a separate commit.

I've removed the `github.com/gohperjs/gopherjs/js` tests from `github.com/gohperjs/gopherjs/tests` package, so it's no longer testing it via Go compiler. Instead, it's only in `github.com/gohperjs/gopherjs/js` package tests (in `js_test.go`).

I've broken down the verbose tests into minimal previously-failing tests.

There are now two compiler commits:

- One fixing interface equality for `undefined` (`js.Undefined`), see cb553f393ba9e6cbab0b47c2d0bbff1ce79e8306. It comes with a test for it.

- Another fixing missing internalize case for `undefined` value, see bf81f3863f6c39cdad954f7f31882c7d967a2890. It comes with a test for it. Together with above, it closes #237.

All tests are passing. Please review again and let me know how this looks.

neelance:
LGTM

PR #249, opened 2015-07-09
Add initial support for Go 1.5.

shurcooL:
- Add time.loadLocation to natives.
- Add runtime.Version to natives.
- Add pprof.StartTrace, pprof.StopTrace to natives.

These are changes I had to make to GopherJS to get it to be able to compile a few simple and medium sized Go packages, using Go version `go1.5beta1`.

I'm not very familiar with how natives works, and why it seems some funcs are provided while others are missing (I'd love to learn more about it), so this may not be entirely correct, but I hope it's helpful as an initial start. After making these changes, everything seemed to work great so far.

This should not be merged into `master` branch until Go 1.5 final version is officially released. Making this PR for visibility, to run CI, and because I can't make a PR into a non-existing branch (`master` is the only choice).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ machine:
 
 dependencies:
   pre:
-    - cd /usr/local && sudo rm -rf go && curl https://storage.googleapis.com/golang/go1.4.2.linux-amd64.tar.gz | sudo tar -xz && sudo chmod a+w go/src/path/filepath
+    - cd /usr/local && sudo rm -rf go && curl https://storage.googleapis.com/golang/go1.5beta1.linux-amd64.tar.gz | sudo tar -xz && sudo chmod a+w go/src/path/filepath
   post:
     - npm install --global node-gyp
     - cd node-syscall && node-gyp rebuild && mkdir -p ~/.node_libraries/ && cp build/Release/syscall.node ~/.node_libraries/syscall.node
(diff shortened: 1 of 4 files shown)


shurcooL:
Despite it working okay on the packages I've tested (no errors, things worked), there are quite a few failing tests in CI.

They seem to be related primarily to:

- `Error: Native function not implemented: os.runtime_beforeExit`
- `Error: Native function not implemented: reflect.typedmemmove`

cryptix:
Very nice!

Wouldn't it be super crazy awesome if we could hook up the tracing into web developer tools that support similar tracing? Not sure if there are browser APIs for this, just typing out loud.

shurcooL:
> Wouldn't it be super crazy awesome if we could hook up the tracing into web developer tools that support similar tracing? Not sure if there are browser APIs for this, just typing out loud.

Yes! I think that might fall out of scope of this PR (depending on if it's trivial to do, or needs some investigation and work), but it can be a followup enhancement.

Note that GopherJS already taps into the JavaScript debugger for its implementation of [`runtime.Breadkpoint`](https://github.com/gopherjs/gopherjs/blob/595af360f71b627f5db58b1929602675892119c1/compiler/natives/runtime/runtime.go#L43-L45), so this is definitely in the realm of doable and a good idea.

shurcooL:
Closing this PR as it has been merged onto [`go1.5` branch](https://github.com/gopherjs/gopherjs/commits/go1.5) in this repository, which is where Go 1.5 support will continue to be worked on.

PR #261, opened 2015-07-18
Do not include _test.go files (of stdlib deps) when not building tests.

### LINKED ISSUE #260: _test.go files of stdlib (natives) get included for normal builds.
When doing `gopherjs build` or `gopherjs serve`, all _test.go files should be ignored. However, there's a bug in the way the compiler works and it ends up including some _test.go files in some cases.

Those _test.go files can import additional packages (that are only used in tests), thus increasing output .js file size (slightly) and performing side-effects.

### Reproducible test case

```Go
package main

import (
	"flag"
	"os"
)

func main() {
	os.Args = []string{"foo", "-help"}
	flag.Parse()
}
```

http://www.gopherjs.org/playground/#/c8k75vP-A9

Expected and actual output is `Usage of foo: `.

However, if you import, for example, `math/rand` package, expected output should not change (and does not change when using `go` compiler natively). However, using GopherJS, it changes.

http://www.gopherjs.org/playground/#/1cS0SgEyq1

Expected output is `Usage of foo: `.

Actual output:

```
Usage of foo: 
  -test.bench="": regular expression to select benchmarks to run 
  -test.benchmem=false: print memory allocations for benchmarks 
  -test.benchtime=1s: approximate run time for each benchmark 
  -test.blockprofile="": write a goroutine blocking profile to the named file after execution 
  -test.blockprofilerate=1: if >= 0, calls runtime.SetBlockProfileRate() 
  -test.coverprofile="": write a coverage profile to the named file after execution 
  -test.cpu="": comma-separated list of number of CPUs to use for each test 
  -test.cpuprofile="": write a cpu profile to the named file during execution 
  -test.memprofile="": write a memory profile to the named file after execution 
  -test.memprofilerate=0: if >=0, sets runtime.MemProfileRate 
  -test.outputdir="": directory in which to write profiles 
  -test.parallel=1: maximum test parallelism 
  -test.run="": regular expression to select tests and examples to run 
  -test.short=false: run smaller test suite to save time 
  -test.timeout=0: if positive, sets an aggregate time limit for all tests 
  -test.v=false: verbose: print additional output 
```

I've already identified where the issue is; I'll start working on a fix now.

shurcooL:
Fixes #260.

Add `IsTest` bool to `PackageData` to specify if package is being build normally or for tests.

`parse` was made unexported. It was undocumented, and not used by any external packages. Unless there's a good reason to export it, it's better not to, as that keeps the API smaller and simpler, allows internal code to change more easily. For this fix, I changed signature of `parse` to accept an extra parameter.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -90,7 +90,10 @@ func Import(path string, mode build.ImportMode, installSuffix string, buildTags 
 	return pkg, nil
 }
 
-func Parse(pkg *build.Package, fileSet *token.FileSet) ([]*ast.File, error) {
+// parse parses and returns all .go files for given pkg, including all transitive dependencies.
+// isTest is true when package is being built for running tests.
+// Standard Go library files are augmented with files in compiler/natives folder.
+func parse(pkg *build.Package, isTest bool, fileSet *token.FileSet) ([]*ast.File, error) {
 	var files []*ast.File
 	replacedDeclNames := make(map[string]bool)
 	funcName := func(d *ast.FuncDecl) string {
@@ -103,14 +106,17 @@ func Parse(pkg *build.Package, fileSet *token.FileSet) ([]*ast.File, error) {
 		}
 		return recv.(*ast.Ident).Name + "." + d.Name.Name
 	}
-	isTestPkg := strings.HasSuffix(pkg.ImportPath, "_test")
+	isXTest := strings.HasSuffix(pkg.ImportPath, "_test")
 	importPath := pkg.ImportPath
-	if isTestPkg {
+	if isXTest {
 		importPath = importPath[:len(importPath)-5]
 	}
 	if nativesPkg, err := Import("github.com/gopherjs/gopherjs/compiler/natives/"+importPath, 0, "", nil); err == nil {
-		names := append(nativesPkg.GoFiles, nativesPkg.TestGoFiles...)
-		if isTestPkg {
+		names := nativesPkg.GoFiles
+		if isTest {
+			names = append(names, nativesPkg.TestGoFiles...)
+		}
+		if isXTest {
 			names = nativesPkg.XTestGoFiles
 		}
 		for _, name := range names {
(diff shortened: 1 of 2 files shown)


shurcooL:
I think I see a much cleaner way to achieve the behavior change I wanted, and there's no need for the new `IsTest` bool.

The `*build.Package` provided when doing `collectTest` is not the original, but a custom-made one:

```Go
if err := collectTests(&build.Package{
	ImportPath: pkg.ImportPath,
	Dir:        pkg.Dir,
	GoFiles:    append(pkg.GoFiles, pkg.TestGoFiles...),
	Imports:    append(pkg.Imports, pkg.TestImports...),
}, "_test", &tests.NeedTest); err != nil {
	return err
}

if err := collectTests(&build.Package{
	ImportPath: pkg.ImportPath + "_test",
	Dir:        pkg.Dir,
	GoFiles:    pkg.XTestGoFiles,
	Imports:    pkg.XTestImports,
}, "_xtest", &tests.NeedXtest); err != nil {
	return err
}
```

Given that, I think a much simpler fix would be to just change `Parse` to only look at `GoFiles`, never `TestGoFiles` or `XTestGoFiles`.

shurcooL:
Oh, not quite. It still needs to use `TestGoFiles` and/or `TestXGoFiles` of natives packages depending on the build mode (normal, internal test, external test).

build/build.go
```diff
@@ -90,7 +90,10 @@ func Import(path string, mode build.ImportMode, installSuffix string, buildTags
 	return pkg, nil
 }
 
-func Parse(pkg *build.Package, fileSet *token.FileSet) ([]*ast.File, error) {
+// parse parses and returns all .go files for given pkg, including all transitive dependencies.
```
neelance: What do you mean by `including all transitive dependencies`?

shurcooL: After looking at the code more, I think that's incorrect.

I originally thought `Parse` was including all dependencies and all dependencies of dependencies, recursively. But now I see it's only including the files of the given package. However, it does special processing for standard library packages that have some code in `natives` directory.

So yeah, that comment is wrong and should be changed. This is why describing what this func does will helpful, since it takes time and effort to figure it out from code.

shurcooL: > This is why describing what this func does will helpful, since it takes time and effort to figure it out from code.

I still can't believe I got it all wrong. I literally took 3-5 minutes carefully reading and looking over the func code to try to understand what exactly it does/is responsible for in order to summarize it, and came up with a completely wrong statement about dependencies. Sorry about that.

neelance:
LGTM

PR #264, opened 2015-07-30
gopherjs serve: Allow specifying host and port via http flag.

shurcooL:
Followup to #233.

Previously, gopherjs serve would print "serving at http://localhost:port" but actually bind at ":port". That means it was serving content at "localhost:port" and also any other interfaces.

This changes the `port` flag from an int (default 8080), to `http` parameter (default ":8080") that allows one to specify either just the port, or port and host to bind at. This "http" flag is common to many Go tools, including `godoc`.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -10,6 +10,7 @@ import (
 	"go/token"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


shurcooL:
I've thought through the IPv4 story, but IPv6 support may be lacking (as far as printing the correct "serving at" message goes).

PR #281, opened 2015-08-29
Add Go 1.5 new std pkgs to compatibility table.

shurcooL:
Currently, they are all marked as "untriaged", because I don't know if they are supported. We need to run their tests and find out.

That can be done in this PR before merging.


circle.yml
```diff
@@ -11,5 +11,5 @@ dependencies:
...
...zip bufio bytes compress/bzip2 compress/flate compress/gzip compress/lzw compress/zlib container/heap container/list container/ring crypto/aes crypto/cipher crypto/des crypto/dsa crypto/ecdsa crypto/elliptic crypto/hmac crypto/md5 crypto/rand crypto/rc4 crypto/rsa crypto/sha1 crypto/sha256 crypto/sha512 crypto/subtle crypto/x509 database/sql database/sql/driver debug/dwarf debug/elf debug/gosym debug/macho debug/pe encoding/ascii85 encoding/asn1 encoding/base32 encoding/base64 encoding/binary encoding/csv encoding/gob encoding/hex encoding/json encoding/pem encoding/xml errors expvar flag fmt go/ast go/constant go/doc go/format go/importer go/parser go/printer go/scanner go/token hash/adler32 hash/crc32 hash/crc64 hash/fnv html html/template image image/color image/draw image/gif image/jpeg image/png index/suffixarray io io/ioutil math math/big math/cmplx math/rand mime mime/multipart mime/quotedprintable net/http/cookiejar net/http/fcgi net/mail net/rpc/jsonrpc net/textproto net/url path path/filepath reflect regexp regexp/syntax sort strconv strings sync sync/atomic testing/quick text/scanner text/tabwriter text/template text/template/parse time unicode unicode/utf16 unicode/utf8
```
shurcooL: In the future (not this PR), let's wrap these kind of lines at 160 columns or something reasonable, to make future diffs easier to visualize. Does that sound good @neelance?

neelance: Yep.

neelance:
LGTM

doc/packages.md
```diff
@@ -70,12 +70,15 @@ On each commit, Circle CI automatically compiles all supported packages with Gop
 | go              |                       |                                   |
 | -- ast          | yes                   |                                   |
 | -- build        | no                    |                                   |
+| -- constant     | yes                   |                                   |
 | -- doc          | yes                   |                                   |
 | -- format       | yes                   |                                   |
+| -- importer     | yes                   | (TODO: unconfirmed, no tests)     |
```
shurcooL: Before merging, let's resolve this TODO.

shurcooL: Well, just looking at [its source](http://gotools.org/go/importer?tests=) is all that was needed.

It imports `go/types`, `runtime`, `go/internal/gcimporter` and `go/internal/gccgoimporter`.

Importing `go/types` is enough to make it not work, since `go/types` tests fail.

PR #282, opened 2015-08-29
don't run main function when testing (fixes #277)

### LINKED ISSUE #277: gopherjs test runs main.main() during tests, it shouldn't be.
When running `gopherjs test`, `main.main()` is executed.

Sample Go package: [`github.com/shurcooL/play/145`](http://gotools.org/github.com/shurcooL/play/145?tests=).

Reproduce with:

```bash
go get -u github.com/shurcooL/play/145
go test -v --bench=. github.com/shurcooL/play/145
gopherjs test -v --bench=. github.com/shurcooL/play/145
```

Expected output, as with Go:

```
$ go test -v --bench=.
=== RUN   TestDoNothingForASecond
--- PASS: TestDoNothingForASecond (1.00s)
PASS
BenchmarkDoNothingForASecond-8	       1	1000238355 ns/op
ok  	github.com/shurcooL/play/145	2.028s
```

Actual output, with GopherJS:

```
$ gopherjs test -v --bench=.
/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:1429
      throw err;
            ^
Error: we should not get here during a test
    at $callDeferred (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:1360:17)
    at $panic (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:1399:3)
    at $packages._/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/shurcooL/play/145.main (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:21211:3)
    at Object.$packages._/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/shurcooL/play/145.$init (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:21247:3)
    at $packages.main.$init (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:27000:13)
    at $goroutine (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:1419:19)
    at $schedule (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:1473:5)
    at $go (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:1449:3)
    at Object.<anonymous> (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:27016:1)
    at Object.<anonymous> (/private/var/folders/tw/kgz4v2kn4n7d7ryg5k_z3dk40000gn/T/test.975862829:27019:4)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3
FAIL	_/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/shurcooL/play/145	0.337s
```

neelance:


shurcooL:
I can confirm this fixes the reproduce case I've provided.

shurcooL:
If it's not too hard, it might be nice to add a regression test for this, since in most cases running `main()` during a test may not be as immediately obvious but lead to strange side-effects.

Feel free to use or base it on [`github.com/shurcooL/play/145`](http://gotools.org/github.com/shurcooL/play/145?tests=) if that's helpful.

If you feel confident this won't break in the future or this test doesn't belong here, it's fine not to add it.

shurcooL:
I am not familiar with the context of the fix, and I don't completely understand the change, so take that into account.

I don't see anything obviously wrong with it, tests pass and the issue is resolved, so it LGTM.

neelance:
Test added.

Can I tell you more about the change? The problem was that I assumed that you can not depend on a `main` package, which is usually the case. Only that testing creates an additional fake `main` package which depends on the `main` package being tested. The `$init` function or a package used to also run the `main` function and this needs to stay there because of code that needs to be generated to support blocking `init` functions. So the solution that I chose is to store the package that needs to run the `main` function in the global `$mainPkg` variable and add an if statement around the call to `main` that checks if the current package being initialized is really the correct main package.

shurcooL:
That was very helpful indeed, thanks for the explanation.

tests/main/main_test.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+import "testing"
+
+func TestMain(t *testing.T) {
```
shurcooL: @neelance, are you aware that this name `TestMain` has special meaning as of Go 1.4?

> The `testing` package has a new facility to provide more control over running a set of tests. If the test code contains a function
> 
> ```
> func TestMain(m *testing.M) 
> ```
>
> that function will be called instead of running the tests directly. The M struct contains methods to access and run the tests.

Source: https://golang.org/doc/go1.4

I kinda suspect it's an unintentional usage of that feature here, since you're testing a feature named main. If so, I think it's better we rename it to avoid doing something unintentionally just due to a coincidence in naming, as it might cause confusion later (if other tests are ever added here, they will not run). Perhaps `func TestNotRunMain()` or so?

shurcooL: @neelance, did you see this?

neelance: I did not know that. Thanks for pointing it out. Fixed via 01e6396.

shurcooL: Great, thanks!

Issue #284, opened 2015-08-29
Proposal: Move towards a more flexible, low level compiler testing system like that of Go.


shurcooL:
#### Motivation

When implementing tests for https://github.com/gopherjs/gopherjs/issues/279, I ran into a difficulty in that I needed my test program to import "time" package but not have any references to `time.Time` so that it would get DCEed. Normally, it's easy to do:

```Go
package main

import "time"

func main() {
	time.Sleep(time.Second)
}
```

However, when I tried to make it a test, that requires importing "testing" package. But the problem is that package already imports "time" and references `time.Time` internally, causing it not get DCEed.

#### Proposal

Let's move towards using a more flexible, low level testing system like is used in internal Go compiler tests (https://github.com/golang/go/tree/master/test folder) for compiler tests.

Using "testing" package is a very high level, large tool. It's great for package tests. But not if you want to test various compiler-level aspects. For example, you might want to test that something fails to compile, or causes a panic, etc.

#### Plan

If accepted, my implementation plan is to convert some of the existing Go compiler tests to run under GopherJS compiler. I was successful in getting 230~ passes and 40 fails. I can triage and skip the failing tests, and add the passing ones to GopherJS test suite.

That naturally paves the way to reuse that existing system for additional GopherJS specific compiler tests. That system has various actions like "rundircmpout", "compile", "compiledir", "build", "runoutput", "rundir", "errorcheck", "errorcheckdir", "errorcheckoutput", "run", "cmpout", "skip" which make it flexible and easy to add specific tests.

#### Closing Remarks

This is not a high priority and doesn't need to be done urgently, rather I propose we naturally move in this direction.

shurcooL:
PR #296 is the first step in this direction.

neelance:
#296 was a really good step in this direction. Thanks for that. Do you think we should keep this issue open and add a checklist for further steps or shall we close it and first triage the failures you spotted and then go from there?

shurcooL:
I'm okay with marking this proposal as "accepted" and closing it perhaps (or how else to deal with accepted proposals?). It could also be kept open until it's done, but that may mean it'll be open for a long time, since this will be a gradual process.

My plan is to port more of the Go test-suite when the list of known unresolved failures starts getting closer to 0. It is also my plan to use this new way to run low level compiler tests if I need to add future compiler tests.

shurcooL:
This should also be included for #331.

PR #285, opened 2015-08-29
Fix possible panic during internalization/externalization of time.Time/Date.

### LINKED ISSUE #279: Error with time.Sleep
if use gopherjs with time.Sleep without any reference to time.Time structure, this code:
```go
package main

import (
    "time"

)

//var _ = time.Now()

func log(i ...interface{}) {
    js.Global.Get("console").Call("log", i...)
}

type Test struct {
    A int
    B string
}

func main() {
    log(Test{1, "hello"})
    time.Sleep(time.Second)
}
```

produces error in browser:
```
go-js.js:1738: Uncaught TypeError: Cannot read property 'ptr' of undefined
```
in function $externalize
```js
    if (timePkg && v.constructor === timePkg.Time.ptr) {
```
timePkg.Time is not defined in javascript code

shurcooL:
Begin by adding a failing test for issue #279.

Proper testing for this issue prevents "testing" package from being imported. So, implement it as a low level single file .go test and invoke it from backend. (See #284 for related context.)

In the `time_inexternalization.go`, you can use these statements for
manual testing:

```Go
var _ = time.Now()      // Force time.Time not to be DCEed.
var _ = time.Unix(0, 0) // Force time.Unix (and time.Time implicitly) not to be DCEed.
```

Adding both of those statements (or just `time.Unix` one, since it also includes `time.Time` indirectly) would cause the test to pass. But a true fix will be in the compiler itself, in a way that will avoid the panics/incorrect behavior even when `time.Time` and/or `time.Unix` are DCEed, or time package is not imported at all.

**Edit:** Implemented fix for the issue, so it now resolves #279.


shurcooL:
Test is currently failing with:

```
lowlevel_test.go:20: exec: "gopherjs": executable file not found in $PATH:
```

@neelance, since you're more familiar with Circle CI (I am more familiar with Travis), do you know how we can make it so the compiled `gopherjs` binary is added to PATH?

shurcooL:
I've implemented a fix for issue #279 that successfully resolves the failing test, please see and review #286.

neelance:
Looks like a good test for the issue. LGTM

shurcooL:
This is ready to merge now and resolves the issue fully. LGTM.

shurcooL:
(I've reworded the second last commit to have a slightly better commit message. It went from "Internalize Date when time package not imported in a reasonable way." to "Internalize Date in a reasonable way when time package not imported.")

PR #286, opened 2015-08-30
Implement possible solution 2 fix for #285.

shurcooL:
This is an implementation of possible solution 2, with modifications based on additional findings.

The reason I went with 2 instead of 1 as originally planned is described in https://github.com/gopherjs/gopherjs/issues/279#issuecomment-136095365.

See commit messages for full details.

This should make all tests pass.

### DIFF
--- a/compiler/natives/time/time.go
+++ b/compiler/natives/time/time.go
@@ -9,6 +9,13 @@ import (
 	"github.com/gopherjs/gopherjs/js"
 )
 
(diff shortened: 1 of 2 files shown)


neelance:
LGTM. I would however prefer to negate the expression to `if (timePkg === undefined) {` and don't use an `else` clause. That would highlight the fact that the normal execution path is to return a `time.Time` and the if clause catches the special case that the time package is not available.

shurcooL:
That sounds like a good idea, I will make that change. Thanks!

PR #288, opened 2015-08-31
Fix internalize case for Date into time.Time struct.

### LINKED ISSUE #287: Date to time.time causes Uncaught Error: cannot internalize time.Time
According to the docs (if I'm reading them correctly) I can expose a method to Javascript and call it using MakeWrapper. GopherJS will take care of the conversion for the arguments within the exposed methods. (http://godoc.org/github.com/gopherjs/gopherjs/js#MakeWrapper)

However I can't seem to get a date object converted in to time.Time 

```
package main

import "time"

func main() {
	js.Global.Set("pet", map[string]interface{}{
		"New": New,
	})
}

type Pet struct {
	birthday time.Time
}

func New(date time.Time) *js.Object {
	return js.MakeWrapper(&Pet{date})
}

func (p *Pet) Birthday() time.Time {
	return p.birthday
}

func (p *Pet) SetBirthday(date time.Time) {
	p.birthday = date
}
```
Playground link: http://www.gopherjs.org/playground/#/JEcteZCXUP

When I call dog.New(new Date()) I get
Uncaught Error: cannot internalize time.Time

Am I missing something here?

shurcooL:
~~Don't merge, this PR is a WIP (more commits to be pushed) and in need of review.~~ Ready for review.

Add a failing test case, then fix #287.

### DIFF
--- a/compiler/prelude/jsmapping.go
+++ b/compiler/prelude/jsmapping.go
@@ -181,6 +181,13 @@ var $internalize = function(v, t, recv) {
   if (t === $jsObjectPtr.elem) {
     $panic(new $String("cannot internalize js.Object, use *js.Object instead"));
   }
+  var timePkg = $packages["time"];
+  if (timePkg !== undefined && t === timePkg.Time) {
+    if (!(v !== null && v !== undefined && v.constructor === Date)) {
+      $panic(new $String("cannot internalize time.Time from " + typeof v + ", must be Date"));
+    }
+    return timePkg.Unix(new $Int64(0, 0), new $Int64(0, v.getTime() * 1000000));
+  }
   switch (t.kind) {
   case $kindBool:
     return !!v;
@@ -270,12 +277,11 @@ var $internalize = function(v, t, recv) {
     case Boolean:
       return new $Bool(!!v);
     case Date:
-      var timePkg = $packages["time"];
       if (timePkg === undefined) {
         /* time package is not present, internalize as &js.Object{Date} so it can be externalized into original Date. */
         return new $jsObjectPtr(v);
       }
-      return new timePkg.Time(timePkg.Unix(new $Int64(0, 0), new $Int64(0, v.getTime() * 1000000)));
+      return new timePkg.Time($internalize(v, timePkg.Time));
     case Function:
       var funcType = $funcType([$sliceType($emptyInterface)], [$jsObjectPtr], true);
       return new funcType($internalize(v, funcType));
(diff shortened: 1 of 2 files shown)


shurcooL:
As desired, commit 098b579 added a failing test case and it indeed [failed](https://circleci.com/gh/gopherjs/gopherjs/551) with:

```
--- FAIL: TestInternalizeDate (0.01s)
/tmp/test.592009139:4
return;}$goroutine.exit=true;}catch(err){$goroutine.exit=true;throw err;}final
                                                                    ^
Error: cannot internalize time.Time
    at $callDeferred (/tmp/test.592009139:4:27597)
    at $panic (/tmp/test.592009139:4:28221)
    at $b (/tmp/test.592009139:27:44660)
    at $callDeferred (/tmp/test.592009139:4:27768)
    at $panic (/tmp/test.592009139:4:28221)
    at $internalize (/tmp/test.592009139:4:39702)
    at v.$externalizeWrapper (/tmp/test.592009139:4:35661)
    at eval (eval at <anonymous> (/tmp/test.592009139:28:16346), <anonymous>:1:18)
    at eval (native)
    at $packages.github.com/gopherjs/gopherjs/js_test.AE [as F] (/tmp/test.592009139:28:16346)
    at $packages.testing.BN (/tmp/test.592009139:27:44958)
    at $goroutine (/tmp/test.592009139:4:28817)
    at $runScheduled [as _onTimeout] (/tmp/test.592009139:4:29579)
    at Timer.listOnTimeout (timers.js:119:15)
FAIL	github.com/gopherjs/gopherjs/js	0.503s
```

The next commit will contain a fix that should get the failing test case to pass.

compiler/prelude/jsmapping.go
```diff
@@ -313,6 +313,13 @@ var $internalize = function(v, t, recv) {
     }
     return s;
   case $kindStruct:
+    if (v !== null && v !== undefined && v.constructor === Date) {
```
shurcooL: There's probably a better way to do this instead of `v !== null && v !== undefined`, isn't there?

I needed it to avoid:

- TypeError: Cannot read property 'constructor' of null
- TypeError: Cannot read property 'constructor' of undefined

neelance: You can do `!v`, but that matches a few more cases. I'd go with the solution you have.

shurcooL: > I'd go with the solution you have.

Sounds good.

shurcooL:
Commit 3844adbd682c6886fb5541fd147161c04c04088a fixed the failing test case.

This PR is ready for full review now.

neelance:
Please see my comment on #287.

shurcooL:
I've resolved that. What else does this PR need to be merged?

neelance:
LGTM

Issue #119, opened 2014-11-04
Better error output if error is in dependency


neelance:

shurcooL:
I do wish there was more detail here though... :(

All I have to go on is the commit message:

> only show error of import if import failed

Can you easily think of a way I can reproduce this issue (if I temporarily revert the commit that fixed it) so I can see the before and after output?

shurcooL:
I was able to figure out what this issue was about.

Given the following incorrect program that has errors and failures to import packages (because they're missing):

```Go
package main

import "foo/bar"
import "doesnt/exist"

func main() {
	bar.Do(exist.Nothing)
	f()
}
```

The output of `go build` shows only import error messages:

```
go build
main.go:4:8: cannot find package "doesnt/exist" in any of:
	/usr/local/go/src/doesnt/exist (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/doesnt/exist (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/doesnt/exist
main.go:3:8: cannot find package "foo/bar" in any of:
	/usr/local/go/src/foo/bar (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/foo/bar (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/foo/bar
```

Output of `gophers build`, before this patch, shows those import errors in addition to other errors.

```
main.go:3:8: could not import foo/bar (cannot find package "foo/bar" in any of:
	/usr/local/go/src/foo/bar (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/foo/bar (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/foo/bar)
main.go:4:8: could not import doesnt/exist (cannot find package "doesnt/exist" in any of:
	/usr/local/go/src/doesnt/exist (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/doesnt/exist (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/doesnt/exist)
main.go:7:2: undeclared name: bar
main.go:7:9: undeclared name: exist
main.go:8:2: undeclared name: f
```

[...]

PR #291, opened 2015-09-01
compiler: Improve display of import errors.

### LINKED ISSUE #290: "cannot find package" errors could be more informative
Consider this error from `gopherjs build`:

    cannot find package "github.com/flimzy/flashback/pages" in any of:
            /usr/local/go/src/github.com/flimzy/flashback/pages (from $GOROOT)
            /home/jonhall/go/src/github.com/flimzy/flashback/pages (from $GOPATH)

It's *almost* identical to the same error from `go build`:

    cmd/server/server.go:4:2: cannot find package "oink" in any of:
            /usr/local/go/src/oink (from $GOROOT)
            /home/jonhall/go/src/oink (from $GOPATH)

Except that the `go` version tells me where the missing dependency is referenced ("cmd/server.go:4:2").

It would be very handy if GopherJS did the same thing :)

shurcooL:
If there are multiple import errors, display up to 10 of them.

Include the filename and line number where the error occurred, just as go tool does. Resolves #290 (which was a regression caused by #119).

#### Effect

Sample error output before:

```
$ gopherjs build 
cannot find package "foo/bar" in any of:
	/usr/local/go/src/foo/bar (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/foo/bar (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/foo/bar
```

After this PR:

```
$ gopherjs build 
main.go:3:8: could not import foo/bar (cannot find package "foo/bar" in any of:
	/usr/local/go/src/foo/bar (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/foo/bar (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/foo/bar)
main.go:4:8: could not import doesnt/exist (cannot find package "doesnt/exist" in any of:
	/usr/local/go/src/doesnt/exist (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/doesnt/exist (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/doesnt/exist)
```

For comparison, output of `go build`:

```
$ go build
main.go:4:8: cannot find package "doesnt/exist" in any of:
	/usr/local/go/src/doesnt/exist (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/doesnt/exist (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/doesnt/exist
main.go:3:8: cannot find package "foo/bar" in any of:
	/usr/local/go/src/foo/bar (from $GOROOT)
	/Users/Dmitri/Dropbox/Work/2013/GoLanding/src/foo/bar (from $GOPATH)
	/Users/Dmitri/Dropbox/Work/2013/GoLand/src/foo/bar
```

#### Implementation Notes

Unfortunately, it required me to add more code rather than remove existing code (e.g., by reverting https://github.com/gopherjs/gopherjs/commit/642a7f3fafe4dc7c01179a830271aadd54641fdd). I hope we can eventually be able to reuse `go` tool with `gopherjs` compiler and not have to worry about/maintain this code. So if this works okay, it should be good enough until then.

#### `go1.5gotypes` Branch

If this PR is merged into `master`, I'll need to merge it to `go1.5gotypes` branch as well.

### DIFF
--- a/compiler/package.go
+++ b/compiler/package.go
@@ -76,16 +76,15 @@ func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, impor
 		Scopes:     make(map[ast.Node]*types.Scope),
 	}
 
-	var importError error
+	var importError bool
+	var importErrorList ErrorList
 	var errList ErrorList
 	var previousErr error
 	config := &types.Config{
 		Packages: importContext.Packages,
 		Import: func(_ map[string]*types.Package, path string) (*types.Package, error) {
 			if _, err := importContext.Import(path); err != nil {
-				if importError == nil {
-					importError = err
-				}
+				importError = true // Next error is an import error.
 				return nil, err
 			}
 			return importContext.Packages[path], nil
@@ -93,17 +92,23 @@ func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, impor
 		Sizes: sizes32,
 		Error: func(err error) {
 			if previousErr != nil && previousErr.Error() == err.Error() {
+				importError = false
 				return
 			}
-			errList = append(errList, err)
+			if importError {
+				importErrorList = append(importErrorList, err)
+				importError = false
+			} else {
+				errList = append(errList, err)
+			}
 			previousErr = err
 		},
 	}
 	typesPkg, err := config.Check(importPath, fileSet, files, typesInfo)
-	if importError != nil {
-		return nil, importError
+	if len(importErrorList) > 0 { // If there are any import errors, don't show other errors (issue #119).
+		errList = importErrorList
 	}
-	if errList != nil {
+	if len(errList) > 0 {
 		if len(errList) > 10 {
 			pos := token.NoPos
 			if last, ok := errList[9].(types.Error); ok {


neelance:
The changes look reasonable. Could you please also check the case that there is an error in a dependency, not in the main package? Because there was the case that it was showing something like `main.go:4:8: foo.go:7:2: error message`.

shurcooL:
Good idea.

I tried this invalid Go program:

```Go
package main

import "fmt"

func main() {
	fmt.Fprintln("Hello")
}
```

The output on this branch and on `master` is identical:

```
$ gopherjs build
main.go:6:15: cannot pass argument "Hello" (constant of type string) to parameter of type io.Writer
```

neelance:
That's not yet a test for an error in a dependency, right?

shurcooL:
It's not. I misread your comment, my mistake.

I'll have to test that later, I'm not on a computer now.

shurcooL:
You were right, it doesn't handle it well when there's a build error in a dependency. Compare:

```
31 $ go build
# github.com/shurcooL/htmlg
../../htmlg/htmlg.go:35: a.AppendChildx undefined (type *"golang.org/x/net/html".Node has no field or method AppendChildx)

31 $ gopherjs build # master
../../htmlg/htmlg.go:35:2: invalid operation: a (variable of type *golang.org/x/net/html.Node) has no field or method AppendChildx

31 $ gopherjs build # this branch
main.go:6:2: could not import github.com/shurcooL/htmlg (/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/shurcooL/htmlg/htmlg.go:35:2: invalid operation: a (variable of type *golang.org/x/net/html.Node) has no field or method AppendChildx)
```

I'll work on fixing that.

shurcooL:
I'm also seeing different behavior depending if you do `gopherjs build` vs. `gopherjs build full/import/path`. It's really odd. This is not an easy fix.

shurcooL:
Files the above as https://github.com/gopherjs/gopherjs/issues/312, and it'd probably need to be resolved before this issue. However, I'm considering closing this PR because I hope we'll eventually be able to use `go` tool and this doesn't seem to be worth spending much time on (because it's tricky to fix and requires a lot of manual regression testing).

neelance:
@shurcooL Have you any specific plan on how to use the `go` tool with GopherJS? How feasible is it?

shurcooL:
I think it's a potential proposal that has a chance of being accepted, on the grounds that currently `go` tool supports 2 Go compilers, and it'd be more general if it didn't have to hard-code them, instead allowing other compilers to be supported if they follow the same schema.

I also know there's quite a bit of hard-coded flags that have gccgo in the names, so it might be hard to outright replace those with more general compiler-agnostic flags or env vars. It might be rejected because of that.

I haven't yet looked into the technical details of how hard, how many changes it would be to make `go` support arbitrary compilers. I suspect it'd be a pretty significant undertaking. But I also think it may happen within the next few years, not necessarily very soon.

neelance:
Given that timeframe, it would still be nice to find a short-term solution for #290.

neelance:
@shurcooL This PR is stale, I'd like to close it. We still have #290 to keep track of the issue.

shurcooL:
Yeah, we can close it. I gave up. When I tried it, it was quite difficult to create a solution that would output optimal error cases in every situation, and I didn't have an automated way of testing different setups, so it was like whack-a-mole, improving errors for one situation would potentially mess them up for another.

By now, trying to resolve this issue from scratch (with this PR as some reference) will be about as hard as trying to pick up this PR.

I do think the best way to resolve this problem will be to start adding some test cases for error messages in order to avoid regressions, and to try to resolve the issue at a lower level than trying to fix it up at higher level.

neelance:
> whack-a-mole

;-)

Issue #294, opened 2015-09-04
Is there any reason not to include wrappers around JS builtins in gopherjs/js?


flimzy:
My apologies if this has already been discussed, but my search of the issue archives, and google, found nothing.

I know Go has its own libraries to handle everything in the [Builtin JS functions](http://www.w3schools.com/jsref/jsref_obj_global.asp) (with the exception of "eval", I suppose), but and that it's also pretty simple to call the JS functions inline:

    encoded := js.Global.Call("encodeURI", "http://foo.com/?foo=bar")

But as light weight as these wrappers would be, why not include them in the standard js package?  I'd rather call `js.EncodeURI(...)` than pull in a dependency on `net/url`, for instance, or define my own wrapper somewhere else.

Is there a reason it was decided not to include these in the standard JS library?

If not, i would be more than happy to submit a PR to add these.  Or if for some reason we don't want to pollute the `js` namespace, perhaps a secondary module would be in order (`jsbuiltin`??).

Thoughts?

shurcooL:
Hi @flimzy,

Thanks for opening the issue and the PR (which contains excellent code)!

I think it's best to think of the `github.com/gopherjs/gopherjs/js` package as a very general interface to the JavaScript world. It's akin to the `reflect` package or `C` pseudo-package, needed only to access another world.

From that perspective, `github.com/gopherjs/gopherjs/js` is not a package standard JS library.

I think its goal is to be as simple and general as possible, and its API should be kept as simple as possible in order to be easier to use.

The wrappers your proposing are a great idea, but I don't think they belong in `github.com/gopherjs/gopherjs/js` package.

I suggest one of two things:

1. Make those wrappers into a Go library available, for example, at `github.com/flimzy/jsbuiltin`.

  Then, add a link to it on the Wiki page for [Bindings to JavaScript APIs and libraries](https://github.com/gopherjs/gopherjs/wiki/bindings).

2. We can consider moving that native JavaScript bindings libary under `gopherjs` organization, so it can have an import path like `github.com/gopherjs/jsbuiltin`, for example.

To conclude, I think that the GopherJS compiler and its `github.com/gopherjs/gopherjs/js` package should be as simple and minimal as possible. Their goal is to compile Go to JavaScript, and to allow accessing the JavaScript world from Go (and vice versa). All libraries, wrappers, etc., are great but belong in separate repos.

What do you think?

shurcooL:
[...]

PR #295, opened 2015-09-04
Add wrappers for some JavaScript built-in functions. Plus tests.

flimzy:
I'm submitting this PR for comment in relation to issue #294 .

Note that I did not wrap the following built in functions:

* **escape()** and **unescape()**: These were deprecated circa 2000.
* **eval()**:  I'm not sure if it would ever be a good idea to eval() JS code from within Go. Maybe, maybe not... but it's not obvious so I'm leaving it out for now.
* **Number()**: This requires handling a bunch of corner cases which don't exist in a strictly typed language such as Go. The impedance mismatch between JS and Go seems too high to be meaningful to wrap this, plus, the parse*() functions handle most of the meaningful use cases anyway. Anyone with a legitimate need to use `Number()` probably need to write their own wrapper to handle the variations that matter to them.
* **parseInt()** and **parseFloat()**: These are already wrapped as *js.Object methods. It probably makes sense to wrap them as global functions, too, but how should NaN be handled? The object method (I'm guessing) returns an uninitialized number (i.e. 0), which doesn't seem right.  Returning, for example, either `int,error` or `int,bool` (a la type reflection), seem to make sense to me. Returning an error seems more idiomatic, but would require adding a dependency on 'errors' (would this be bad?)

### DIFF
--- a/jsbuiltin/jsbuiltin_test.go
+++ b/jsbuiltin/jsbuiltin_test.go
@@ -2,7 +2,6 @@ package jsbuiltin
 
 import (
 	"testing"
-//	"github.com/gopherjs/gopherjs/jsbuiltin"
 )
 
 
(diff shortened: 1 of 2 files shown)


shurcooL:
With the understanding this won't be merged into `github.com/gopherjs/gopherjs/js` package, I have the following comments:

1. The tests are great.

2. Exported identifiers [should have docs](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments).

3. I can't provide additional comments because I haven't use and have no plans to use these funcs. I prefer to use `net/url` and other standard Go funcs that do the same thing and work both in backend and frontend, given the option.

jsbuiltin/jsbuiltin_test.go
```diff
@@ -0,0 +1,138 @@
+package jsbuiltin
```
shurcooL: The `jsbuiltin.go` file has a `// +build js` build tag. This `jsbuiltin_test.go` file should too, otherwise this package tests will fail with `go test`.

shurcooL:
These `.go` files are not `gofmt`ed. You should use `gofmt` or `goimports` tool to fix that.

flimzy:
This code now lives in its own repository, available here: https://github.com/gopherjs/jsbuiltin

Issue #303, opened 2015-09-12
`gopherjs test` + nodejs require == broken and possible security issue


flimzy:
I'm trying to write some tests for my PouchDB bindings, which naturally require the node pouchdb library. So I was trying this:

    GlobalPouch := js.Global.Call("require", "pouchdb")

But this fails when running `gopherjs test`, because in this case, the file runs from /tmp, and `npm` didn't install my dependencies in `/tmp` for obvious reasons. :)

For now I worked around the problem thusly:

    cwd := js.Global.Get("process").Call("cwd").String()
    GlobalPouch = js.Global.Call("require",  cwd + "/node_modules/pouchdb")

And my code is working.

But as I was considering the implications of this, I think it's more than a minor inconvenience. It could be a potential security concern, as a malicious user of the same system might put a phony 'pouchdb' library in /tmp, and the next time I run my test... 

A couple possible solutions I've thought of:

1. Simply have GopherJS modify node's runtime environment, changing `process.paths` to reflect the location of the Go source, rather than GopherJS's temporary output. If this is easy, it may be the best solution. To avoid the security issue described above, it ought to completely replace the existing `process.paths` values, not simply add to them.

2. Have GopherJS put its temp file in the same location as the .go files, perhaps with a name like `.test.1234567` in place of `/tmp/1234567`.  This might be preferable in the case that other modules (in Node *or* Go) try to do the same trick, and read files (executable code, or config files, etc) from a path relative to the running file.

flimzy:
I thought I would try my hand at this again today, and came to the conclusion that it's not really a straight-forward problem to solve.  I think some discussion will be necessary to decide how it should even be approached.

Imagine an app, `foo.go`, which depends on `a.go` and `b.go`.  `a.go` requires* JS module `a`, and `b.go` requires* JS module `b`. (This requiring might happen via `js.Global.Call("require", "a")`, or might be included in `*.inc.js` files--either should have the same result for this situation.)

The question is, where should node look for the `a` and `b` modules?

When I first filed this bug, my assumption was that it should look in the `node_modules/` directory of the GopherJS module's root.  But this would mean multiple directories in this scenario, and there's likely a good chance that some of them won't have `node_modules` populated anyway (at least there's no GopherJS infrastructure to insure that it is).

Assuming `a.go` and `b.go` are upstream modules, they would typically be installed via:

> go get -u -d -tags=js github.com/foo/a
> go get -u -d -tags=js github.com/bar/b

[...]

PR #307, opened 2015-09-14
Use the current working directory for temp files for 'run' and 'test'

flimzy:
I have tested this with respect to issue #303, and have found that it does solve the reported problem.

That in no way means it's the best solution, but it is a simple one.


flimzy:
I clearly spoke prematurely, since it breaks a bunch of other things.

Issue #319, opened 2015-10-08
Unable to correctly handle certain unicode/utf-8 characters.


shurcooL:
Go source code is Unicode text encoded in UTF-8. ([Source](https://golang.org/ref/spec#Source_code_representation).)

GopherJS fails to process some unicode characters in Go code and emits JavaScript that does not print the correct characters.

I think it fails on "astral planes", or unicode characters that require more than 2 bytes using utf-8 encoding. But this is just a hypothesis from limited evidence/investigation.

Compare:

```JavaScript
console.log("Hello, 世界. 😀!"); // U+1F600 emoji.
```

```Go
package main

import "fmt"

func main() {
	fmt.Println("Hello, 世界. 😀!") // U+1F600 emoji.
}
```

https://play.golang.org/p/M1rXP8bm6L

http://www.gopherjs.org/playground/

(You'll have to manually paste the code into GopherJS Playground. Due to the bug in GopherJS, the snippet cannot be loaded correctly.)

![image](https://cloud.githubusercontent.com/assets/1924134/10359664/88ea60f6-6d4f-11e5-8bc9-d2571767703e.png)

![image](https://cloud.githubusercontent.com/assets/1924134/10359669/9044189c-6d4f-11e5-8102-78c9091f12ec.png)

Thank you to @peggyl for the original report of this issue at https://gist.github.com/peggyl/72d47f6652d30a4f566a!

#### Relevant Reading

- http://www.fileformat.info/info/unicode/char/1f600/index.htm
- https://mathiasbynens.be/notes/javascript-unicode
- https://mathiasbynens.be/notes/javascript-escapes#unicode-code-point
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint

siongui:
I have a function:

```
func GetFirstCharacterOfWord(word string) string {
	runeValue, _ := utf8.DecodeRuneInString(word)
	return string(runeValue)
}
```

when this is compiled to JS by GopherJS (Go 1.12.17), and ḍ (%E1%B8%8D in Chromium console) is given as input, it will return %, which is obviously incorrect. I test the code in Go and it works fine.

Not sure where goes wrong. Please advice! Thanks.

Issue #320, opened 2015-10-08
Stance on ECMAScript 6?


shurcooL:
(I'm not an expert on this topic, so please correct me if I'm wrong.)

The README does not mention anything about ECMAScript 6 and as far as I know, either no, or very few of its features are being used in the generated code.

As far as I can tell, ECMAScript 6 has been officially released for over 3 months now, and latest stable versions of most browsers are supporting it by now (to various degree). See https://kangax.github.io/compat-table/es6/.

What is GopherJS's official stance on ECMAScript 6 features? If I make a PR, can I use some ECMAScript 6 features?

In my opinion, GopherJS is a relatively bleeding edge technology and assuming modern browser is expected; legacy support for extremely outdated browsers is a lesser concern.

### Proposal

I think the best attitude to have here is the following. **If some ECMAScript 6 feature that is commonly available on today's current stable channels of various browsers makes it possible to do something more efficiently and easily, it should be considered and possibly used.**

I don't think it's in the interest of GopherJS project to hold itself significantly back by attempting to support legacy browsers older than current stable channels, is it? (But if the improvements are very minor, then it may not be worth using new features.)

#### Current status

Is there a policy on this? How does my proposal above sound? Are there things that ECMAScript 6 could enable significant improvement in?

Right now, I'm learning about the internalization/externalization of strings (in order to fix bug #319), and looking at [ECMAScript 6: Unicode code point escapes](https://mathiasbynens.be/notes/javascript-escapes#unicode-code-point). I don't yet know for sure, but there's a chance being able to use them can lead to better generated output.

neelance:
Usually when you use ES6 features you have a compiler like Babel + polyfills to make it compatible with older browsers. Obviously that's not an option for GopherJS. I don't want to sacrifice compatibility just for some nicer syntax in generated files. If you find some major performance gain, then we can talk about it. Otherwise I'd like to stick with widely supported features.

shurcooL:
Thanks for the answer. I agree.

I'll close since the question is answered. Hopefully others can search and find this if they have the same question.

shurcooL:
> Hopefully others can search and find this if they have the same question.

#331 proposes having a Developers' Guidelines document, and this information can be moved there for easy access.

shelby3:
> If you find some major performance gain, then we can talk about it. 

[Here](https://github.com/gopherjs/gopherjs/issues/15#issuecomment-390494239).

[...]

PR #321, opened 2015-10-08
WIP: Fix unicode/utf8 issue.

shurcooL:
Do not merge. This is a WIP initial attempt to find a good fix for #319.

I have an initial attempt that seems to maybe kinda partially work to fix the exact reproduce case, but it needs to be significantly reviewed and generalized.

I think I have a test that correctly catches the issue.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 0.12.7
+    version: 4.1.2
 
 dependencies:
   pre:
--- a/circle.yml
+++ b/circle.yml
@@ -4,6 +4,8 @@ machine:
 
 dependencies:
   pre:
+    - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 20 # Temporary fix for node-gyp with Node.js 4.x.
+    - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 20 # Temporary fix for node-gyp with Node.js 4.x.
     - cd /usr/local && sudo rm -rf go && curl https://storage.googleapis.com/golang/go1.5.linux-amd64.tar.gz | sudo tar -xz && sudo chmod a+w go/src/path/filepath
   post:
     - mv ./gopherjs $HOME/bin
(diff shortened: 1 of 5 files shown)


shurcooL:
I am fairly confident in the test being correct.

The externalization case likely makes sense.

The internalization fix is a complete hack and most likely incorrect in the general case. I just pushed it for now, but will continue to look into it later.

At a high level, I'd like to learn more about the current strategy that GopherJS uses to internalize/externalize strings, and discuss some ideas for how ECMAScript 6 features may potentially be useful in improving it (related to #320).

shurcooL:
Well, both the externalization and internalization fixes broke a lot of tests, so they're clearly incomplete/incorrect. It does fix the original issue though:

image

They also resolve the `TestInternalizeExternalizeUnicodeString` test. But need to fix all other tests.

shurcooL:
Resolved in c67fada883858bf571530da0400bb7db3a8d52fe by adding support for UTF-8 surrogate pairs.

@neelance, I'd still like to do this sometime:

> At a high level, I'd like to learn more about the current strategy that GopherJS uses to internalize/externalize strings, and discuss some ideas for how ECMAScript 6 features may potentially be useful in improving it (related to #320).

Also, thoughts on updating to Node 4.x as done in this PR, since it is supported as of https://github.com/gopherjs/gopherjs/commit/d942555ec097c1b3260ed89fff69afc2ceb93bec?

neelance:
@shurcooL Internally, GopherJS uses JS strings, but with a "one character per byte" policy. This fits better to how most operations on strings work. We can adopt ES6 features if they improve performance and we can include a polyfill for them.

shurcooL:
Can you elaborate on "one character per byte" policy? I'm not quite sure what that means.

As an example, how would it represent the Go string `"Hello, 世界. 😀!"`?

In Go, that string is 20 bytes but 13 runes.

neelance:
That string would have 20 characters in the internal JS representation, one for each byte. So if `s` is this string, then `s.length === 20` holds.

shurcooL:
Oh, so "one JavaScript character per Go byte". Gotcha, thanks.

Issue #306, opened 2015-09-14
`gopherjs test` does not include *.inc.js files.


flimzy:
The following description is inaccurate. Jump to [here](https://github.com/gopherjs/gopherjs/issues/306#issuecomment-140523480) to see a proper explanation of the issue.

<strike>I don't know if this is a good idea, but I want to propose it for discussion, because it seems, at least at first glance, that it could make a few things easier for me.</strike>

<strike>I've read through issue #111 and the related commits. It appears (from reading 51e33b76d50c8d6648dd80c24db26a5a81a02fcf, as well as the observed behavior in testing) that this feature is only applicable when compiling the `main` package.</strike>

<strike>Would it be reasonable to do the same for library packages?</strike>

<strike>At present I'm looking at this in respect to gopherjs/jsbuiltin#2. The current solution I'm using there (an `eval()`, triggers warnings in browsers with CSP enabled).  So I was looking at including a simple `jsbuiltin.inc.js` that contained simply `function typeoffunc(x) { return typeof x }`.  But, since this is the `jsbuiltin` package, and not `main`, that feature isn't enabled.</strike>

<strike>I can also see some potential benefits when writing wrappers/bindings for existing JavaScript functions--it might be handy (if/when licensing permits) to include the relevant JavaScript code in the bundled Go package.</strike>

<strike>I haven't thought this through entirely. I'm not 100% convinced it's all a good idea... But I imagine you guys have thought about many more of the underlying implications than I have to know if this is a good idea or not.</strike>

shurcooL:
I don't see why not.

In fact, I would be inclined to think the only reason it's not already the case is because of a bug or unintended omission.

Unless I'm missing something and there was a reason this `.inc.js` thing was added for `main` packages only?

That said, this is another feature I do not plan to use much, if at all. But if others want to, I think they should be able to.

shurcooL:
> Unless I'm missing something and there was a reason this .inc.js thing was added for main packages only?

Now that I think about it more, it might be on purpose.

Adding additional JavaScript makes sense when building the final output, such as the case for a `main` Go package. But if it's done on a per-package level, it may cause conflicts between different packages, etc.

So, I probably take back what I said above.

flimzy:
Yes, that's true. For it to really work well, it would almost certainly have to be in some sort of a CommonJS-style closure. And trying to mix that with GopherJS doesn't sound fun.

flimzy:
[...]

PR #322, opened 2015-10-09
Include *.inc.js files in `gopherjs build` mode.

flimzy:
This simply copies the same functionality from ImportPackage(). Perhaps it
should be factored into a common function.

This appears to solve bug #306, but causes some test failures, which I will investigate later.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -309,15 +309,11 @@ func (s *Session) BuildDir(packagePath string, importPath string, pkgObj string)
 	}
 	pkg := &PackageData{Package: buildPkg}
 	pkg.ImportPath = "main"
-	files, err := ioutil.ReadDir(pkg.Dir)
+	jsFiles, err := jsFilesFromDir(pkg.Dir)
 	if err != nil {
 		return err
 	}
-	for _, file := range files {
-		if strings.HasSuffix(file.Name(), ".inc.js") && file.Name()[0] != '_' {
-			pkg.JsFiles = append(pkg.JsFiles, file.Name())
-		}
-	}
+	pkg.JsFiles = jsFiles
 	if err := s.BuildPackage(pkg); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


neelance:
The code seems to be copied from `ImportPackage`. Would probably be good to put it into a new function.

flimzy:
Yesterday on my work machine, I was getting some test failures. On my personal laptop, the tests hang, and never complete. However, CircleCI seems to think the tests all pass. So... I'm not entirely sure what to make of all that.

I'm also not sure if/where I should add tests for this change.

So in summary, if someone else is confident that this is passing tests, it can be merged. If I should be adding tests, are there existing tests for build.go that I can add to? Or should I create some, and if so where?

flimzy:
(How) should this change be tested?

Perhaps add gopherjs/jsbuiltin to the list of packages auto built/tested (since it includes a *.inc.js file)?

Should I also add a build/build_test.go to test my new function, or is unit testing in this sort of case not a priority for this project?

flimzy:
I've done as much as I know to do with tests. Is there anything else you'd like me to do on this before it's merged?

build/build.go
```diff
@@ -574,6 +575,20 @@ func NewMappingCallback(m *sourcemap.Map, goroot, gopath string) func(generatedL
 	}
 }
 
+func jsFilesFromDir(dir string) ([]string, error) {
+	files, err := ioutil.ReadDir(dir)
```
shurcooL: I just wanted to comment here. I realize it is not a regression since the code contained `ioutil.ReadDir(pkg.Dir)` previously.

But this looks like it won't work in environments where direct disk access isn't available (e.g., the Go playground, or when using GopherJS as a library to compile in-memory without writing to disk).

It'd be outside the scope of this PR, but are there plans to try to make this operate/support virtual filesystems and avoid having duplicate code/logic for building with gopherjs-the-binary vs. gopherjs-the-library?

neelance:
I think it is unlikely that the tests for `jsFilesFromDir` will ever catch a regression, since the function is very straightforward and unlikely to change in the future. It would be more interesting to test it on the `BuildDir` and `ImportPackage` layer.

If we want to keep the current test case, then the directory `test-inc.js` should be renamed or moved to `testdata`, because that's the usual way to do it and Go skips those directories when looking for files to build: `Directory and file names that begin with "." or "_" are ignored by the go tool, as are directories named "testdata".` Also `foo.inc.js` should at least contain a comment.

For now, I would also be okay with just dropping the test and merge the implementation. The `build` package still lacks tests in general. This needs to be improved at some point, but not necessarily now. The compiler itself has a much higher risk of regressions.

flimzy:
Dropping tests seems quite reasonable to me, so I'll just remove that commit from the PR.

flimzy:
Done... I think this is ready for merging then.

build/build.go
neelance: Sure, if there is a need for it then we'll improve it. I haven't checked the playground yet, but I think it should not be affected by this change.

theclapp:
Is the inclusion of *.inc.js documented anywhere?  I didn't realize gopherjs did that, and spent awhile trying to figure out where it came from.

shurcooL:
> Is the inclusion of *.inc.js documented anywhere?

Yes, see https://github.com/gopherjs/gopherjs/wiki/Bundle-additional-JavaScript-with-Go-packages.

theclapp:
Thanks!

PR #330, opened 2015-10-18
Include *.inc.js files consistently

flimzy:
An alternative proposal to finish fixing #306, in response to comments on #328.

Moving the call to `jsFilesFromDir()` into `Import()` doesn't work with the current code layout, as `Import()` doesn't know anything about the `gbuild.Package` data structure, only the official Go `build.Package` structure.  So I instead moved it to `BuildPackage()`.

The only (potential) drawback to this, is that it means we include `*.inc.js` files completely consistently for all call methods.  The one place where this causes a change of behavior is when calling `gopherjs build <files>`.  Under the old behavior, executing `gopherjs build foo.go` would *not* include foo.inc.js in the final output. Under the new behavior, it will.  If the old behavior is important to preserve*, something more clever needs to be done (which I'm willing to tackle myself, too, if it's deemed important).

*The only way I can imagine the old behavior being useful is for testing... someone might want to choose which JS file(s), among many, to include at build time. But I don't know if that was actually the intention, or if this "feature" was unintentional.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -309,11 +309,6 @@ func (s *Session) BuildDir(packagePath string, importPath string, pkgObj string)
 	}
 	pkg := &PackageData{Package: buildPkg}
 	pkg.ImportPath = "main"
-	jsFiles, err := jsFilesFromDir(pkg.Dir)
-	if err != nil {
-		return err
-	}
-	pkg.JSFiles = jsFiles
 	if err := s.BuildPackage(pkg); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


neelance:
I considered this solution, but I think it breaks architecture too much. `Import` and `ImportDir` are meant for gathering all information about a package. Adding the code to `BuildPackage` adds functionality to the function which does not belong there. It might solve the current problem, but such solutions can hurt in the long term. You can already see right now that it breaks the `gopherjs build <files>` case and you would need to add a workaround for this case. Those workarounds tend to stack over time and become a burden. That's why I'd rather go for the bigger change and improve `Import`, which would in fact involve changing its return type.

flimzy:
Would you like me to work on that solution?  It sounds like you have a pretty good idea of what should happen, so if you have the time to tackle it, no doubt your time spent would be more efficient.  But I'm willing to try my hand at it if you want me to, and if it wouldn't actually be a bigger use of your time responding to my changes.

Issue #302, opened 2015-09-10
`gopherjs test ./with/relative/path` failing


flimzy:
The `gopherjs test` command fails when I specify an explicit path, while this works with standard `go test`.  A simple proof:

```
$ go test
PASS
ok      github.com/flimzy/flashback     0.003s
$ go test .
ok      github.com/flimzy/flashback     0.005s
```

Similar tests with gopherjs in another directory:

```
$ gopherjs test
PASS
warning: system calls not available, see https://github.com/gopherjs/gopherjs/blob/master/doc/syscalls.md
ok      github.com/flimzy/flashback/webclient/pages/all 0.319s
$ gopherjs test .
import ".": import relative to unknown directory
```

Of course testing '.' isn't that useful, that just proved a simple test case.  The same happens if I specify a full path:

```
$ go test ./couch/
ok      github.com/flimzy/flashback-server/couch        0.010s
```

And with gopherjs:

```
$ gopherjs test ./webclient/pages/all/
import "./webclient/pages/all/": import relative to unknown directory
```

shurcooL:
This is a known issue, was decided as "wontfix" at the time. See https://github.com/gopherjs/gopherjs/issues/148.

shurcooL:
Also, to correct your terminology, `go test ./couch/` is a relative import path. It's not a full import path. `go test github.com/foo/bar/couch` is a full import path. Relative import paths are those that contain `.` or `..` elements.

flimzy:
Thanks for the explanation.  I hadn't thought to try `github.com/.../`... that's good enough for me.

shurcooL:
I'm going to reopen and resolve this issue. /cc @slimsag

Sure, there's always the workaround of doing:

```
$ gopherjs test $(go list ./...)
```

But that's not nice and not friendly to new people who are used to being able to do things like `go test ./...`.

I've prototyped a fix. It's not the most trivial fix, but still very manageable. Will send a PR soon.

PR #332, opened 2015-10-18
README: Suggest working with Go packages by default.

shurcooL:
When writing Go code that is something more than just a very short snippet, it helps to organize code as a Go package. It may contain 1 or more .go files in a directory.

`go build` is primarily used with Go packages, as in `go build import/path`. We should suggest the same use of `gopherjs build`. That will make it easier for users to get started and grow their code, because they'll be able to split into multiple .go files without having to update their build step.

#### Note

It's still possible to use `gopherjs build` with a list of files, just like `go build` can be used in the same way. But I don't think we should be recommending that as the primary usage mode. Users that still want this can discover it via `gopherjs help build`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Get or update GopherJS and dependencies with:
 go get -u github.com/gopherjs/gopherjs
 ```
 
-Now you can use  `gopherjs build [files]` or `gopherjs install [package]` which behave similar to the `go` tool. For `main` packages, these commands create a `.js` file and `.js.map` source map in the current directory or in `$GOPATH/bin`. The generated JavaScript file can be used as usual in a website. Use `gopherjs help [command]` to get a list of possible command line flags, e.g. for minification and automatically watching for changes.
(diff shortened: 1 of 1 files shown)


neelance:
@shurcooL How do we continue on this one? Personally I used either `go build` or `go build [files]`, but nearly never `go build [package]`.

shurcooL:
I use `gopherjs build -m -o /path/to/output.js [import/path]` quite often to rebuild projects. But I think in context of GopherJS, `gopherjs build [files]` is much more common and useful than `go build [files]`. I think we should explain all 3 ways of doing it (or rather make it work exactly like the Go tool so people already familiar with that don't need to think), and let people use whatever form they prefer.

Perhaps to really resolve this issue, we'd need to resolve #302. So maybe we can just close this if we're not going to resolve #302 for now.

shurcooL:
Or how about just making this change:

> Now you can use  `gopherjs build [package]`, `gopherjs build [files]` or `gopherjs install [package]` which behave similar to the `go` tool.

shurcooL:
> Fixing relative import paths for the CLI should be easy.

That's what I meant by #302. Let's do that.

> Fixing them for import statements is more complicated.

As I said, this is fine to leave as wont-fix, it's a much less commonly needed feature.

neelance:
I'd be fine with `Now you can use gopherjs build [package], gopherjs build [files] or gopherjs install [package] which behave similar to the go tool.` We also can add support for relative import paths on the CLI.

shurcooL:
Ok, I'll update this PR to say that and we can merge it.

neelance:
Cool.

shurcooL:
Updated, PTAL.

PR #333, opened 2015-10-19
Set NODE_PATH environment variable

flimzy:
This is a proposed partial fix for #303.

This PR does not address the potential security issues, or do anything to handle node dependencies of Go dependencies.  But those issues can be addressed in separate PRs if this one is deemed appropriate for merge in the mean time.


neelance:
As mentioned in #303, the temporary files should be created in the current working directory to make `node` behave as expected. In this case we would not need this change, right?

flimzy:
#334 accomplishes the same thing with a different means, so closing this PR.

PR #334, opened 2015-10-19
Use current working directory, or GOPHERJS_TEMPDIR variable for temp files

flimzy:
This uses the current working directory for temp files, instead of /tmp, to solve #303.

This introduces a problem with the `go test -v -race ./...` command in circle.yml, because the current directory for the tests is unwritable.  So I also introduced an environment variable `GOPHERJS_TEMPDIR`, to optionally specify a temp directory in cases where one might build a package hosted in a non-writable directory.  Is this the best way?


neelance:
I think the issue with `go test -v -race ./...` is with the low-level tests. @shurcooL Can you help here? Is there some way to avoid the `GOPHERJS_TEMPDIR` variable?

tool.go
```diff
@@ -44,6 +45,10 @@ func init() {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
+	temporaryDirectory = os.Getenv("GOPHERJS_TEMPDIR")
+	if len(temporaryDirectory) == 0 {
```
shurcooL: `if temporaryDirectory == "" {` is more readable.

shurcooL:
> @shurcooL Can you help here? Is there some way to avoid the `GOPHERJS_TEMPDIR` variable?

I'll think about that and get back here. I'm sure we'll find a way. Maybe it's something that can go in a `TestMain` func.

flimzy:
Any further thoughts on this, @shurcooL ?

shurcooL:
Sorry for the delay.

I've looked more into this. I need to gain a better understanding of the problem and the tradeoffs of various solutions first, so I asked a bunch of questions in https://github.com/gopherjs/gopherjs/issues/303#issuecomment-158859817.

I'd also like to learn the implications of using cwd for temporary files (especially as a default behavior). It's something that feels bad and I'd rather avoid if possible, but maybe it's not that bad.

flimzy:
@neelance: After discussion on issue #303, it seems @shurcooL and I are in agreement that this is the best solution for now, and that the temp variable is probably a necessary evil, at least until something more ingenious is thought of. Does anything else need to be addressed on this issue before a merge?

flimzy:
I'll mention that it probably would be possible to avoid the temp variable if we're certain that nothing other than our low-level testing will ever need it. Maybe you can offer that assurance, @neelance? If so, I'd be happy to work on that patch.  If there's a chance something else might need it, exposing an environment variable seems the wiser course.

PR #335, opened 2015-10-19
Another try at handling *.inc.js cleanly

flimzy:
This PR attempts to follow @neelance's suggestion in #328 & #330, with some minor refactoring of build/build.go.

With this PR, `jsFilesFromDir()` is now called from `Import()`, the new `ImportDir()`, and `BuildDir()` (which seems to be a bit of an exception to the idea of using `Import*` to gather all the data used to build a package (not to mention the other exception of `ImportPackage()`).  I'm beginning to understand what Richard was referring to in his comment on #330.  And as such, in the grand scheme of things, this PR may make things worse rather than better... But it does seem to at least solve this immediate issue, so if a more proper fix is a long time off, I hope this might be considered for merge.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -43,7 +43,15 @@ func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
 	}
 }
 
-func Import(path string, mode build.ImportMode, installSuffix string, buildTags []string) (*build.Package, error) {
+func ImportDir(path string, mode build.ImportMode) (*PackageData, error) {
+	pkg,err := build.ImportDir(path, mode)
+	if err != nil {
+		return nil,err
+	}
+	return &PackageData{Package: pkg},nil
+}
+
+func Import(path string, mode build.ImportMode, installSuffix string, buildTags []string) (*PackageData, error) {
 	buildContext := NewBuildContext(installSuffix, buildTags)
 	if path == "runtime" || path == "syscall" {
 		buildContext.GOARCH = build.Default.GOARCH
@@ -87,7 +95,7 @@ func Import(path string, mode build.ImportMode, installSuffix string, buildTags 
 		}
 	}
 
-	return pkg, nil
+	return &PackageData{Package: pkg}, nil
 }
 
 // parse parses and returns all .go files of given pkg.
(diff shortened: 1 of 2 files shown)


neelance:
This looks really good. We should rename `Session.ImportPackage` to `BuildImportPath` to have less confusing names. And I'd like to move `ImportDir` below `Import` if you don't mind. The API is then still a bit messy, but definitely better than before.

flimzy:
The rename is done in 8c83af6 above, and I also re-ordered `Import` & `ImportDir` as you suggested. Let me know if you'd like me to address anything else in this PR.

build/build.go
```diff
@@ -87,7 +101,28 @@ func Import(path string, mode build.ImportMode, installSuffix string, buildTags
...
+	if err != nil {
+		return nil,err
+	}
+
+	return &PackageData{Package: pkg, JSFiles: jsFiles}, nil
+}
+
+// ImportDir is like Import but processes the Go package found in the named
+// directory.
+func ImportDir(path string, mode build.ImportMode) (*PackageData, error) {
```
neelance: `path` should be renamed to `dir`, just like with https://golang.org/pkg/go/build/#Context.ImportDir

flimzy:
I'm beginning to feel schizophrenic with the multiple PRs to address this same issue, and different attempts solving different portions of the bug.

As luck would have it, this PR doesn't solve the case of 'gopherjs test'... as the test case doesn't exercise any of the modified code in this PR.  The test `pkg` struct is populated in `tool.go`, rather than in `build.go`'s `Import()` or `ImportDir()`.

This leaves me with the immediately-obvious solution being, once again, to export `jsFilesFromDir()` for use in `tool.go` (near line 347).

Would it be cleaner to instead export a function to return a populated `build.PackageData` struct, and call it from tool.go?  Such a function could be used to do some DRY within `build.go` as well.  We already have `build.NewBuildContext()`, but it covers slightly different use cases.

neelance:
I think you just need to modify `collectTests` to also accept `PackageData` and then pass the JSFiles to only one of the two calls to `collectTests`.

flimzy:
A nice solution, @neelance, implemented in e5aa7c2.

neelance:
LGTM

neelance:
And thanks for contributing to the project!

build/build.go
```diff
@@ -87,7 +101,28 @@ func Import(path string, mode build.ImportMode, installSuffix string, buildTags
...
+
+	return &PackageData{Package: pkg, JSFiles: jsFiles}, nil
+}
+
+// ImportDir is like Import but processes the Go package found in the named
+// directory.
+func ImportDir(dir string, mode build.ImportMode) (*PackageData, error) {
+	pkg,err := build.ImportDir(dir, mode)
+	if err != nil {
+		return nil,err
```
shurcooL: I'm on mobile, but isn't there a `gofmt` issue here? Also line 122.

flimzy: You have a sharp eye. See PR #338.

Am I imagining things, or didn't this cause CI failures at some point in the past?  Regardless, should circle.yml be configured to fail for gofmt issues?

shurcooL: > You have a sharp eye.

Thanks. :)

> Am I imagining things, or didn't this cause CI failures at some point in the past?

I am guessing it might've done that, but it might've been lost in the transition from Travis CI to circle CI.

> Regardless, should circle.yml be configured to fail for gofmt issues?

Agreed, opened #339.

PR #338, opened 2015-10-21
gofmt fixes

flimzy:


shurcooL:
LGTM, thanks!

I'll follow this up with an issue and PR to add checking for gofmt issues to CI, so it can be caught automatically.

PR #344, opened 2015-10-25
CircleCI: Test for gofmt issues.

### LINKED ISSUE #339: CI: Should also check for gofmt issues, possibly go vet, etc.
It'd be better if issues like https://github.com/gopherjs/gopherjs/pull/335#discussion_r42596903 would be caught with CI for future PRs.

Most of my Go projects have a .travis.yml file that does it, e.g.:

https://github.com/gopherjs/snippet-store/blob/master/.travis.yml

I'm sure the same can be done with circle CI.

shurcooL:
Resolves #339.

(See https://circleci.com/gh/gopherjs/gopherjs/659 for CI results.)

It turns out that running `gofmt -d .` will traverse inside hidden folders, and I see no way of skipping that. But it seems mostly harmless. Still, if there are suggestions for how to improve it, please share.

## `go vet` Check

I've tried adding a `go tool vet .` check, but it fails because of a valid test that tests `panic` behavior and triggers "unreachable code" issue:

```
gopherjs $ go tool vet .
tests/goroutine_test.go:70: unreachable code
```

Any ideas how to get around that?

Given that:

```
$ go tool vet -help
Usage of vet:
	vet [flags] directory...
	vet [flags] files... # Must be a single package
```

The only thing I can think of is explicitly listing all directories and .go files in root directory:

```
go tool vet build
go tool vet compiler
go tool vet doc
go tool vet js
go tool vet node-syscall
go tool vet nosync
go tool vet tool.go
```

(It excludes `go tool vet tests`).

But I don't like that because it'll be out of date if a new directory is added or an existing one is removed... Any suggestions?


dominikh:
> It turns out that running gofmt -d . will traverse inside hidden folders, and I see no way of skipping that. But it seems mostly harmless. Still, if there are suggestions for how to improve it, please share

This should work:

```
gofmt -d */
```

`*` ignores hidden files, the trailing `/` makes it only look for directories. I'm not sure if this is POSIX or a bash extension. Both /bin/sh (bash in alleged POSIX mode) and dash support it. Drop the trailing slash if you want to match files in the same directory.

shurcooL:
Thanks for the idea!

I tried those:

```
gopherjs $ gofmt -d .
.gitreally/really.go:1:1: expected 'package', found 'EOF'
gopherjs $ gofmt -d *
LICENSE:1:1: expected 'package', found 'IDENT' Copyright
README.md:1:1: expected 'package', found 'IDENT' GopherJS
README.md:8:1: illegal character U+0023 '#'
README.md:8:2: illegal character U+0023 '#'
README.md:8:3: illegal character U+0023 '#'
README.md:8:22: illegal character U+003F '?'
README.md:11:1: illegal character U+0023 '#'
README.md:11:2: illegal character U+0023 '#'
README.md:11:3: illegal character U+0023 '#'
circle.yml:1:1: expected 'package', found 'IDENT' machine
circle.yml:2:3: expected ';', found 'IDENT' node
gopherjs $ gofmt -d */
```

`gofmt -d */` skips `tool.go` though. So maybe:

```
gofmt -d */ *.go
```

Should do the trick?

shurcooL:
Although it's worth pointing out that even with `gofmt -d */` it will recurse into hidden directories inside the root directories, so it's not a completely bulletproof approach. But in most cases the only hidden directory will be the `.git` at root anyway.

dominikh:
You could do this, if the CI environment allows it: 

```
find . -name "*.go" -not -path '*/\.*' -exec gofmt -d {} \;
```

This will run `gofmt -d` for each `*.go` file found in non-hidden directories as well as the current directory.

shurcooL:
I think keeping it as is is best for now. The warning is harmless, and I don't think the bash black magic is worth it.

This is ready to merge from my side if it LGTU @neelance.

shurcooL:
I forgot about the `go vet issue`. Fine, how about a little black magic then. Is this okay?

```
for f in */; do echo $f; done | grep -v tests/ | xargs go tool vet
go tool vet *.go
```

The first line runs `go tool vet` on all non-hidden directories in root, except "tests".

The second line runs `go tool vet` with the .go files in the root directory (making up a single Go package).

neelance:
I personally like the proposal of @dominikh, because I think we should not treat the toplevel directory any different than subdirectories. You could also use an exclude pattern for `find` in the `go vet` case. That way, both tools would be invoked in the same way.

shurcooL:
`gofmt` will work with random individual .go files, but I think `go vet` needs or benefits from working with Go packages rather than .go files.

shurcooL:
I'm quite happy with the current solution and it's an improvement over the previous state (not running gofmt nor go vet at all).

What does this PR need, concretely, to be merged?

neelance:
I thought that the `gofmt` issue still needed to be solved. But seems like traversing into hidden directories is not really a problem. Merging...

PR #352, opened 2015-11-12
include *.inc.js in seperate scope to avoid package variable shadowing 

### LINKED ISSUE #351: [FATAL ERROR] variable shadowing of the constructor functions in *.inc.js
A typical javascript module would contains the following boilerplate to support `AMD loading` and direct usage as a normal script tag:
```javascript
(function(factory) {
    //AMD
    if (typeof define === "function" && define.amd) {
        define(["knockout", "exports"], factory);
    //normal script tag
    } else {
        factory(ko);         // ko from global javascript namespace 
    }
})(function(ko, exports) {  // using ko as real argument
    var t = ko.observable(123);
    ko.testVar = "testVar in a.inc.js";
   // other function goes here
    console.log("in a.inc.js:", t());
});
```
When this code is in one *.inc.js files, the following `gopherjs js code` would be produced:
```javascript
$packages["github.com/Archs/js/examples/incTest/a"] = (function() {
	var $pkg = {}, $init, ko, js, Hello; // ko here is the imported golang package name, 
          // which shadows the real ko in global javascritp namespace
	(function() {
(function(factory) {
    //AMD
    if (typeof define === "function" && define.amd) {
        define(["knockout", "exports"], factory);
    //normal script tag
    } else {
        factory(ko);  // notice, variable shadowling happends here
    }
})(function(ko, exports) {    // notice here, ko is shadowed
    var t = ko.observable(123);
    ko.testVar = "testVar in a.inc.js";
    console.log("in a.inc.js:", t());
});

}).call($global);
ko = $packages["github.com/Archs/js/gopherjs-ko"];  // here is what variable ko is used for
js = $packages["github.com/gopherjs/gopherjs/js"];
... // more line to go
```

When you run the code above code, error happeds and the script execution aborts:
```
Uncaught TypeError: Cannot read property 'observable' of undefined   // this is the results of variable shadowing
```

The fact is that when an imported golang package has the same name as the variables of the entrance function contained in the `js moduels` of the `*.inc.js` files, it will result in this fatal error.

Since we'd better not modify the javascript library we included, would it be ok if the golang package names are mangled in `gopherjs generated code` :question:

Archs:
fix #351 

by adding field `compiler.Archive.JSDecls` to seperate included js code from the genereate js code and  then write the included js code just before gopherjs package declarations to avoid variable name shadowing.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -501,8 +501,9 @@ func (s *Session) BuildPackage(pkg *PackageData) error {
 		if err != nil {
 			return err
 		}
-		code = append(append([]byte("\t(function() {\n"), code...), []byte("\n\t}).call($global);\n")...)
+		pkg.Archive.IncJSCode = append(pkg.Archive.IncJSCode, []byte("\t(function() {\n")...)
 		pkg.Archive.IncJSCode = append(pkg.Archive.IncJSCode, code...)
+		pkg.Archive.IncJSCode = append(pkg.Archive.IncJSCode, []byte("\n\t}).call($global);\n")...)
 	}
 
 	if s.options.Verbose {
--- a/build/build.go
+++ b/build/build.go
@@ -496,17 +496,16 @@ func (s *Session) BuildPackage(pkg *PackageData) error {
 		return err
 	}
 
-	var jsDecls []*compiler.Decl
+	pkg.Archive.JSDecls = []*compiler.Decl{}
 	for _, jsFile := range pkg.JSFiles {
 		code, err := ioutil.ReadFile(filepath.Join(pkg.Dir, jsFile))
 		if err != nil {
 			return err
 		}
-		jsDecls = append(jsDecls, &compiler.Decl{
+		pkg.Archive.JSDecls = append(pkg.Archive.JSDecls, &compiler.Decl{
 			DeclCode: append(append([]byte("\t(function() {\n"), code...), []byte("\n\t}).call($global);\n")...),
 		})
 	}
-	pkg.Archive.Declarations = append(jsDecls, pkg.Archive.Declarations...)
 
 	if s.options.Verbose {
 		fmt.Println(pkg.ImportPath)
--- a/build/build.go
+++ b/build/build.go
@@ -496,15 +496,13 @@ func (s *Session) BuildPackage(pkg *PackageData) error {
 		return err
 	}
 
-	pkg.Archive.JSDecls = []*compiler.Decl{}
 	for _, jsFile := range pkg.JSFiles {
 		code, err := ioutil.ReadFile(filepath.Join(pkg.Dir, jsFile))
 		if err != nil {
 			return err
 		}
-		pkg.Archive.JSDecls = append(pkg.Archive.JSDecls, &compiler.Decl{
-			DeclCode: append(append([]byte("\t(function() {\n"), code...), []byte("\n\t}).call($global);\n")...),
-		})
+		code = append(append([]byte("\t(function() {\n"), code...), []byte("\n\t}).call($global);\n")...)
+		pkg.Archive.IncJSCode = append(pkg.Archive.IncJSCode, code...)
 	}
 
 	if s.options.Verbose {
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -37,6 +37,7 @@ type Archive struct {
 	Imports      []string
 	ExportData   []byte
 	Declarations []*Decl
+	JSDecls      []*Decl
 	FileSet      []byte
 	Minified     bool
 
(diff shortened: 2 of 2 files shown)


compiler/compiler.go
```diff
@@ -183,6 +184,11 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, minify bool, w
 			panic(err)
 		}
 	}
+	for _, d := range pkg.JSDecls {
```
neelance: Would it make sense to add a `(function() { ... })()` wrapper around it, so it can't interfere with the generated code?

build/build.go
```diff
@@ -496,17 +496,16 @@ func (s *Session) BuildPackage(pkg *PackageData) error {
...
 
-	var jsDecls []*compiler.Decl
+	pkg.Archive.JSDecls = []*compiler.Decl{}
 	for _, jsFile := range pkg.JSFiles {
 		code, err := ioutil.ReadFile(filepath.Join(pkg.Dir, jsFile))
 		if err != nil {
 			return err
 		}
-		jsDecls = append(jsDecls, &compiler.Decl{
+		pkg.Archive.JSDecls = append(pkg.Archive.JSDecls, &compiler.Decl{
```
Archs: Yes, It's done herer

build/build.go
```diff
@@ -496,17 +496,16 @@ func (s *Session) BuildPackage(pkg *PackageData) error {
...
 			DeclCode: append(append([]byte("\t(function() {\n"), code...), []byte("\n\t}).call($global);\n")...),
```
Archs: Yes, It's done in the `buid` package

neelance: You are right, I didn't see that.

build/build.go
```diff
@@ -496,17 +496,16 @@ func (s *Session) BuildPackage(pkg *PackageData) error {
 		return err
 	}
 
-	var jsDecls []*compiler.Decl
+	pkg.Archive.JSDecls = []*compiler.Decl{}
```
neelance: No need for initialization.

compiler/compiler.go
```diff
@@ -37,6 +37,7 @@ type Archive struct {
 	Imports      []string
 	ExportData   []byte
 	Declarations []*Decl
+	JSDecls      []*Decl
```
neelance: I think we could use a simple `[]byte` here instead of `[]*Decl` and rename it, maybe `IncJSCode`?

Archs:
[]byte is more simpler :+1:

compiler/compiler.go
```diff
@@ -183,6 +184,11 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, minify bool, w
 			panic(err)
 		}
 	}
+	if pkg.IncJSCode != nil {
```
neelance: I think we can drop this `if` statement, a `nil` slice should not write anything.

shurcooL: Nice idea. :+1:

build/build.go
```diff
@@ -496,17 +496,14 @@ func (s *Session) BuildPackage(pkg *PackageData) error {
...
 	for _, jsFile := range pkg.JSFiles {
 		code, err := ioutil.ReadFile(filepath.Join(pkg.Dir, jsFile))
 		if err != nil {
 			return err
 		}
-		jsDecls = append(jsDecls, &compiler.Decl{
-			DeclCode: append(append([]byte("\t(function() {\n"), code...), []byte("\n\t}).call($global);\n")...),
-		})
+		code = append(append([]byte("\t(function() {\n"), code...), []byte("\n\t}).call($global);\n")...)
+		pkg.Archive.IncJSCode = append(pkg.Archive.IncJSCode, code...)
```
shurcooL: Given than these may be relatively large byte slices, it may help to do this in a more performant way.

I think it'd be more efficient to keep appending to a single byte slice, rather than creating a temporary one.

E.g.:

```Go
pkg.Archive.IncJSCode = append(pkg.Archive.IncJSCode, []byte("\t(function() {\n")...)
pkg.Archive.IncJSCode = append(pkg.Archive.IncJSCode, code...)
pkg.Archive.IncJSCode = append(pkg.Archive.IncJSCode, []byte("\n\t}).call($global);\n")...)
```

I have not benchmarked it, and don't have numbers of how much it's faster or slower, but my intuition tells me is that it should be quite better.

neelance: Besides performance this looks a bit easier to read for me.

neelance:
LGTM, merging...

neelance:
@Archs Thanks for contributing to the project. :+1:

PR #361, opened 2015-12-19
Use CWD for temp files, and fall back to /tmp only for upstream tests.

### LINKED ISSUE #303: `gopherjs test` + nodejs require == broken and possible security issue

flimzy:
This is an alternative to #334 to fix #303.  It uses the CWD for temp files, but falls back to using /tmp only for files found in the system GOROOT path, which should not be vulnerable to the security risks of running nodejs code in /tmp.

If there are ever other situations where we need to run tests in read-only directories, this will fail.  I don't know if such cases are likely ever to exist, but if they do, then #334 is probably the preferred solution.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -223,7 +223,10 @@ func main() {
 				return fmt.Errorf("gopherjs run: no go files listed")
 			}
 
-			tempfile, err := ioutil.TempFile("", filepath.Base(args[0])+".")
+			tempfile, err := ioutil.TempFile(currentDirectory, filepath.Base(args[0])+".")
+			if err != nil && strings.HasPrefix(currentDirectory, runtime.GOROOT()) {
+				tempfile, err = ioutil.TempFile("", filepath.Base(args[0])+".")
+			}
 			if err != nil {
 				return err
 			}
@@ -363,7 +366,7 @@ func main() {
 					return err
 				}
 
-				tempfile, err := ioutil.TempFile("", "test.")
+				tempfile, err := ioutil.TempFile(currentDirectory, "test.")
 				if err != nil {
 					return err
 				}


flimzy:
Bump?
I'm sure this PR got lost in the recent semi-flurry of other bug reports, not to mention the Real Life™ that we all hopefully have.

But are there any outstanding concerns regarding this PR (or #334) that I should try to tackle, @neelance?

neelance:
I like this much better than #334.
@shurcooL What do you think?

shurcooL:
I'm in agreement. LGTM.

However, one comment about the implementation:

```
strings.HasPrefix(currentDirectory, runtime.GOROOT())
```

I believe that is brittle, specifically that it may result in false negatives. However, it shouldn't result in false positives, which is why the PR LGTM and can be merged as is.

It may result in false negatives in situations where symlinks are involved in the relevant paths, and also in situations where case-insensitive filesystems are used (e.g., imagine `currentDirectory` has value "/usr/local/GO" but `runtime.GOROOT()` returns `/USR/LOCAL/go`).

Those situations are possible but should be rare, so if you're okay with it, we can merge as is, and if the false negatives result in visible problems in situations where it matters, then you can work on improving it. Otherwise it'll be fine as is.

flimzy:
Thanks for the feedback, @shurcooL.

Since, in the specific case of upstream Go tests, which this code is meant to address, do a [Chdir to runtime.GOROOT()](https://github.com/gopherjs/gopherjs/blob/master/tests/run.go#L138), then read the input file names from disk, one would hope that this would address both the symlink and case normalization issues.

Still, running one of these tests manually, with a symlink or mismatched case, could lead to a failure. But if someone is running these tests manually in such an environment, hopefully they'll have enough clues to work around the shortcoming... and/or fix the problem at that time, as you suggest.

shurcooL:
Yep, that sounds good to me.

neelance:
@flimzy As always, thanks for contributing!

PR #362, opened 2015-12-19
Filter gopherjs warnings from test output.

flimzy:
For some reason, I'm unable to get nodejs to find the source-map-support module when running the test suite, which causes every one of the official go tests to fail with "invalid output":

    FAIL    fixedbugs/issue9006.go  0.326s
    # go run run.go -- fixedbugs/issue9110.go
    incorrect output

As running tests with source maps doesn't seem especially valuable to me, this patch attempts to remedy the problem by filtering out this text from the output of the tests

### DIFF
--- a/tests/run.go
+++ b/tests/run.go
@@ -734,8 +734,8 @@ func (t *test) run() {
 					t.err = err
 					return
 				}
-				if strings.Replace(string(out), "\r\n", "\n", -1) != t.expectedOutput() {
-					t.err = fmt.Errorf("incorrect output\n%s", out)
+				if fo := filterOutput(out); fo != t.expectedOutput() {
+					t.err = fmt.Errorf("incorrect output\n%s", fo)
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


shurcooL:
That implementation should not be merged as is. This can and should be done by arranging so the warnings are not printed in test output, rather than using a regex to filter out from the result output.

flimzy:
That's a good idea.

flimzy:
I'm not seeing an obvious way to accomplish this.

The official go tests aren't actually run as tests; they're invoked with 'gopherjs run'. So suppressing the output in the case of 'gopherjs test' isn't enough.  Perhaps a command-line flag should be added, which could be used by tests/run.go when invoking `gopherjs run`, and could also be used by others who simply don't want the annoyance?

flimzy:
#363 adds and uses a --quiet flag to suppress output for the tests in question. If you can think of a more clever solution, let me know, and I'll be happy to work on that.

neelance:
Oops, the GitHub UI tricked me into merging this (wasn't me xD). I've removed the commit immediately, since I already merged #363 which is a much nicer solution. This one can stay closed, right?

flimzy:
Yes, this can remain closed. Thanks for merging #363.

PR #363, opened 2015-12-20
Add --quiet commandline option to suppress non-fatal warnings in output

flimzy:
And use the -q flag when running official Go tests, to avoid test failures for "invalid output" when gopherjs emits warnings about not loading the source-map-support module:

    FAIL    fixedbugs/issue9006.go  0.326s
    # go run run.go -- fixedbugs/issue9110.go
    incorrect output


neelance:
LGTM. Thanks! Merging...

PR #366, opened 2015-12-22
compiler: fix runtime exceptions with blocking deferred functions

bpowers:
This issue arises when re-winding the stack to complete a blocking function call inside a deferred function.  More specifically it occurs in functions that call blocking functions both with the defer keyword and in return statements.  A trivial example is given in [tests/deferblock_test.go](tests/deferblock_test.go), `ioutil.ReadFile` is an example of such a function from the stdlib.

All the test cases pass for me, and I've added a new file containing a test for this.  Let me know if there is any better way to test this.

### DIFF
--- a/compiler/statements.go
+++ b/compiler/statements.go
@@ -282,7 +282,13 @@ func (c *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 			}
 			results = c.resultNames
 		}
-		c.Printf("return%s;", c.translateResults(results))
+		rVal := c.translateResults(results)
+		if c.Flattened[s] {
+			resumeCase := c.caseCounter
+			c.caseCounter++
+			c.Printf("/* */ $s = %[1]d; case %[1]d:", resumeCase)
+		}
+		c.Printf("return%s;", rVal)
 
 	case *ast.DeferStmt:
 		isBuiltin := false
--- a/tests/deferblock_test.go
+++ b/tests/deferblock_test.go
@@ -1,0 +1,42 @@
+package tests
+
+import (
+	"testing"
+	"time"
+)
+
+func inner(ch chan struct{}, b bool) ([]byte, error) {
+	// ensure gopherjs thinks that this inner function can block
+	if b {
+		<-ch
+	}
+	return []byte{}, nil
+}
+
+// this function's call to inner never blocks, but the deferred
+// statement does.
+func outer(ch chan struct{}, b bool) ([]byte, error) {
+	defer func() {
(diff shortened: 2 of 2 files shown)


compiler/statements.go
```diff
@@ -794,8 +796,14 @@ func (c *funcContext) translateResults(results []ast.Expr) string {
 			values[i] = c.translateImplicitConversion(result, tuple.At(i).Type()).String()
 		}
 		c.delayedOutput = nil
-		return " [" + strings.Join(values, ", ") + "]"
+		rVal = " [" + strings.Join(values, ", ") + "]"
+	}
+	if flatten {
+		resumeCase := c.caseCounter
+		c.caseCounter++
+		c.Printf("/* potential blocking in defer */ $s = %[1]d; case %[1]d:", resumeCase)
```
neelance: Please keep the pattern that the comment is either empty or has the corresponding non-blocking/non-flattened code.

compiler/statements.go
```diff
@@ -282,7 +282,7 @@ func (c *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 			}
 			results = c.resultNames
 		}
-		c.Printf("return%s;", c.translateResults(results))
+		c.Printf("return%s;", c.translateResults(results, c.Flattened[s]))
```
neelance: Seems like the new logic is only used here. Can we move it above this line, e.g. `rVal := c.translateResults(results); if c.Flattened[s] { [...] }`? I'd rather avoid having `c.Printf` in the helper function `translateResults`.

neelance:
@bpowers I am really happy that you not only reported the bug, but actually went into the code to fix it. Awesome! Please don't get discouraged by my comments, this pull request is close to be merged. :-)

compiler/statements.go
bpowers: ah, yea!  This was for debugging, I will take it out.

bpowers:
Hi @neelance!  Thanks for the encouragement.  I had played around with a couple of different approaches (which didn't end up working out) which resulted in things being a little convoluted -- your suggestions are spot on.  I've updated the PR, PTAL.

neelance:
LGTM. Please squash the two commits into one, then I'll merge.

neelance:
LGTM. Please squash the two commits into one, then I'll merge.

bpowers:
great! squash pushed

neelance:
Merged, thanks!

bpowers:
Thank you!

Issue #367, opened 2015-12-23
How do I debug native packages?


flimzy:
I'm trying to tinker with the GopherJS implementation of runtime.go, to flesh out the `Callers()` method, which is needed by a third party package I want to use.

I'm having a heck of a time debugging, though.  I haven't found the "secret sauce" to make my changes to GopherJS's `./compiler/natives/runtime/runtime.go` file effectual for my third-party packages.

There must be a simple step I'm overlooking.

I've tried (within the gopherjs checkout):

    go clean; go build; go install

This successfully updates the gopherjs binary, but doesn't seem to (always?) update the native packages.

In my third party package, I've attempted

    go clean; gopherjs test

But that also has no effect, as demonstrated by utterly breaking `runtime.go` by introducing obvious typos... But the tests still execute, with no compilation failure.

Clues?

neelance:
Yeah, I know what you're talking about. The reason is that GopherJS only compiles if the source files or the GopherJS binary itself have a newer timestamp than the output file. This check currently does not include the `natives` files. The easiest solution right now is to `touch` the GopherJS binary to force it to recompile everything. In the long term we should probably use `go generate` to include the native files into the binary itself. @shurcooL Can your tool help here?

flimzy:
Thanks for the `touch` hint, @neelance.

shurcooL:
> In the long term we should probably use `go generate` to include the native files into the binary itself. @shurcooL Can your tool help here?

Absolutely. You're referring to [vfsgen](https://github.com/shurcooL/vfsgen), right?

shurcooL:
@neelance, if you can tell me what files need to be embedded, and where they are used, I can look at making a PR. It's quite easy.

neelance:
@shurcooL The files here https://github.com/gopherjs/gopherjs/tree/master/compiler/natives are used here https://github.com/gopherjs/gopherjs/blob/6ed759488f1dc291c2f04de268cc609342e4c70d/build/build.go#L149. It's not that trivial, because we need to add the VFS abstraction at several places.

shurcooL:
I can get it started by making a PR that simply creates that VFS. Next step will be to use it in all appropriate places.

shurcooL:
An important question is where to place the virtual filesystem variable that will hold the native package data.

[...]

PR #369, opened 2015-12-30
Create a VFS variable for native packages.

shurcooL:
This commit creates a package with import path "github.com/gopherjs/gopherjs/compiler/natives" that contains one exported variable `Data`. `Data` is a virtual filesystem that holds the native packages.

When using `dev` build tag, that variable reads directly from disk. Otherwise, it uses the statically generated version.

The data can be updated with a standard `go generate`.

This is for #367. It's not used anywhere yet. This PR is for review. Some notes:

- Developers of GopherJS will need to get additional Go packages like "github.com/shurcooL/vfsgen" and "github.com/shurcooL/httpfs", as well as the "github.com/shurcooL/vfsgen/cmd/vfsgendev" binary, in order to run `go generate` and build using `dev` build tag. Users of GopherJS will not need anything extra.

- Since the package lies at the root of the directory being embedded, a filter that skips all files (not directories) in root folder needs to be used.

- ~~A proposed enhancement to `vfsgen` is to create a binary, which would replace the need for `data_generate.go` file. See https://github.com/sourcegraph/appdash/pull/91 for proposed change details.~~ **Edit:** This is already done, see mention of `vfsgendev` above.

/cc @neelance @slimsag


shurcooL:
Is this PR still needed?

neelance:
At some point I'd like to bake the files into the binary. I'm okay with keeping this PR until then.

shurcooL:
This has been resolved via ab917e062e6a7940608ee3c9d422382784f5a7ab. Closing.

PR #370, opened 2015-12-31
Replace data_generate.go with vfsgendev.

shurcooL:
Requires `vfsgendev` binary to run `go generate`.

Depends on https://github.com/shurcooL/vfsgen/pull/16.

/cc @slimsag


slimsag:
Not much for me to say other than the fact that I think this is a good change (assuming vfsgen is a good choice for GopherJS's needs, that is). :+1:

shurcooL:
To summarize:

### Advantages

- Less code in the repository.

### Disadvantages

- Requires `vfsgendev` binary in PATH, in addition to `vfsgen` library (previous version only required `vfsgen` library).

PR #376, opened 2016-01-13
Add support for alternate JS interpreters

flimzy:
My motivation for this is to run GopherJS tests using [PhantomJS](http://phantomjs.org/).  I can see it also being useful on systems with multiple versions of NodeJS.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -69,6 +69,7 @@ func main() {
 	flagColor := pflag.Lookup("color")
 	tags := pflag.String("tags", "", "a list of build tags to consider satisfied during the build")
 	flagTags := pflag.Lookup("tags")
+	jsCmd := pflag.String("exec", "", "specify alternate js interpreter (default: node)")
 
 	cmdBuild := &cobra.Command{
 		Use:   "build [packages]",
@@ -249,7 +250,7 @@ func main() {
 			if err := s.BuildFiles(args[:lastSourceArg], tempfile.Name(), currentDirectory); err != nil {
 				return err
 			}
-			if err := runNode(tempfile.Name(), args[lastSourceArg:], "", options.Quiet); err != nil {
+			if err := runJS(tempfile.Name(), *jsCmd, args[lastSourceArg:], "", options.Quiet); err != nil {
 				return err
 			}
 			return nil
@@ -405,7 +406,7 @@ func main() {
 				}
 				status := "ok  "
 				start := time.Now()
-				if err := runNode(tempfile.Name(), args, pkg.Dir, options.Quiet); err != nil {
+				if err := runJS(tempfile.Name(), *jsCmd, args, pkg.Dir, options.Quiet); err != nil {
 					if _, ok := err.(*exec.ExitError); !ok {
 						return err
 					}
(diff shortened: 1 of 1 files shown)


neelance:
Sorry, but I think this use case is too uncommon for built-in support. For `gopherjs run`, you can just use `gopherjs build` instead and run the generated file. For `gopherjs test` we could add an option to also just generate a JS file instead of immediately running the code. Would that be okay for you?

shurcooL:
I was going to make a comment here.

If we're going to merge this, we should use an existing flag name for this that the `go` tool already has, rather than inventing our own that's slightly different.

Specifically, the `go` tool has an optional flag `-exec xprog` for commands that execute the result of a build (there's only one such command, it's `go run`):

```
~ $ go run -help
usage: run [build flags] [-exec xprog] gofiles... [arguments...]

Run compiles and runs the main package comprising the named Go source files.
A Go source file is defined to be a file ending in a literal ".go" suffix.

By default, 'go run' runs the compiled binary directly: 'a.out arguments...'.
If the -exec flag is given, 'go run' invokes the binary using xprog:
	'xprog a.out arguments...'.
If the -exec flag is not given, GOOS or GOARCH is different from the system
default, and a program named go_$GOOS_$GOARCH_exec can be found
on the current search path, 'go run' invokes the binary using that program,
for example 'go_nacl_386_exec a.out arguments...'. This allows execution of
cross-compiled programs when a simulator or other execution method is
available.

For more about build flags, see 'go help build'.

See also: go build.
```

So if we were to merge this, I think we should also use `-exec` as the flag name rather than `-jsinterpretor`.

flimzy:
@shurcooL: Indeed, --exec would be a much better name.

Generating the js output for tests would be a step in the right direction.

Without built-in support, it will be harder to run the entire rest suite with an alternate interpreter, which this patch alone doesn't provide either, but is a goal of mine.  Running the entire test suite under both PhantomJS and SpiderMonkey seem within reach this way.

There might be better ways to accomplish that, though.

flimzy:
As Go supports -exec, for the sake of executing cross-compiled programs, it makes me wonder if GopherJS supporting -exec for the sake of alternate JS runtime environments is that much of a stretch. It seems like the same problem in spirit.

Of course outputting compiled tests allows a lot of flexibility, but the same could be said of Go as well.

Anyway, just some food for thought.

shurcooL:
> As Go supports -exec, for the sake of executing cross-compiled programs, it makes me wonder if GopherJS supporting -exec for the sake of alternate JS runtime environments is that much of a stretch. It seems like the same problem in spirit.

Agreed. That's why I posted that.

I'm actually planning on trying to add more flexibility to the `go` tool so that one could specify `gopherjs` as the -compiler value for it. If that works out, I have no problem with making its -exec flag support both node and phantomjs. Or any other binary.

neelance:
Shortcuts are okay for the very common use cases, e.g. `gopherjs run` with node. Using a few lines of bash script for the not so common use cases seems better to me than adding more flags and making the code more complicated. As far as I can see right now, the `go` tool would also work very well without the `-exec` flag. #379 will enable the desired use case, so I'd like to close this one if there is no other use case which is not easily possible otherwise. @flimzy Is that okay for you?

flimzy:
It's okay with me.

PR #377, opened 2016-01-13
Fix Error prototype for PhantomJS environments

flimzy:
The JavaScript interpreter built into PhantomJS is apparently rather ancient, and has some oddities.  Of note for this PR, `throw new Error()` creates an Error object with a stack trace, but `new Error()`simply creates a string.  This is important in a few places where GopherJS needs to generate a JS stack trace, such as [here](https://github.com/gopherjs/gopherjs/blob/master/compiler/natives/runtime/runtime.go#L48), which caused a runtime crash prior to this PR.  I will follow on this work with proper parsing error stack traces from PhantomJS.

This PR works around this short-coming by doing a `throw` in the Error object constructor.

There may be a better way to do this (my JS-foo is weak).

There is also likely a better place, organizationally speaking, at least, to do this. I'm open for suggestions, if the concept is deemed appropriate.

### DIFF
--- a/compiler/prelude/prelude.go
+++ b/compiler/prelude/prelude.go
@@ -16,6 +16,20 @@ if (typeof window !== "undefined") { /* web page */
   $global = this;
 }
 
+// Fix Error prototype for PhantomJS
+if (typeof phantom !== "undefined") {
+	var OriginalErrorPrototype = Error.prototype;
+	var OriginalError = Error;
+	$global.Error = function(m) {
+		try {
+			throw new OriginalError(m);
+		} catch (e) {
+			return e;
+		}
+	};
+	$global.Error.prototype = OriginalErrorPrototype;
+}
+
 if ($global === undefined || $global.Array === undefined) {
   throw new Error("no global object found");
 }


neelance:
I think it is not GopherJS's responsibility to fix the odd behavior of PhantomJS. Maybe you can just prepend the workaround to the output?

shurcooL:
Doesn't GopherJS already do a little bit to support various execution modes?

https://github.com/gopherjs/gopherjs/blob/a240d35ae72a9f92e4115ce0bd8efb288caf26b1/compiler/prelude/prelude.go#L7-L17

flimzy:
I can see the argument against supporting PhantomJS (or any other odd JS implementation). Especially as there are likely to be other necessary workarounds.  (As an example, I suspect that GopherJS's stack trace parsing is specific to V8, and may be broken in other browsers).

Adding compatibility layers for all browsers under the sun is one of the main reasons for JS bloat in popular libraries like jQuery.

And speaking just for my personal use case, if the PhantomJS hacks ended up being much larger than what I have already written, I wouldn't want that in my production output.  Likewise, if someone needs compatibility shims for IE10, I don't want those in my output code which I know will only run in Chrome, etc.

So in that regard, I completely agree with @neelance about separating this stuff from GopherJS.  On the other hand, it sure would be nice if there was an easy library of such shims, and ideally an easy way to include them in GopherJS.

I wonder if it would be tenable to ship such browser shims with GopherJS in a separate directory, or possibly as a separate repo, and possibly include a command-line argument to optionally include the shims in output, when requested.

This might not be the most straight forward way to address all compatibility issues (again, I'm thinking of stack trace parsing, which may be more cleanly addressed with interpreter-specific regexes).

PR #379, opened 2016-01-16
Add the ability to save output of test compile

flimzy:
This is a followup to @neelance's [comment](https://github.com/gopherjs/gopherjs/pull/376#issuecomment-171467723) on issue #376.  It adds support for the -o/--output option (borrowed from 'build' mode) to `test` mode.

The current implementation *only* outputs the file, rather than also running the tests. This may or may not be best. If the consensus is to run the tests, too, that change will be simple to make.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -264,6 +264,8 @@ func main() {
 	run := cmdTest.Flags().String("run", "", "Run only those tests and examples matching the regular expression.")
 	short := cmdTest.Flags().Bool("short", false, "Tell long-running tests to shorten their run time.")
 	verbose := cmdTest.Flags().BoolP("verbose", "v", false, "Log all tests as they are run. Also print all text from Log and Logf calls even if the test succeeds.")
+	compileOnly := cmdTest.Flags().BoolP("compileonly", "c", false, "Compile the test binary to pkg.test.js but do not run it (where pkg is the last element of the package's import path). The file name can be changed with the -o flag.")
+	outputFilename := cmdTest.Flags().StringP("output", "o", "", "Compile the test binary to the named file. The test still runs (unless -c is specified).")
 	cmdTest.Flags().AddFlag(flagMinify)
 	cmdTest.Flags().AddFlag(flagColor)
 	cmdTest.Run = func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


tool.go
```diff
@@ -264,6 +264,7 @@ func main() {
 	run := cmdTest.Flags().String("run", "", "Run only those tests and examples matching the regular expression.")
 	short := cmdTest.Flags().Bool("short", false, "Tell long-running tests to shorten their run time.")
 	verbose := cmdTest.Flags().BoolP("verbose", "v", false, "Log all tests as they are run. Also print all text from Log and Logf calls even if the test succeeds.")
+	output := cmdTest.Flags().StringP("output", "o", "", "write to output file instead of executing the test directly")
```
dominikh: I think we should use `-c` to stay consistent with the official Go tool

dominikh: Hm. I just realised that in the Go tool, it's a combination of `-c` and `-o`. `-c` to "Compile the test binary to pkg.test but do not run it" and `-o` to overwrite the default of where the file should be written to. I still think we should imitate the Go tool, though.

flimzy: Thanks for the suggestions.  I'll make those changes.

shurcooL: > I still think we should imitate the Go tool, though.

Agreed. It's the comment I was going to make too.

flimzy:
This now implements both -c/--compileonly and -o/--output.

tool.go
```diff
@@ -404,15 +428,17 @@ func main() {
 					args = append(args, "-test.v")
 				}
 				status := "ok  "
-				start := time.Now()
-				if err := runNode(tempfile.Name(), args, pkg.Dir, options.Quiet); err != nil {
-					if _, ok := err.(*exec.ExitError); !ok {
-						return err
+				if !*compileOnly {
```
neelance: Please transform this into an early exit and move it above the `var args []string` line.

tool.go
```diff
@@ -376,17 +378,39 @@ func main() {
...
-				tempfile, err := ioutil.TempFile(currentDirectory, "test.")
-				if err != nil {
-					return err
+				if *compileOnly && len(*outputFilename) == 0 {
+					*outputFilename = pkg.Package.Name + "_test.js"
+				}
+
+				var outfile *os.File
+				if len(*outputFilename) > 0 {
+					err := os.Remove(*outputFilename)
```
neelance: Are those `Remove` calls necessary? Shouldn't `Create` just overwrite the files?

neelance:
The functionality itself looks good. Please take a look at my comments on the code.

neelance:
The code LGTM now. Please stash the commits, then we can merge. :+1:

tool.go
```diff
@@ -376,20 +378,38 @@ func main() {
 					return err
 				}
 
-				tempfile, err := ioutil.TempFile(currentDirectory, "test.")
-				if err != nil {
-					return err
+				if *compileOnly && len(*outputFilename) == 0 {
```
shurcooL: I think it's nicer to do empty string checks with the `*outputFilename == ""` form. That makes it immediately obvious it's a string, rather than a slice of strings or something higher level.

flimzy: Done.

neelance:
@flimzy Thanks for your PR!

Issue #239, opened 2015-06-07
Feature Request: `gopherjs test` support running and verifying example code.


shurcooL:
One feature of `go test` command that is handy for short/quick/easy to write tests is that it runs and verifies example code (if it has a concluding comment with "// Output:").

See https://godoc.org/testing#hdr-Examples.

It'd be a nice enhancement if `gopherjs test` supported that too.

shurcooL:
It would also be good if GopherJS supported the `TestMain` functionality added in Go 1.4 to allow custom setup and teardown per package (rather than per test). It's described [here](https://godoc.org/testing#hdr-Main). /cc @slimsag

shurcooL:
This _would_ be resolved by WIP implementation of #388 or by @flimzy's PR, however, both efforts are blocked on issue https://github.com/gopherjs/gopherjs/issues/381 being resolved.

flimzy:
Now that #381 has been fixed, is it worth me working on supporting examples, or wait for #485 to be completed?

Thoughts, @shurcooL ?

shurcooL:
It's up to you. I don't expect #485 to be done soon, not within the next month. So feel free to go for it, or wait if you prefer.

PR #380, opened 2016-01-19
Add support for TestMain function in test mode

flimzy:
This adds support for the `TestMain` function in test mode, described [here](https://godoc.org/testing#hdr-Main), and requested in issue #239.  Naturally, I took inspiration from the relevant official [go source](https://github.com/golang/go/blob/master/src/cmd/go/test.go).

I have tested it to the extent I know to. It's possible I may have missed some functionality. Any other suggestions or feedback is welcome.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -334,10 +334,6 @@ func main() {
 
 				tests := &testFuncs{Package: pkg.Package}
 				collectTests := func(testPkg *gbuild.PackageData, testPkgName string, needVar *bool) error {
-					_, err := s.BuildPackage(testPkg)
-					if err != nil {
-						return err
-					}
 					if testPkgName == "_test" {
 						for _, file := range pkg.TestGoFiles {
 							if err := tests.load(filepath.Join(pkg.Package.Dir, file), testPkgName, &tests.ImportTest, &tests.NeedTest); err != nil {
@@ -350,6 +346,10 @@ func main() {
 								return err
 							}
 						}
+					}
+					_, err := s.BuildPackage(testPkg)
+					if err != nil {
+						return err
 					}
 					return nil
 				}
--- a/tool.go
+++ b/tool.go
@@ -335,11 +335,16 @@ func main() {
 					}
 
 					for _, decl := range archive.Declarations {
-						if strings.HasPrefix(decl.FullName, testPkg.ImportPath+".Test") {
+						switch {
+						case decl.FullName == testPkg.ImportPath+".TestMain":
+							if tests.TestMain != nil {
+								return fmt.Errorf("multiple definitions of TestMain")
+							}
+							tests.TestMain = &testFunc{Package: testPkgName, Name: decl.FullName[len(testPkg.ImportPath)+1:]}
+						case strings.HasPrefix(decl.FullName, testPkg.ImportPath+".Test"):
 							tests.Tests = append(tests.Tests, testFunc{Package: testPkgName, Name: decl.FullName[len(testPkg.ImportPath)+1:]})
 							*needVar = true
-						}
-						if strings.HasPrefix(decl.FullName, testPkg.ImportPath+".Benchmark") {
+						case strings.HasPrefix(decl.FullName, testPkg.ImportPath+".Benchmark"):
 							tests.Benchmarks = append(tests.Benchmarks, testFunc{Package: testPkgName, Name: decl.FullName[len(testPkg.ImportPath)+1:]})
 							*needVar = true
 						}
@@ -773,6 +778,7 @@ type testFuncs struct {
 	Tests      []testFunc
 	Benchmarks []testFunc
 	Examples   []testFunc
+	TestMain   *testFunc
 	Package    *build.Package
 	NeedTest   bool
 	NeedXtest  bool
@@ -788,6 +794,9 @@ var testmainTmpl = template.Must(template.New("main").Parse(`
 package main
 
 import (
+{{if not .TestMain}}
+	"os"
+{{end}}
 	"regexp"
 	"testing"
 
@@ -832,7 +841,12 @@ func matchString(pat, str string) (result bool, err error) {
 }
 
 func main() {
-	testing.Main(matchString, tests, benchmarks, examples)
+	m := testing.MainStart(matchString, tests, benchmarks, examples)
+{{with .TestMain}}
+	{{.Package}}.{{.Name}}(m)
+{{else}}
+	os.Exit(m.Run())
+{{end}}
 }
 
 `))
--- a/tool.go
+++ b/tool.go
@@ -334,7 +334,7 @@ func main() {
 
 				tests := &testFuncs{Package: pkg.Package}
 				collectTests := func(testPkg *gbuild.PackageData, testPkgName string, needVar *bool) error {
-					archive, err := s.BuildPackage(testPkg)
+					_, err := s.BuildPackage(testPkg)
 					if err != nil {
 						return err
 					}
--- a/tool.go
+++ b/tool.go
@@ -2,9 +2,11 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"go/ast"
 	"go/build"
+	"go/doc"
 	"go/parser"
 	"go/scanner"
 	"go/token"
@@ -18,11 +20,14 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	// 	"sort"
 	"strconv"
 	"strings"
 	"syscall"
 	"text/template"
 	"time"
+	"unicode"
+	"unicode/utf8"
 
 	gbuild "github.com/gopherjs/gopherjs/build"
 	"github.com/gopherjs/gopherjs/compiler"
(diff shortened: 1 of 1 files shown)


neelance:
LGTM. Thanks for the contribution!
@shurcooL PTAL

shurcooL:
For reference, this is the commit that added this feature to Go.

https://github.com/golang/go/commit/182d1316dd975f426451cee34ba2e3e0953e084f

So far, the changes in this PR look good, but I see a few things missing from that commit.

tool.go
```diff
@@ -319,11 +319,13 @@ func main() {
 					}
 
 					for _, decl := range testPkg.Archive.Declarations {
-						if strings.HasPrefix(decl.FullName, testPkg.ImportPath+".Test") {
+						switch {
+						case decl.FullName == testPkg.ImportPath+".TestMain":
```
shurcooL: This only checks that the func name matches "TestMain", but doesn't check that the func signature takes `m *testing.M` parameters and has no return variables, does it?

That potentially means that some Go package that has a `func TestMain(t *testing.T)` will get interpreted differently by `go` and `gopherjs`, if I understand this correctly.

tool.go
```diff
@@ -319,11 +319,13 @@ func main() {
...
+							tests.TestMain = &testFunc{Package: testPkgName, Name: decl.FullName[len(testPkg.ImportPath)+1:]}
```
shurcooL: The equivalent `go` commit has a check here:

```Go
if t.TestMain != nil {
	return errors.New("multiple definitions of TestMain")
}
```

I don't understand what situation it's for. How can you declare two `TestMain` funcs without a "redeclared in this block, previous declaration at" error?

Oh, okay, from trying it, it appears this may just be a special case to override the error. Compare what happens when there's two `Test1` funcs:

```
31 $ go test -v
# github.com/shurcooL/play/31
./main_test.go:24: Test1 redeclared in this block
	previous declaration at ./main_test.go:21
FAIL	github.com/shurcooL/play/31 [build failed]
```

On the other hand, the error message is different when there are two `TestMain`s:

```
31 $ go test -v
# github.com/shurcooL/play/31
multiple definitions of TestMain
FAIL	github.com/shurcooL/play/31 [setup failed]
```

I'd say that's a good reason to add that if guard to be consistent with Go tool.

tool.go
shurcooL: Confirmed, if the signature is `func TestMain(t *testing.T)` then the `go` tool treats it like a normal test:

```
31 $ go test -v
=== RUN   TestMain
--- PASS: TestMain (0.00s)
=== RUN   Test1
--- PASS: Test1 (0.00s)
PASS
ok  	github.com/shurcooL/play/31	0.022s
```

shurcooL:
I think if you add those 2 pieces of missing functionality I mentioned in inline comments above, this should be good to merge.

flimzy:
Thanks, @shurcooL, for your critique.  My original work included the extra error message, but somehow I missed that in my final cut. I've added it back in.

The other issue will be more involved, as the GopherJS implementation differs significantly from the Go implementation at this point.  Let me see what I can come up with.

shurcooL:
Thanks.

I've looked a little into it, and I saw that the original Go code works with the Go AST, while here in GopherJS you're ranging over `testPkg.Archive.Declarations`, which has type [`Decl`](https://sourcegraph.com/github.com/gopherjs/gopherjs@master/.GoPackage/github.com/gopherjs/gopherjs/compiler/.def/Decl).

Unfortunately, it's not documented, so it's hard to tell, but I hope/suspect you can find the func signature in the [`Vars`](https://sourcegraph.com/github.com/gopherjs/gopherjs@master/.tree/compiler/compiler.go?seldef=%2Fgithub.com%2Fgopherjs%2Fgopherjs%40master%2F.GoPackage%2Fgithub.com%2Fgopherjs%2Fgopherjs%2Fcompiler%2F.def%2FDecl%2FVars) field.

Finally, if you don't mind, please do the additional changes in separate commits, and squash at the end when we're ready to merge. That makes the PR easier to review as it changes.

flimzy:
My latest commit probably needs some work.  It is in some ways a blind copy-and-paste of the relevant bits of logic from Go into GopherJS, adapting them where necessary.

This means that we now run `parser.ParseFile()` on test files a second time, but it appears this is the way Go does it, too--and perhaps there's a reason for this (beyond simply being easier to write) that would be revealed with further investigation.

Also, Go builds 'test' as a separate executable, which lends itself to a clean separation from the other `go` commands. GopherJS is more monolythic, meaning that adding this functionality into `tool.go` might be considered more clutter. A simple step toward clarity might be just to create a `test.go`, but within the same `main` package. I think @neelance has hinted at some long-term plans to potentially refactor parts of `tool.go`, so I don't know how this might fit into that larger picture.

This also gets us much closer to supporting [Examples](https://godoc.org/testing#hdr-Examples), which depends on an `*ast.File` argument to [doc.Examples](https://golang.org/pkg/go/doc/#Examples) to extract the expected output for each example function.  This commit doesn't yet make Examples fully functional, though (so I have some of that commented out), but I don't think it will be terribly hard to get it working (which I'll work on next).

But this seemed like a good place to solicit feedback.

neelance:
What's the status of this PR?

flimzy:
I have not been following closely all the work happening with GopherJS lately, in particular the long-term plans to be more compatible with upstream tools, so I don't know if this PR is, or will be, obsoleted by some of that work.

If the approach I took in this PR is appropriate, I would be happy to try to resolve the conflicts and get it back into a mergable state.

shurcooL:
I think I went a little trigger-happy with that label. It applies to issues only, and this is a PR. I'll remove it.

It's true this issue will become resolved once we move to using a modified version of cmd/go, but the timeline for that is unclear and it may not happen very soon. There's still value in resolving this. But it's good to be aware of which issues are relevant to the build tools and not the compiler.

bep:
This PR would be really useful.

flimzy:
My personal hacking time  has been limited lately, but I'll try to get to this again soon, and bring the PR up to date.

I suspect it will mostly involve re-applying the relevant bits of code from Go 1.6.

If someone else wants to get to it before I do, please feel free.

flimzy:
I have updated the PR to merge cleanly.  I have further verified that @shurcooL's request for the TestMain-specific error messages is addressed.

As mentioned in a previous comment, the other issue, of strictly checking the function signature, is not as easily solved. I'm continuing to investigate.

flimzy:
This PR is now up to date, and as far as I know, ready to merge. Both of @shurcooL's issues have been addressed:  Proper function signature checking on TestMain(), and the TestMain-related error message 'multiple definitions of TestMain' takes priority over the standard 'TestMain redeclared in this block' message.

I welcome any further feedback or suggestions.

tool.go
```diff
@@ -18,11 +20,14 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	// 	"sort"
```
shurcooL: Remove this.

(I guess it's commented out because of the example code being commented out, right? In any case, it's unclear when examples will work, so this may stay here for a while, and it's not helpful. Better to remove it now and add it when it's needed.)

tool.go
```diff
@@ -784,18 +794,111 @@ type testFunc struct {
...
+	// 	sort.Sort(byOrder(ex))
+	// 	for _, e := range ex {
+	// 		*doImport = true // import test file whether executed or not
+	// 		if e.Output == "" && !e.EmptyOutput {
+	// 			// Don't run examples with no output.
+	// 			continue
+	// 		}
+	// 		t.Examples = append(t.Examples, testFunc{pkg, "Example" + e.Name, e.Output})
+	// 		*seen = true
+	// 	}
```
shurcooL: Is this the code to support examples that's commented out because it's blocked on #381, right?

I have a simple request. Either remove the code and just make note of it in the PR (in a comment or so), so we can look it up and add it later after #381 is resolved.

Either that, or, if you want to keep it, then please add a comment explaining why this is commented out. E.g.:

```Go
// This is the code to support examples. It's blocked on https://github.com/gopherjs/gopherjs/issues/381 being resolved.
/*ex := doc.Examples(f)
sort.Sort(byOrder(ex))
for _, e := range ex {
	*doImport = true // import test file whether executed or not
	if e.Output == "" && !e.EmptyOutput {
		// Don't run examples with no output.
		continue
	}
	t.Examples = append(t.Examples, testFunc{pkg, "Example" + e.Name, e.Output})
	*seen = true
}*/
```

If you're indifferent, I prefer removing it (just make sure a copy of it is accessible somewhere in this PR for lookup at a later time).

shurcooL:
Thank you for working on this and your patience @flimzy.

I've left 2 minor comments, but if this works, I see no reason not merge this. It has LGTM from @neelance, and even though there have been changes since then, as long as it works, we can merge.

Sorry for letting this drag on for longer than it should've; I've been focusing on other work and did not have enough time to spend on this.

shurcooL:
Tested locally, `gopherjs test` still works and `TestMain` is working. Merging. I'll apply the 2 comments I left above myself.

Thanks a lot again! :)

flimzy:
Thanks for merging!

PR #385, opened 2016-01-29
Override os.Hostname()

abrander:
`os.Hostname()` doesn't work in compiled code because it will trigger a syscall.

This replacement tries to do the right thing. In a node.js environment it will call `require("os").hostname()`. In the browser it will use `window.location.hostname`, which is the closest thing we got.

### DIFF
--- a/compiler/natives/os/os.go
+++ b/compiler/natives/os/os.go
@@ -24,3 +24,27 @@ func init() {
 }
 
 func runtime_beforeExit() {}
+
+// Override the builtin os.Hostname(). The builtin Hostname() will result in
+// a non-working syscall on most (all?) platforms
+func Hostname() (string, error) {
+	// Browser
+	// The browser doesn't know anything about any hostname, so we use the host part of
+	// the current url.
+	if location := js.Global.Get("location"); location != js.Undefined {
+		if browserHostname := location.Get("hostname"); browserHostname != js.Undefined {
+			return browserHostname.String(), nil
+		}
+	}
+
+	// Node.js
+	// We call node's require("os").hostname() which should return current hostname
+	if require := js.Global.Get("require"); require != js.Undefined {
+		// "os" is in nodejs core, it should always exist
+		if nodeHostname := require.Invoke("os").Call("hostname"); nodeHostname != js.Undefined {
+			return nodeHostname.String(), nil
+		}
+	}
+
+	return "", ErrNotExist
+}


compiler/natives/os/os.go
```diff
@@ -24,3 +24,27 @@ func init() {
...
+		if browserHostname := location.Get("hostname"); browserHostname != js.Undefined {
+			return browserHostname.String(), nil
+		}
+	}
+
+	// Node.js
+	// We call node's require("os").hostname() which should return current hostname
+	if require := js.Global.Get("require"); require != js.Undefined {
+		// "os" is in nodejs core, it should always exist
+		if nodeHostname := require.Invoke("os").Call("hostname"); nodeHostname != js.Undefined {
```
flimzy: What happens here if the "os" module isn't available or fails to load, as may be common in a browser when using browserify?

abrander: For that to happen a few planets have to be aligned:

- The browser doesn't implement the global `location` object. Is there any browsers that doesn't?
- The user uses browserify.
- The user expects `os.Hostname()` to not fail.

After this PR the result would be a catastrophic Javascript failure.
Before this PR the result would be a catastrophic gopherjs failure.

I think we can safely ignore this corner case, but I'm open for suggestions.

neelance:
@abrander What is your use case for `os.Hostname()`?

compiler/natives/os/os.go
flimzy: You're right, it would be perhaps an odd failure case in a browser. 

I think I was assuming (without closely reading your code) something more sophisticated than window.location, as that does not seem like the right host name to be returning. If the client host name cannot be determined, would it not make more sense to return '' or perhaps 'localhost' in the browser? 

On January 30, 2016 1:01:21 PM GMT+01:00, Anders Brander <notifications@github.com> wrote:
>> +// a non-working syscall on most (all?) platforms
>> +func Hostname() (string, error) {
>> +	// Browser
>> +	// The browser doesn't know anything about any hostname, so we use
>the host part of
>> +	// the current url.
>> +	if location := js.Global.Get("location"); location != js.Undefined
>{
>> +		if browserHostname := location.Get("hostname"); browserHostname !=
>js.Undefined {
>> +			return browserHostname.String(), nil
>> +		}
>> +	}
>> +
>> +	// Node.js
>> +	// We call node's require("os").hostname() which should return
>current hostname
>> +	if require := js.Global.Get("require"); require != js.Undefined {
>> +		// "os" is in nodejs core, it should always exist
>> +		if nodeHostname := require.Invoke("os").Call("hostname");
>nodeHostname != js.Undefined {
>
>For that to happen a few planets have to be aligned:
>
>- The browser doesn't implement the global `location` object. Is there
>any browsers that doesn't?
>- The user uses browserify.
>- The user expects `os.Hostname()` to not fail.
>
>After this PR the result would be a catastrophic Javascript failure.
>Before this PR the result would be a catastrophic gopherjs failure.
>
>I think we can safely ignore this corner case, but I'm open for
>suggestions.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/gopherjs/gopherjs/pull/385/files#r51344206
>
>!DSPAM:56aca6ec24651115619531!

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

abrander:
I encountered the problem when sharing MongoDB client code between backend and browser frontend. Mongo uses hostname to generate its "unique" ObjectId. It initializes a "machineId" no matter if you generate an ObjectId or not.

Relevant code:
http://bazaar.launchpad.net/+branch/mgo/v2/view/head:/bson/bson.go#L193
`os.Hostname()` call on line 203

I'm not sure patching gopherjs is the best way to solve this thou.

neelance:
How does the MongoDB code handle the case in which `os.Hostname()` returns an error?

abrander:
Mongo handles an error gracefully by generating a machineId with output from `"crypto/rand".Reader`.

The effect on Mongo is minimal. What annoyed me enough to implement this was the warning from gopherjs in the browser console:
```
```

neelance:
Hmm. What about adding a way to disable the warning instead?

compiler/natives/os/os.go
abrander: I'm not sure what the right thing to do is. We cannot return the correct `client` hostname in the browser no matter what we do.

I just wan't the warning to go away. An almost perfect implementation for me would be:
```go
// Make @abrander stop complaining :)
func Hostname() (string, error) {
	return "not-supported.example.com", nil
}
```

The current implementation is the most correct in browser land, but there's no reason that it shouldn't work as expected in Node land.

abrander:
Disabling the warning would be a workaround.

How about something like this?
```go
func Hostname() (string, error) {
	return "", ErrNotImplemented
}
```

The caller would get an error, we're not suppressing any warnings and everyone is happy. It's still not supported in Node land, but we're not worse off than before.

neelance:
It should work in node land if the syscalls module is available.

abrander:
Oh! I didn't know that.

I'm unsure how to move forward. Should we close this and move on to more important things?

neelance:
Yeah, let's close this for now. I'm still open to a way of disabling the warning.

PR #386, opened 2016-01-29
Implement runtime.NumCgoCall()

abrander:
We fake a cgo environment to catch errors. Therefor we have to implement this and always return 0. 

Files compiled with `// +build cgo` can expect this even when not using any cgo.


shurcooL:
@neelance, what do you think about this?

I know GopherJS does not support cgo (is this documented somewhere though?), so I'm unsure what to think here.

neelance:
I think it's fine to add this stub to be able to compile code that does not really use it at runtime. Will merge now. Thanks, @abrander.

PR #387, opened 2016-01-29
Added GCCPUFraction to runtime.MemStats

abrander:
The `runtime.MemStats` `GCCPUFraction` field  is expected by packages compiled for Go 1.5. This PR adds exactly that.

Please note that this PR does *not* update the field in any way. It will always be 0.0.


neelance:
Thanks a lot for the pull request!

PR #391, opened 2016-02-01
$externalize fixes

MJKWoolnough:
Found a couple of bugs in how $externalize converts structs to js objects.

First, when converting an anonymous field, it gives it a name of empty string. This is not ideal, and actually causes naming collisions if you have two or more anonymous fields.

The following playground link shows this off: -

<http://www.gopherjs.org/playground/#/paSROP9iN_>

The js console will show an object with two fields, the NotEmpty and the empty string struct, which will be C, which has overridden B.

The second occurs when you have an empty struct, searchJsObject doesn't handle the case where there are no fields to check.

This can be shown by just printing an empty struct{} to the log: -

http://www.gopherjs.org/playground/#/PfMcT_bbbz

### DIFF
--- a/compiler/prelude/jsmapping.go
+++ b/compiler/prelude/jsmapping.go
@@ -114,6 +114,9 @@ var $externalize = function(v, t) {
         return searchJsObject(v.$get(), t.elem);
       case $kindStruct:
         var f = t.fields[0];
+        if (f === undefined) {
+          return noJsObject;
+        }
         return searchJsObject(v[f.prop], f.typ);
       case $kindInterface:
         return searchJsObject(v.$val, v.constructor);
--- a/compiler/prelude/jsmapping.go
+++ b/compiler/prelude/jsmapping.go
@@ -132,7 +132,11 @@ var $externalize = function(v, t) {
       if (f.pkg !== "") { /* not exported */
         continue;
       }
-      o[f.name] = $externalize(v[f.prop], f.typ);
+      var name = f.name;
+      if (name === "") {
+        name = f.prop;
+      }
+      o[name] = $externalize(v[f.prop], f.typ);
     }
     return o;
   }


compiler/prelude/jsmapping.go
```diff
@@ -114,6 +114,9 @@ var $externalize = function(v, t) {
         return searchJsObject(v.$get(), t.elem);
       case $kindStruct:
         var f = t.fields[0];
+        if (f === undefined) {
```
neelance: Can we make that `t.fields.length === 0`? That captures the intention a bit better in my regard.

compiler/prelude/jsmapping.go
```diff
@@ -132,7 +135,11 @@ var $externalize = function(v, t) {
       if (f.pkg !== "") { /* not exported */
         continue;
       }
-      o[f.name] = $externalize(v[f.prop], f.typ);
+      var name = f.name;
+      if (name === "") {
+        name = f.prop;
```
neelance: We should probably not do this, because `f.prop` is an internal name and may be anything, e.g. it might be minified. What about doing the same as `encoding/json`: `Anonymous struct fields are usually marshaled as if their inner exported fields were fields in the outer struct, subject to the usual Go visibility rules`

compiler/prelude/jsmapping.go
MJKWoolnough: I agree that it looks better, but the defensive coder in me, that says not to trust user input, reminds me that we haven't checked that `t.fields` is an array; `t.fields.length` could be `undefined` and that does not equal 0.

We could have it check `t.fields.length > 0`, and reverse the `return` statements, although I'm not a fan of coercing `undefined`.

compiler/prelude/jsmapping.go
MJKWoolnough: I'm not well versed in minification, but I would have thought that strings (and all other constants) would have been left alone, especially with javascript where an object key can be referred to with and without a string. Also, would that same minification not also apply to f.name?

I do agree, though - using f.prop feels iffy, it just seemed to contain the correct value. Glancing at the gopherjs code, it seems that if the type name is a javascript keyword, it will append a $ and the struct position to the name. Not ideal, but, if reserved names aren't allowed, then there's not much alternative. Saying that, however, I was under the impression that you could use reserved words as object keys in javascript - that may need looking at.

Also, the encoding/json rules have issues. The problem is how to deal with multiple anonymous fields which share field names: -

https://play.golang.org/p/iv1x_3xVgr

...encoding/json library just removes the identically named fields, that doesn't seem appropriate for this case.

shurcooL:
This is a pretty old PR with no recent activity. Let's figure out what we can do with it: either resume making progress, or extract something useful out of it, or just close it.

@neelance There have been replies to your comments, can you see if that's enough to make progress?

MJKWoolnough:
The overriding of embedded fields now appears to have been fixed from the other end, as of commit [`640751f`](https://github.com/gopherjs/gopherjs/commit/640751f#diff-24e819883d03263dc0c2d37dcca1d2bd).

The `panic`ing when using `$externalize` on an empty struct still occurs, however, as the code incorrectly assumes that there will always be at least one field in a struct.

shurcooL:
Thanks for following up @MJKWoolnough, that makes sense to me.

> The `panic`ing when using `$externalize` on an empty struct still occurs, however, as the code incorrectly assumes that there will always be at least one field in a struct.

So we just need to focus on fixing this part:

> The second occurs when you have an empty struct, searchJsObject doesn't handle the case where there are no fields to check.
>
> This can be shown by just printing an empty struct{} to the log: -
>
> http://www.gopherjs.org/playground/#/PfMcT_bbbz

compiler/prelude/jsmapping.go
shurcooL: > but the defensive coder in me, that says not to trust user input

As far as I can tell, `fields` is not user input, it's a property that is generated by the GopherJS compiler. So it should be safe to rely on it existing.

If it always exists (I haven't confirmed, but that's my understanding), then `t.fields.length === 0` would be better because it captures the intention better (as Richard said) and makes it easier to read the code.

If it doesn't always exist, we should consider either making it always exist, even if it's empty for empty structs, or consider going with `if t.fields === undefined` or so.

Basically, I suggest we take the least amount of precautions that are needed to fix the issue.

We can add a test to make sure the behavior doesn't regress. Having more readable code will help with future development and maintenance. What do you think @MJKWoolnough?

flimzy:
As this has been partly fixed, and the remainder is now tracked separately in #715, I'm going to close this PR. It's still available for reference when we get to fixing #715.

PR #393, opened 2016-02-09
Fix "shorthand redefinition" panic when running `gopherjs tool` command.

### LINKED ISSUE #392: gopherjs tool panics with "shorthand redefinition"
When I run `gopherjs tool`, it panics with `"shorthand redefinition"` from the `github.com/spf13/flag` package.

shurcooL:
Fixes #392.

/cc @neelance I have no idea if this `gopherjs tool` command is even supported, or what it's meant to do with all those flags. But this fixes the panic.


neelance:
`gopherjs tool` is not officially supported. It was initially used for running core tests.

PR #406, opened 2016-02-18
compiler/natives/runtime: Add missing function body for ReadTrace.

shurcooL:
Addresses https://github.com/gopherjs/gopherjs/commit/a6b5a2662eeefb2063c09e39baf4104828f3d589#commitcomment-16172129.

Needs review. (It's possible I'm misunderstanding what's going on and the missing func body is valid (in that case, a comment is likely worthwhile), but given the history the change, this seems unlikely.)

### DIFF
--- a/compiler/natives/runtime/runtime.go
+++ b/compiler/natives/runtime/runtime.go
@@ -174,7 +174,7 @@ func Version() string {
 
 func StartTrace() error { return nil }
 func StopTrace()        {}
-func ReadTrace() []byte
+func ReadTrace() []byte { return nil }
 
 // We fake a cgo environment to catch errors. Therefor we have to implement this and always return 0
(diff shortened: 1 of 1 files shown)


neelance:
A missing function body is valid for functions implemented in assembly. GopherJS compiles this to a function that panics, see https://github.com/gopherjs/gopherjs/blob/master/compiler/package.go#L595. I think having no function body for `ReadTrace` was no conscious decision, but a panic might be better than returning nil, since the feature is clearly not supported. What do you think?

shurcooL:
So this was intentional after all. That means this PR is invalid, since it was based on the assumption that it wasn't intentional.

Panic is definitely better than returning an empty byte slice. It might've been more readable just to implement a body that panics, but now that I'm aware of this possibility, it is okay as is.

PR #408, opened 2016-02-22
README: Document gopherjs serve command.

### LINKED ISSUE #268: README does not mention gopherjs serve mode.
`gopherjs serve` is one of most convenient ways to work with GopherJS in most cases (rather than `gopherjs build` or `gopherjs install`, yet the README does not mention it at all.

shurcooL:
Fixes #268.

The `gopherjs serve` command was originally added in https://github.com/gopherjs/gopherjs/issues/121, motivation and background can be found there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,14 @@ For system calls (file system access, etc.), see [this page](https://github.com/
 
 *Note: GopherJS will try to write compiled object files of the core packages to your $GOROOT/pkg directory. If that fails, it will fall back to $GOPATH/pkg.*
 
+#### gopherjs serve
+
+`gopherjs serve` is a useful command you can use during development. It will start an HTTP server serving on ":8080" by default, and dynamically compile Go packages with GopherJS and serve them.
+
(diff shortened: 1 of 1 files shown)


shurcooL:
An alternative text I created first, before refactoring it, was:

> For example, an import path `example.com/user/project` will be served at http://localhost:8080/example.com/user/project/project.js. It will also serve all static files in those directories. If the directory contains `index.html`, it will be served, otherwise a minimal `index.html` that includes `<script src="{{base}}.js"></script>` will be served, so navigating to http://localhost:8080/example.com/user/project/ will execute your code.

It tried to explain how it works first, and what the end result is last. I re-arranged it so that it starts by saying what works, and then goes into details of how it works exactly. I hope that's an improvement.

neelance:
LGTM

gorakhargosh:
Do you have a working example of this?

baram204:
@gorakhargosh 

if `$GOPATH ` is ``c:\Users\someone\go\` and your gopherjs files at `c:\Users\someone\go\src\project\src\client\main.go`.

then `gohperjs serve` works.

`http://localhost:8080/project/src/client/main.go` works fine.

but, `{{base}}` not replaced actual filename.  I don't know what base value is.

shurcooL:
> but, `{{base}}` not replaced actual filename.  I don't know what base value is.

"{{base}}.js" was just a way of suggesting that the name of the file is the base of the directory. E.g., if the file is `http://localhost:8080/project/src/client/main.go`, the server JS file will be `http://localhost:8080/project/src/client/client.js`.

That's because the base of path "http://localhost:8080/project/src/client/main.go" is "client". This is similar to how `go install` works for commands, it uses the base of the directory as the command name. See https://godoc.org/path#Base for reference.

baram204:
@shurcooL Thank you!

PR #409, opened 2016-02-24
Don't panic  if import returns without package

tgulacsi:
gopherjs build panic'ed when couldn't find some imports. With this fix, it nicely prints the missing package.

### DIFF
--- a/compiler/package.go
+++ b/compiler/package.go
@@ -113,6 +113,9 @@ func (pi packageImporter) Import(path string) (*types.Package, error) {
 		}
 		return nil, err
 	}
+	if a == nil {
+		return nil, fmt.Errorf("cannot find %q", path)
+	}
 
 	return a.types, nil
 }


compiler/package.go
```diff
@@ -113,6 +113,9 @@ func (pi packageImporter) Import(path string) (*types.Package, error) {
 		}
 		return nil, err
 	}
```
shurcooL: The `a, err` are coming from `pi.importContext.Import(path)`.

Via some indirection, that ends up calling:

https://github.com/gopherjs/gopherjs/blob/f8510c9aef19750ec5a1178b33932a7f9defd197/build/build.go#L401-L418

Which is code we control. I think the root problem is somewhere else; when returning `pkg.Archive, nil`, the value of `pkg.Archive` should not be `nil` I would imagine. We should find out why it's possible for it to be.

This fix may work as a temporary solution, but it's better to fix the issue at the root instead (given my assumption that a `nil` value for `pkg.Archive` should never occur under normal conditions).

neelance: I agree with @shurcooL. How can we reproduce this?

tgulacsi: See Issue #410. For example honnef.co/... is missing.

Richard Musiol <notifications@github.com> ezt írta (időpont: 2016. febr.
24., Sze, 12:43):

tgulacsi:
I can confirm that actual master (1a0f90fdef7bbf5d894d1651a9db7e2cb6a5b62f) works flawlessly (does not panic)
```
:gthomas@waterhouse: ~/src/github.com/influx6/govfx/examples/shadowDOM
$ gopherjs build
cannot find package "github.com/fatih/camelcase" in any of:
        /usr/local/go/src/github.com/fatih/camelcase (from $GOROOT)
        /home/gthomas/src/github.com/fatih/camelcase (from $GOPATH)

```

PR #413, opened 2016-02-29
added cache for local import paths

neelance:


build/build.go
```diff
@@ -505,11 +505,19 @@ func (s *Session) BuildPackage(pkg *PackageData) (*compiler.Archive, error) {
 		return nil, err
 	}
 
+	localImportPathCache := make(map[string]*compiler.Archive)
```
shurcooL: Map key should be documented.

shurcooL: Unless you think `localImportPath` in the variable name is descriptive enough?

neelance: I think so.

shurcooL: So is it the full import path without /vendor/ stripped?

PR #429, opened 2016-03-22
Add optional root argument to gopherjs serve. Fixes #384

### LINKED ISSUE #384: Feature Request: gopherjs serve import/path
It's convenient that using `gopherjs serve` makes all of your GOPATH workspace available in browser via URLs like http://localhost:8080/import/path.

However, some GopherJS projects use the URL as part of the input and may not expect an import path prefix, which would cause unintended issues for it.

I propose having a flag/value that can be passed to `gopherjs serve` that restricts serving to the specified import path (it doesn't have to be a valid Go package at that location, it's just a path).

For example, if my frontend project is at `example.com/foo/bar` I want the URL to be http://localhost:8080/index.html, then I could use:

```bash
gopherjs serve example.com/foo/bar
```

Otherwise, if I used `gopherjs serve` as is now, I'd have to visit http://localhost:8080/example.com/foo/bar/index.html instead.

jargv:
Fixes #384.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ For example, navigating to `http://localhost:8080/example.com/user/project/` sho
 
 Refreshing in the browser will rebuild the served files if needed. Compilation errors will be displayed in terminal, and in browser console. Additionally, it will serve $GOROOT and $GOPATH for sourcemaps.
 
+If you include an argument, it will be the root from which everything is served. For example, if you run gopherjs serve github.com/user/project then the generated JavaScript for the package github.com/user/project/mypkg will be served at http://localhost:8080/mypkg/mypkg.js.
+
 ### Performance Tips
 
 - Use the `-m` command line flag to generate minified code.
--- a/tool.go
+++ b/tool.go
@@ -563,10 +563,7 @@ type serveCommandFileSystem struct {
 }
 
 func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
-	name := requestName[1:] // requestName[0] == '/'
-	if !strings.HasSuffix(requestName, ".go") {
-		name = path.Join(fs.serveRoot, name)
-	}
+	name := path.Join(fs.serveRoot, requestName[1:]) // requestName[0] == '/'
 
 	dir, file := path.Split(name)
 	base := path.Base(dir) // base is parent folder name, which becomes the output file name.
@@ -627,7 +624,15 @@ func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
 	}
 
 	for _, d := range fs.dirs {
-		f, err := http.Dir(filepath.Join(d, "src")).Open(name)
+		dir := http.Dir(filepath.Join(d, "src"))
+
+		f, err := dir.Open(name)
+		if err == nil {
+			return f, nil
+		}
+
+		// source maps are served outside of serveRoot
+		f, err = dir.Open(requestName)
 		if err == nil {
 			return f, nil
 		}
--- a/tool.go
+++ b/tool.go
@@ -563,7 +563,11 @@ type serveCommandFileSystem struct {
 }
 
 func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
-	name := path.Join(fs.serveRoot, requestName[1:]) // requestName[0] == '/'
+	name := requestName[1:] // requestName[0] == '/'
+	if !strings.HasSuffix(requestName, ".go") {
+		name = path.Join(fs.serveRoot, name)
+	}
+
 	dir, file := path.Split(name)
 	base := path.Base(dir) // base is parent folder name, which becomes the output file name.
 
--- a/tool.go
+++ b/tool.go
@@ -485,7 +485,7 @@ func main() {
 	}
 
 	cmdServe := &cobra.Command{
-		Use:   "serve",
+		Use:   "serve [root]",
 		Short: "compile on-the-fly and serve",
 	}
 	cmdServe.Flags().AddFlag(flagVerbose)
@@ -498,7 +498,24 @@ func main() {
 	cmdServe.Run = func(cmd *cobra.Command, args []string) {
 		options.BuildTags = strings.Fields(*tags)
 		dirs := append(filepath.SplitList(build.Default.GOPATH), build.Default.GOROOT)
-		sourceFiles := http.FileServer(serveCommandFileSystem{options: options, dirs: dirs, sourceMaps: make(map[string][]byte)})
+		var root string
+
+		if len(args) > 1 {
+			cmdServe.Help()
+			return
+		}
+
+		if len(args) == 1 {
+			root = args[0]
+		}
+
+		sourceFiles := http.FileServer(serveCommandFileSystem{
+			serveRoot:  root,
+			options:    options,
+			dirs:       dirs,
+			sourceMaps: make(map[string][]byte),
+		})
+
 		ln, err := net.Listen("tcp", addr)
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err)
@@ -539,12 +556,14 @@ func (ln tcpKeepAliveListener) Accept() (c net.Conn, err error) {
 }
 
 type serveCommandFileSystem struct {
+	serveRoot  string
 	options    *gbuild.Options
 	dirs       []string
 	sourceMaps map[string][]byte
 }
 
-func (fs serveCommandFileSystem) Open(name string) (http.File, error) {
+func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
+	name := path.Join(fs.serveRoot, requestName[1:]) // requestName[0] == '/'
 	dir, file := path.Split(name)
 	base := path.Base(dir) // base is parent folder name, which becomes the output file name.
 
@@ -555,7 +574,7 @@ func (fs serveCommandFileSystem) Open(name string) (http.File, error) {
 	if isPkg || isMap || isIndex {
 		// If we're going to be serving our special files, make sure there's a Go command in this folder.
 		s := gbuild.NewSession(fs.options)
-		pkg, err := gbuild.Import(path.Dir(name[1:]), 0, s.InstallSuffix(), fs.options.BuildTags)
+		pkg, err := gbuild.Import(path.Dir(name), 0, s.InstallSuffix(), fs.options.BuildTags)
 		if err != nil || pkg.Name != "main" {
 			isPkg = false
 			isMap = false


neelance:
This looks really good, thanks for contributing! Two small wishes:
1. Make `gopherjs help serve` show the new optional argument.
2. Print the same help content if more than one argument is given.

shurcooL:
Thanks for working on resolving that issue! I have some questions about the code changes.

tool.go
```diff
@@ -555,7 +571,7 @@ func (fs serveCommandFileSystem) Open(name string) (http.File, error) {
 	if isPkg || isMap || isIndex {
 		// If we're going to be serving our special files, make sure there's a Go command in this folder.
 		s := gbuild.NewSession(fs.options)
-		pkg, err := gbuild.Import(path.Dir(name[1:]), 0, s.InstallSuffix(), fs.options.BuildTags)
+		pkg, err := gbuild.Import(path.Dir(strings.Trim(name, "/")), 0, s.InstallSuffix(), fs.options.BuildTags)
```
shurcooL: What's the reason for this change?

tool.go
```diff
@@ -498,7 +498,21 @@ func main() {
 	cmdServe.Run = func(cmd *cobra.Command, args []string) {
 		options.BuildTags = strings.Fields(*tags)
 		dirs := append(filepath.SplitList(build.Default.GOPATH), build.Default.GOROOT)
-		sourceFiles := http.FileServer(serveCommandFileSystem{options: options, dirs: dirs, sourceMaps: make(map[string][]byte)})
+		root := ""
```
shurcooL: Optional style suggestion. Since you you're not trying to set a particular initial value for `root`, it might be more idiomatic to write this in the `var root string` form.

shurcooL:
How does this change affect the source maps that are served as well?

tool.go
jargv: When the first character was unconditionally dropped with `name[1:]` it was always a slash. With this change the first character is sometimes a slash, but not always. I was seeing it pass strings like `ithub.com/jargv/...` into path.Dir, which fails, of course. `strings.Trim` seemed like a good way to conditionally drop the first character if it's a '/'. I'm new(ish) to go so I'd love to learn if there's a more idiomatic approach. :)

tool.go
jargv: Since i'm new to Go, I'll take your word for it, but that is surprising to me. Is the explicitly typed form considered idiomatic even in the case of simple values? I recognize the value of leaning on zero values (a technique that I'm a fan of from C/C++), but in this case, the explicit assignment is shorter and no less clear. Am I missing something?

jargv:
@shurcooL Source maps seem to work as expected with this change.

tool.go
shurcooL: I can point you to [this article](http://dave.cheney.net/2014/05/24/on-declaring-variables) by Dave Cheney that lists some suggestions for variable declaration styles and when to use them. I think they're good suggestions and they match well most of the idiomatic Go code I see.

Both forms are okay, but I think `var root string` is slightly nicer. It emphasizes that you're just declaring a variable and not initializing it with an important value.

You can see examples of this style in the same file, e.g., [here](https://github.com/gopherjs/gopherjs/blob/2fecf3fff15b993f5ea9c095868ef1ac0ba6a9a8/tool.go#L404).

jargv: That's a good point about emphasizing that the exact value is unimportant. I'll make that change. Thanks for the feedback :D

tool.go
shurcooL: > With this change the first character is sometimes a slash, but not always.

I see. I asked I'd like to understand exactly in what situations it'a slash, and in what situations it's not.

> `strings.Trim` seemed like a good way to conditionally drop the first character if it's a '/'. I'm new(ish) to go so I'd love to learn if there's a more idiomatic approach. :)

One immediate improvement suggestion would be to use [`strings.TrimLeft`](https://godoc.org/strings#TrimLeft) rather than `strings.Trim`, since you said you only want to drop the first character being `/`, not the last character.

However, I understand what's happening better now, and I have a suggestion for how you can go back to just having an unconditional trim of first character (which should be simpler to reason about).

The signature of this method used to be `Open(name string)`, and `name[1:]` used to be done to drop the first "/" that's a part of each `*http.Request.URL.Path`. Since you've re-arranged the code to be `Open(requestName string)`, I think you can move the the `[:1]` operation to happen earlier:

```Go
func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
	name := path.Join(fs.serveRoot, requestName[1:])
```

Then this line can become simply:

```Go
pkg, err := gbuild.Import(path.Dir(name), 0, s.InstallSuffix(), fs.options.BuildTags)
```

What do you think of that?

tool.go
neelance: Is the value really unimportant? If I would replace it with `var root string = "foo"`, then the case with `len(args) == 0` would be incorrect, right?

shurcooL: @neelance, you're right, it's actually import for it to be the zero value.

I misread the code below, thinking it sets `root` in all cases, but it only sets it when `len(args) == 1`.

Given that's the case (that value of `root` must be zero value for string), I think both styles are okay.

tool.go
jargv: Yes, I agree TrimLeft would be better, and the move you suggest would be better still! I'll make that change shortly.

README.md
```diff
@@ -35,6 +35,8 @@ For example, navigating to `http://localhost:8080/example.com/user/project/` sho
 
 Refreshing in the browser will rebuild the served files if needed. Compilation errors will be displayed in terminal, and in browser console. Additionally, it will serve $GOROOT and $GOPATH for sourcemaps.
 
+If you include an argument, it will be the root from which everything is served. For example, if you run `gopherjs serve example.com/user` then the generated JavaScript will be served at `http://localhost:7979/project/project.js`.
```
dominikh: I don't understand how the command `gopherjs serve example.com/user` maps to something being served at `/project/`. 

I also feel like making it a positional argument is confusing. Other tools (build, install) take the package as an argument. I can accept `serve` not accepting a package, but IMO it shouldn't interpret the argument in an entirely different way. Wouldn't it make more sense to make this an explicit flag?

tool.go
```diff
@@ -498,7 +498,21 @@ func main() {
 	cmdServe.Run = func(cmd *cobra.Command, args []string) {
 		options.BuildTags = strings.Fields(*tags)
 		dirs := append(filepath.SplitList(build.Default.GOPATH), build.Default.GOROOT)
-		sourceFiles := http.FileServer(serveCommandFileSystem{options: options, dirs: dirs, sourceMaps: make(map[string][]byte)})
+		var root string
+		if len(args) == 1 {
```
dominikh: I'd rather write it like this:

```
if len(args) > 1 {
  cmdServe.Help()
  return
}

if len(args) == 1 {
  root = args[0]
}
```

@shurcooL, any opinion on this?

README.md
neelance: What about changing it to this:

"For example, if you run `gopherjs serve github.com/user/project` then the generated JavaScript for the package `github.com/user/project/mypkg` will be served at `http://localhost:7979/mypkg/mypkg.js`."

I think the semantics are not too far off compared to other commands.

dominikh: I guess I don't understand what the argument does. I thought `gopherjs serve foo` would serve the package in $PWD at `/foo/`.

tool.go
shurcooL: That sounds better, since it handles errors first. That is more in line with the recommended style, see https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

README.md
shurcooL: @dominikh, no, it does what @neelance described more clearly above. Maybe see #384 that this PR resolves for more context.

dominikh: Hm, okay. I guess it's fine then.

jargv: This does behave like other commands which take a package. It just _also_ handles the case where the argument is not a complete package path, in which case it serves all packages at that root.

tool.go
```diff
@@ -560,7 +560,7 @@ type serveCommandFileSystem struct {
 }
 
 func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
-	name := path.Join(fs.serveRoot, requestName)
+	name := path.Join(fs.serveRoot, requestName[1:]) //requestName[0] == '/'
```
shurcooL: In Go, the convention for user-facing comments is to have a single space between the `//` (or `/*`) and the text. So this comment should be:

```Go
// requestName[0] == '/'
```

The only times not to have a space is when it's a temporarily commented out development code (usually, that code is not committed, it's a temporary state during development) and for commands meant for non-compiler programs. For example, `//go:generate ...`.

README.md
```diff
@@ -35,6 +35,8 @@ For example, navigating to `http://localhost:8080/example.com/user/project/` sho
 
 Refreshing in the browser will rebuild the served files if needed. Compilation errors will be displayed in terminal, and in browser console. Additionally, it will serve $GOROOT and $GOPATH for sourcemaps.
 
+If you include an argument, it will be the root from which everything is served. For example, if you run gopherjs serve github.com/user/project then the generated JavaScript for the package github.com/user/project/mypkg will be served at http://localhost:7979/mypkg/mypkg.js.
```
neelance: One last thing: Let's use `:8080` here, since that is the default.

neelance:
I'd like the port number changed, see above. LGTM otherwise. @dominikh? @shurcooL?

dominikh:
LGTM

shurcooL:
I can see some minor followup enhancements, but no blockers for this getting merged to resolve #384.

LGTM.

jargv:
I found an issue with sourcemaps that needs to be resolved before this can be merged.

The source maps work well enough to map errors to the right file/line. So for example instead of the error being reported at `output.js:15912`  it is reported correctly at `main.go:7`.

The problem occurs when I click on `main.go:7`, the devtools can't load `main.go` to display the error in the context of the file. I hope to have this fixed soon.

jargv:
I added a fix but it has an assumption. When the requested file ends in ".go" I ignore the serveRoot. The assumption is that source maps will only map to `.go` files, and `.go` files will only be requested because the source maps reference them. Is that true?

flimzy:
GopherJS also accepts *.inc.js files as input. I don't know if these are presently supported for source maps. If it's easy to support such files as well, I would encourage doing so. If that's presently not supported, then at least it's not a regression not to support them here.

jargv:
@flimzy source maps don't appear to work for .inc.js files.

neelance:
What about trying both? First try the prefixed path, but if that file does not exist then try the non-prefixed path? I think this should be possible by adding to `serveCommandFileSystem.dirs`. And also for the implementation in general, it might be easier to prefix entries of `dirs` instead of adding `serveRoot`. I just realized that now, sorry for not mentioning it earlier.

jargv:
@neelance I actually tried that first. The trouble I ran into was how serveCommandFileSystem.Open determines when to compile or not. Specifically these variables never evaluated to true: 
```go
isPkg := file == base+".js"
isMap := file == base+".js.map"
```
because base is derived from the request path.

Of course this could be changed to work in some other way, but it's not clear to me that that would be simpler. For example, how would you disambiguate a request to `"/"`? Should that compile the package at serveRoot, or serve the directory listing with http.Dir(...).Open(...)?

That said, I'm happy to make the change if that's determined to be the best way forward.

jargv:
Ok I changed it to serve files both inside and outside of the serveRoot. There's no longer special casing around `.go` files. 

@neelance is this what you meant? I tried adding to serveCommandFileSystem.dirs in order to get rid of the serveRoot variable, but the problem becomes mapping back to a valid package path. Right now we can assume that the request path (with the serve root prepended) is a valid package without having to look through GOPATH and GOROOT looking for a match. 

Anything else to make this good to merge?

PR #445, opened 2016-04-27
perform implicit casting for tuple returns if needed (fixes #444)

### LINKED ISSUE #444: unknown issue related to javascript 'undefined' values
When doing a type conversion from `bool` to `interface{}` as part of a return statement, the code does not behave as expected.

See: http://www.gopherjs.org/playground/#/GM1vWfnqn_

Note that if you change the runPredicate to store the values in an intermediate before passing it on, it behaves as expected.  Also note that if the predicate test method is single-value rather than multi-value, it works as well.

brett19:


brett19:
P.S. In terms of the 'danger' of doing explicit type-casting without being sure.  The new method explicitly assumes that results[0] is a tuple.  In the previous version of this method, it implicitly assumed this was the case (and would fail at runtime if it was not).

brett19:
In addition to the above, the code assumes that the expression ends up being a single identifier (_r for instance).  Originally it mapped this into a new temporary variable before mapping it out for implicit casting, but this caused some weird code to be generated.  I don't believe there are cases where we would get to this point in the code with something other than a single identifier, but it might be nice to get someone to confirm this.

Code Generated Now:
```javascript
_r = SomeCall();
return [new $Bool(_r[0]), _r[1]]
```
Alternative I chose not to commit:
```javascript
_r = SomeCall();
_tmp = _r;  // Possibly useful in the case that the expression is not as trivial as `_r`.
return [new $Bool(_tmp[0]), _tmp[1]]
```

neelance:
Thanks for contributing! Some remarks:
* CI is badly broken. Seems like there is stuff to fix. :-)
* I think you can use `types.Identical` directly on the tuple.
* The third comment is helpful, the other two are quite redundant to adjacent lines

brett19:
@neelance Thanks for the suggestions.  Should be all sorted.

compiler/statements.go
```diff
@@ -748,7 +748,23 @@ func (c *funcContext) translateResults(results []ast.Expr) string {
...
+				panic("invalid tuple return assignment")
+			}
+
+			resultExpr := c.translateExpr(results[0]).String()
+
+			if types.Identical(resultTuple, tuple) {
+				return " " + resultExpr
+			}
+
+			// Not all the return types matched, lets map everything out for implicit casting
```
shurcooL: Minor typo here, I think you meant to write `let's`.

brett19: You're correct.  I actually simply removed the word as it was unneccessary.

neelance:
Cool! Can we have a regression test for #444 in misc_test.go? Please make it minimal.

brett19:
@neelance That should do it.  I'll assume that the test runner will bail from a JS exception being thrown.

brett19:
@neelance @shurcooL Can either of you comment in regards to my concerns that I posted in the first two comments to this pull request?

neelance:
Have you tried commenting out your changes to see the test failing?

Your concern about `_r` is valid. You only have this if the call is considered blocking. All `func` values are considered blocking. Try to come up with a test case that does not involve `func` values, just static function calls. This will probably give a scenario that needs a proper temporary variable.

brett19:
@neelance I can't actually think of any other cases which cause this particular scenario to occur.

neelance:
Try this test:
```

func TestTupleReturnImplicitCast(t *testing.T) {
	x, _ := tuple1()
	if x != 14 || tuple2called != 1 {
		t.Fail()
	}
}

func tuple1() (interface{}, error) {
	return tuple2()
}

var tuple2called = 0

func tuple2() (int, error) {
	tuple2called++
	return 14, nil
}
```

brett19:
@neelance Would you say that it's worth trying to cast away the extra variable assignment that occurs in the original test case? IE:
```javascript
		_r = fn();
		_tuplecast = _r;
		return [new $Int(_tuplecast[0]), _tuplecast[1]];
```

brett19:
Improved the test coverage, and fixed the issue you identified.

neelance:
Cool! Did you delete the other comment? My answer would have been: I don't mind the extra variable, because this case of a tuple not matching is pretty rare.

LGTM
@shurcooL PTAL

brett19:
@neelance I did remove the other comment as there was no sane way for me to actually detect that it was trivial.  As a side-note.  It would be nice if there was a better way to manage case scoping throughout Gopher, I see some wierd cases where the cases being generated pretty unoptimally.

neelance:
Yeah, there is still a lot of room for improvements.

shurcooL:
> Did you delete the other comment? My answer would have been: I don't mind the extra variable, because this case of a tuple not matching is pretty rare.

I think it's still there. He asked that in 3rd https://github.com/gopherjs/gopherjs/pull/445#issuecomment-215242353 from the top:

> Code Generated Now:
> ```javascript
> _r = SomeCall();
> return [new $Bool(_r[0]), _r[1]]
> ```
> Alternative I chose not to commit:
> ```javascript
> _r = SomeCall();
> _tmp = _r;  // Possibly useful in the case that the expression is not as trivial as `_r`.
> return [new $Bool(_tmp[0]), _tmp[1]]
> ```

tests/misc_test.go
```diff
@@ -535,3 +535,53 @@ func TestTrivialSwitch(t *testing.T) {
...
+	return 14, nil
+}
+func TestTupleReturnImplicitCast(t *testing.T) {
+	x, _ := tuple1()
+	if x != 14 || tuple2called != 1 {
+		t.Fail()
+	}
+}
+
+func TestDeferNamedTupleReturnImplicitCast(t *testing.T) {
```
shurcooL: Is this `TestDeferNamedTupleReturnImplicitCast` test here "just in case" of future regressions?

I ask, because this test is currently passing on `master`, without the fix applied in `compiler/statements.go`. I just want to confirm that was intentional.

shurcooL:
Left one inline question, but otherwise LGTM.

tests/misc_test.go
brett19: Hey @shurcooL,
I didn't actually check whether that test failed on the current master, I did however check the to see where the code I modified would actually be used form and I saw a special case when dealing with deferred parameters.  Additionally, the logic I modified was affected by named parameters as well.  I chose to put those two tests together and include it to ensure that case was tested, as the original singular test might have missed it.
tl;dr: Yea, entirely to catch future regressions from other people modifying the behaviour of the things surrounding my changes.
Cheers, Brett

neelance:
Merged. @brett19 Thx again!

Issue #388, opened 2016-01-30
proposal: Replace in-house implementation of `gopherjs` build tool with augmented forked `cmd/go` build tool.


shurcooL:
The GopherJS project has developed the `gopherjs` tool. Its CLI mimics a lot of the commands that `go` tool (import path `cmd/go`, in GOROOT) offers, in addition to one new command `serve` (see #268). However, it's missing some features of the real `cmd/go` (see [issues labeled cmd/go](https://github.com/gopherjs/gopherjs/labels/cmd%2Fgo) for some examples).

Instead of continuing to develop and maintain the `gopherjs` tool ourselves, I propose we replace the custom implementation of `gopherjs` with a modified fork of the real `cmd/go` that has 2 changes cherry-picked on top. All existing functionality would be preserved, gaining missing features, fixing bugs, and reducing overall maintenance cost. The only functionality that the current `gopherjs` tool offers but `cmd/go` doesn't is the `serve` command, and we will preserve it in some way (details to be determined).
### Plan

At a high level, my proposed plan is:
1. Wait for Go 1.6 release, or combine this plan with adding suppport for Go 1.6.
2. Create a fork of `cmd/go`.
3. Create a change that can be cherry-picked (or rebased) on top of latest upstream `cmd/go` that makes the command go-gettable.
   - This is a small change.
4. Create a change that can be cherry-picked (or rebased) on top that adds support for -compiler=gopherjs value.
   - This is a larger change, but it can be done cleanly and will result in significant simplification in the work that GopherJS needs to do, stripping away a lot of its responsibility as a build tool, leaving only the compilation and linking to be done.
5. Continue to make the augmented `cmd/go` available and maintain it by simply taking the latest upstream code and rebasing the 2 cherry-picked changes on top.
   - When the plan is complete and the changes are reviewed and made clean, and everything proves to be stable, we can attempt to make a CL to merge the -compiler=gopherjs support addition upstream into core Go repository, and learn if there are any blockers from making that possible. If not accepted, we can continue to maintain the forked tool ourselves with low maintenance overhead.

The end result will be a fully capable `cmd/go`-like command that will support compiling with `GOARCH=js` set. It will be able to `go run` and `go test` by setting `-exec=node`. It will resolve all [these issues](https://github.com/gopherjs/gopherjs/labels/cmd%2Fgo) and simplify original code that GopherJS project has to maintain.
### Research and viability

[...]

PR #449, opened 2016-04-30
WIP: Install and use source-map-support module for tests in Circle CI.

shurcooL:
**DO NOT MERGE.** There is an outstanding task to be resolved before this is ready.

@neelance What do you think about requiring source-map-support for tests? Right now they're disabled in Circle CI. Enabling it fixes one failing test from Go compiler test suite (fixedbugs/issue5856.go).

This PR:
- ~~Updates to Go 1.6.2.~~ **Edit:** This has been cherry-picked as 5fbc21536734f6804dacf5dafb9d24eb64eabed9.
- ~~Updates to Node 6.0.~~ **Edit:** Factored out into https://github.com/gopherjs/gopherjs/pull/484.
- Installs and uses source-map-support module in Circle CI.
- ~~Contains other minor miscellaneous improvements (see commits fdad0efda96882075c2edb6cf17fe6e9ce686bb4 and fcc469a4d7af30aae23e1f854ea50d5048583de5).~~ **Edit:** These have been cherry-picked as e28f4b49efae2d138c419f51a6a49a7fd5a46d7f and 028c56b1a74aa0b24f8861f78bceafa5bbcd3eb8.


tests/gorepo_test.go
```diff
@@ -5,11 +5,19 @@ package tests_test
...
 	"os"
 	"os/exec"
+	"strconv"
 	"testing"
 )
 
 // Go repository basic compiler tests, and regression tests for fixed compiler bugs.
 func TestGoRepositoryCompilerTests(t *testing.T) {
+	if b, _ := strconv.ParseBool(os.Getenv("SOURCE_MAP_SUPPORT")); os.Getenv("SOURCE_MAP_SUPPORT") != "" && !b {
+		t.Fatal("Source maps disabled, but required for this test. Use SOURCE_MAP_SUPPORT=true or unset it completely.")
```
neelance: t.Skip?

neelance:
LGTM, one comment.

tests/gorepo_test.go
shurcooL: I considered skip, but was worried it might be bad to simply skip such an important test so easily.

Hmm...

Maybe skip if short flag is given?

neelance: > Maybe skip if short flag is given?

I don't think that would be good, better be consistent.

I think I've seen `Skip` in the core packages for this situation. Reasoning might go like this:
Fail -> Implementation broken
Skip -> Environment broken/insufficient

shurcooL: I can change to skip, sure, I just don't want us to get into a situation where those tests are always skipped and we don't notice.

shurcooL: I've looked in https://github.com/golang/go/blob/master/src/cmd/go/go_test.go to see when skip is used.

It's used to skip individual specific tests when environment is insufficient. For example, skip testing cgo when when cgo is not enabled.

The difference is that `TestGoRepositoryCompilerTests` runs many, many tests. Only one of those actually depends on source-map-support to be enabled. Skipping all of `TestGoRepositoryCompilerTests` just because source-map-support is equivalen to skipping _all_ of `cmd/go` tests if cgo is not enabled. I don't think it's acceptable.

Let's find a better solution...

shurcooL: In the PR description, I asked:

> What do you think about requiring source-map-support for tests? Right now they're disabled in Circle CI. Enabling it fixes one failing test from Go compiler test suite (fixedbugs/issue5856.go).

Maybe it's not worth it then to require source-map-support to be mandatory for tests?

neelance: I think the most important aspect is that the CI system tests properly. Local testing has a much lower priority and I would be okay with skipping tests in that situation. If people download the source code, run the tests and they fail, then they will probably think that the tests are broken.

shurcooL: I'll find an acceptable solution and update this PR; let's not merge it until then.

shurcooL:
I've edited the PR description to reflect which parts of this PR have already been cherry-picked into `master`. What's left are updates to Node 6.0 and installing/using source-map-support module in Circle CI.

I will get to resolving the remaining task needed to merge this sometime, but it's lower priority than #388.

shurcooL:
I've factored out the Node update (and updated it from 6.0.0 to latest 6.2.2 version) into https://github.com/gopherjs/gopherjs/pull/484.

neelance:
@shurcooL What's up with this PR?

shurcooL:
The status continues to be the same as in [Jun 26, 2016](https://github.com/gopherjs/gopherjs/pull/449#issuecomment-228637693). Except now it's higher priority than #388. :)

I still need to think about the best thing to do about source-map-support module and tests. I'll do that, and update the PR.

shurcooL:
I'm going to close this. I still haven't found a way to get this PR to a point where I'm happy with it to be merged. If I ever do that, I'll reopen.

PR #453, opened 2016-05-08
compiler/natives/syscall: Don't import bytes package.

shurcooL:
The real syscall package doesn't import bytes. This keeps the order in which stdlib packages must be built in for GopherJS consistent with the real standard library.

Source: https://godoc.org/syscall?imports

Helps #388.


neelance:
LGTM

PR #454, opened 2016-05-08
compiler/natives/src/net/http: Implement http.RoundTripper on Fetch API.

shurcooL:
This is a WIP PR in order to get initial review comments, to communicate and avoid duplicate efforts. Known TODOs are listed in code.

The motivation to use Fetch API is because it's lower level, and allows implementing an http.RoundTripper more efficiently (less conversions between strings and []byte). It also supports streaming body responses efficiently.

References:

- https://fetch.spec.whatwg.org/
- https://streams.spec.whatwg.org/
- http://caniuse.com/#search=Fetch

Tested in stable channel of Chrome, latest Safari, Firefox (developer edition).

Dynamically determine which of XHR, Fetch APIs are available and gracefully fallback to the best available API.

Here's a quick demo, using a local instance of https://http2.golang.org/reqinfo and https://http2.golang.org/clockstream endpoints. I've asked @bradfitz if it's possible to enable CORS on http2.golang.org itself, and if so, you could try this out in a playground.

```Go
func streamingDemo() error {
	resp, err := http.Get("https://localhost:4430/clockstream")
	if err != nil {
		return err
	}
	defer resp.Body.Close()
	_, err = io.Copy(os.Stdout, resp.Body)
	return err
}
```

### Results

(It's a video, click it to view.)

[image](http://virtivia.com:27080/15kvcw9ib4f2u.mov)

### DIFF
--- a/compiler/natives/src/net/http/fetch.go
+++ b/compiler/natives/src/net/http/fetch.go
@@ -112,26 +112,32 @@ func (t *fetchTransport) RoundTrip(req *Request) (*Response, error) {
 				contentLength = cl
 			}
 
-			respCh <- &Response{
+			select {
+			case respCh <- &Response{
 				Status:        result.Get("status").String() + " " + StatusText(result.Get("status").Int()),
 				StatusCode:    result.Get("status").Int(),
 				Header:        header,
 				ContentLength: contentLength,
 				Body:          &streamReader{stream: result.Get("body").Call("getReader")},
 				Request:       req,
+			}:
+			case <-req.Cancel:
 			}
 		},
 		func(reason *js.Object) {
-			errCh <- errors.New("net/http: fetch() failed")
+			select {
+			case errCh <- errors.New("net/http: fetch() failed"):
+			case <-req.Cancel:
+			}
 		},
 	)
 	select {
+	case <-req.Cancel:
+		// TODO: Abort request if possible using Fetch API.
+		return nil, errors.New("net/http: request canceled")
 	case resp := <-respCh:
 		return resp, nil
 	case err := <-errCh:
 		return nil, err
-	case <-req.Cancel:
-		// TODO: Abort request if possible using Fetch API.
-		return nil, errors.New("net/http: request canceled")
 	}
 }
--- a/compiler/natives/src/net/http/fetch.go
+++ b/compiler/natives/src/net/http/fetch.go
@@ -1,0 +1,126 @@
+// +build js
+
+package http
+
+import (
+	"errors"
+	"io"
+	"strconv"
+
+	"github.com/gopherjs/gopherjs/js"
+)
+
+// streamReader implements a wrapper for ReadableStreamDefaultReader of https://streams.spec.whatwg.org/.
+type streamReader struct {
+	pending []byte
+	reader  *js.Object
+}
+
+func (r streamReader) Read(p []byte) (n int, err error) {
+	if len(r.pending) == 0 {
+		var (
+			bCh   = make(chan []byte)
+			errCh = make(chan error)
+		)
+		r.reader.Call("read").Call("then",
+			func(result *js.Object) {
+				if result.Get("done").Bool() {
+					errCh <- io.EOF
+					return
+				}
+				bCh <- result.Get("value").Interface().([]byte)
+			},
+			func(reason *js.Object) {
+				// Assumes it's a DOMException.
+				errCh <- errors.New(reason.Get("message").String())
+			},
+		)
+		select {
+		case b := <-bCh:
+			r.pending = b
+		case err := <-errCh:
+			return 0, err
+		}
+	}
+	n = copy(p, r.pending)
+	r.pending = r.pending[n:]
+	return n, nil
+}
+
+func (streamReader) Close() error {
+	// TODO: Implement. Use r.reader.cancel(reason) maybe?
+	return errors.New("not yet implemented")
+}
+
+// fetchTransport is a RoundTripper that is implemented using Fetch API. It supports streaming
+// response bodies.
+type fetchTransport struct{}
+
+func (t *fetchTransport) RoundTrip(req *Request) (*Response, error) {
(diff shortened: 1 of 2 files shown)


shurcooL:
Ok, I've setup an instance of `h2demo` with CORS enabled, so now it's possible to try out this change in the GopherJS Playground:

http://www.gopherjs.org/playground/#/5NfiqqGZCs

Wait for it to load, press run, and wait a few seconds as it compiles the program and runs it!

It will make a request to /reqinfo, then another to /clockstream and copy (streaming) first 1500 bytes before returning.

neelance:
Nice!

compiler/natives/net/http/fetch.go
```diff
@@ -0,0 +1,126 @@
...
+			//       property, maybe it's deprecated and we shouldn't use it? It does not seem to be deprecated
+			//       from a quick look at the docs, so maybe use it and hope the implementations are fixed soon?
+			//statusText := result.Get("statusText").String()
+			statusText := StatusText(result.Get("status").Int())
+
+			// TODO: Make this better.
+			header := Header{}
+			result.Get("headers").Call("forEach", func(value, key *js.Object) {
+				header[CanonicalHeaderKey(key.String())] = []string{value.String()} // TODO: Support multiple values.
+			})
```
shurcooL: Suggestions for how to do this better are very welcome.

What's a good way to iterate all keys (and values) of a JS `Iterable`?

In JS, one could do something like:

```JS
for (let value of ["a", "b", "c"]) {
    console.log(value);
}
```

But what's a good way in GopherJS?

shurcooL:
@neelance I've completed the critical functionality (closing response body, full header support, and sending request body). PTAL.

This should be functionally ready to be merged. There are still some style and minor TODOs to clean up. Also, before merging, this should be squashed into logical commits with clean commit message, and `gopherjs/compiler/natives/fs_vfsdata.go` needs to be regenerated.

compiler/natives/src/net/http/http.go
```diff
@@ -13,18 +13,29 @@ import (
 	"github.com/gopherjs/gopherjs/js"
 )
 
-var DefaultTransport RoundTripper = &XHRTransport{}
+var DefaultTransport = func() RoundTripper {
+	if fetchAPI, streamsAPI := js.Global.Get("fetch"), js.Global.Get("ReadableStream"); fetchAPI != js.Undefined && streamsAPI != js.Undefined {
```
neelance: Why the extra variables?

shurcooL: It checks that both `fetch` API and `ReadableStream` is available. The former is needed to make Fetch requests, and the latter is to be able to execute this line successfully:

```Go
Body: &streamReader{stream: result.Get("body").Call("getReader")},
```

For example, Firefox currently supports Fetch API but does not support streaming response bodies. So `result.Get("body")` would be nil, and using Fetch over XHR is not very advantageous.

compiler/natives/src/net/http/fetch.go
```diff
@@ -0,0 +1,143 @@
...
+	opt := map[string]interface{}{
+		"method":  req.Method,
+		"headers": headers,
+	}
+	if req.Body != nil {
+		// TODO: Find out if request body can be streamed into the fetch request rather than in advance here.
+		//       See BufferSource at https://fetch.spec.whatwg.org/#body-mixin.
+		body, err := ioutil.ReadAll(req.Body)
+		if err != nil {
+			req.Body.Close() // RoundTrip must always close the body, including on errors.
```
neelance: Can this be ensured via a `defer` at the beginning on the function?

compiler/natives/src/net/http/http.go
neelance: I meant this:
```
if js.Global.Get("fetch") != js.Undefined && js.Global.Get("ReadableStream") != js.Undefined {
```

compiler/natives/src/net/http/fetch.go
shurcooL: Yes, but I don't think it'd be cleaner. You still have to check that `req.Body` is not `nil` before doing it.

There are no other returns before this section of code (the code above simply deals with setting up the fields needed for the request), so `req.Body.Close()` will always be called.

compiler/natives/src/net/http/http.go
shurcooL: Ah, I overlooked that or didn't think of it. It looks simpler, so I'll do it.

I think I initially tried to pass the variable into the transport, e.g.:

```
if xhrAPI := js.Global.Get("XMLHttpRequest"); xhrAPI != js.Undefined {
	return &XHRTransport{xhrAPI}
}
```

But reverted because it was a bad idea. It meant that zero value was no longer valid, and there's no performance gain in storing `js.Global.Get("XMLHttpRequest")` in a variable.

shurcooL:
In case it's helpful, I've tested the implementation significantly during development using https://github.com/shurcooL/play/blob/master/192/go-js-fetch/http/test/main.go. You can run that suite by `go get -u -d github.com/shurcooL/play/192/go-js-fetch/http/test` and `gopherjs serve`ing it.

compiler/natives/src/net/http/fetch.go
neelance: K.

neelance:
LGTM

shurcooL:
Squashed and merged as 00254cbca59bacb672a3698e0490367cfc64b36c and db27c7c470d7404b6b201f82d6fd4821260bd13e (with logically separated changes, clean commit messages).

I did not force-push to the `fetch-transport` branch of this PR, so that we can still see the development history here. Instead, I just added "Closes #454." to the commit message of the squashed commit. But its content is identical as the diff in this PR.

shurcooL:
For anyone interested in an updated demo, I've updated the playground to use latest GopherJS compiler and created this simple demo:

http://www.gopherjs.org/playground/#/okdxAaGlfF

PR #467, opened 2016-05-24
compiler/natives: Update filter API.

shurcooL:
DO NOT MERGE until shurcooL/httpfs#4 is merged, since this change depends on that.

Rewrite filter in terms of `Keep` rather than `Skip` to make the logic simpler and more readable.

#### Test Plan

Tested locally with shurcooL/httpfs#4 merged (locally, not yet pushed upstream).


shurcooL:
Assigning you @neelance since you're the one that requested the filter API to be made more readable. Let me know if this resolves that issue.

neelance:
LGTM 👍

PR #469, opened 2016-05-31
use golang.org/x/tools/go/gcimporter15 for exporting types data

neelance:


shurcooL:
@neelance Can we delete the 9 month old `gcimporter` branch, or do you still need it?

neelance:
I'll take another look tomorrow.

dmitshur:
I'll delete it now. If it's needed, it can be restored.

PR #471, opened 2016-06-05
Remove tool command.

shurcooL:
It was initially added to support Go language tests. See https://github.com/gopherjs/gopherjs/commit/da075debc443510928caee184375aa76898c6d38#commitcomment-17247656.

Remove it since it's no longer needed, and it's unlikely to be used.

Helps #388.

In my local testing, `gopherjs test` still worked. [CI passes](https://circleci.com/gh/gopherjs/gopherjs/879) so that means standard library and language tests are working. If anyone knows of something that this breaks, please let me know.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -457,33 +457,6 @@ func main() {
 		}, options, nil))
 	}
 
-	cmdTool := &cobra.Command{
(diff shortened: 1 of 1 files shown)


neelance:
LGTM

PR #476, opened 2016-06-12
doc: Update net/http notes.

shurcooL:
Followup to #454. Mention that Fetch API can be used.

Mention that only `net/http` client is supported, server is not.

Also mention that by default, without third party polyfill modules for XHR or Fetch APIs, node.js will not support net/http.


dominikh:
LGTM

PR #477, opened 2016-06-13
WIP-388: Rewrite gopherjs tool.go to use cmd/go functionality.

shurcooL:
DO NOT MERGE. This is a PR for review only, it's not in a mergable state because not all components are present. I want to break review of the final PR into smaller pieces, to make it more diffs smaller and manageable.

This change rewrites tool.go to use the functionality of cmd/go with GopherJS compiler support added, which will be bundled here.

Changes:

- -watch flag is removed. Supporting it is possible, but difficult at this time and it's hard to test it. It may be brought back later, or maybe it'll be decided that it's out of scope, and another tool can help with watching. With gopherjs serve command available, the need for -watch flag is greatly reduced.

This is a mostly mechanical change, done by hand, so there's a chance for typos or other screw ups, forgetting flags, etc. This is what I want this review to catch.

Helps #388.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -1,32 +1,23 @@
+// GopherJS is a tool for compiling Go source code to JavaScript.
 package main
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
-	"go/ast"
 	"go/build"
-	"go/doc"
-	"go/parser"
 	"go/scanner"
-	"go/token"
 	"go/types"
 	"io"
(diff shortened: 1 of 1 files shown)


shurcooL:
Ok, this has WIP PR has served it's purpose (to share a part of the upcoming bigger change, and give an opportunity to spot typos/errors in this change). I'll close it now.

PR #479, opened 2016-06-25
Allow assignment to $

### LINKED ISSUE #374: js.Global.Set("$", ...) doesn't work
For some tests, I have a need to set the global `$` variable (I'm setting up some jQuery mocks), and it's not working, thanks to [this code](https://github.com/gopherjs/gopherjs/blob/master/compiler/expressions.go#L593):

					globalRef := func(id string) string {
						if recv.String() == "$global" && id[0] == '$' {
							return id
						}
						return recv.String() + "." + id
					}

I can work around this by calling `js.Global.Set("$global.$", ...)` instead, which is good enough for me, now that I found the solution.  But it makes me wonder if it could be improved.

At the very least, perhaps `$` could be a special-case exception to the special case, since we know GopherJS doesn't use `$` alone, and many other libraries do, so it is likely the most common use case.

Another option might be to allow some sort of escape sequence to indicate a literal $. Perhaps a variable name beginning with `$$` would be treated as `$`. Likely something more appropriate could be thought of.

At the very least, I'd suggest a documentation update (at least in the gopherjs/js package's description of the `Set()` and `Get()` methods).  I'd be happy to submit a PR with the documentation fixes, but I'll wait to see if a behavioral change is in order first.

flimzy:
This fixes #374 for the case of `$`.

### DIFF
--- a/compiler/expressions.go
+++ b/compiler/expressions.go
@@ -592,7 +592,7 @@ func (c *funcContext) translateExpr(expr ast.Expr) *expression {
 				declaredFuncRecv := sel.Obj().(*types.Func).Type().(*types.Signature).Recv().Type()
 				if typesutil.IsJsObject(declaredFuncRecv) {
 					globalRef := func(id string) string {
-						if recv.String() == "$global" && id[0] == '$' {
+						if recv.String() == "$global" && id[0] == '$' && len(id) > 1 {
 							return id
 						}
 						return recv.String() + "." + id


compiler/expressions.go
```diff
@@ -592,7 +592,7 @@ func (c *funcContext) translateExpr(expr ast.Expr) *expression {
 				declaredFuncRecv := sel.Obj().(*types.Func).Type().(*types.Signature).Recv().Type()
 				if typesutil.IsJsObject(declaredFuncRecv) {
 					globalRef := func(id string) string {
-						if recv.String() == "$global" && id[0] == '$' {
+						if recv.String() == "$global" && id[0] == '$' && len(id) > 1 {
```
shurcooL: I tried to read the code outside and found it hard to see what guarantees that length of `id` is always at least one (so that the `id[0]` access doesn't panic).

neelance: There simply are no identifiers with zero length. ;-)

shurcooL: I see, that makes sense.

Still, if it [were me](https://github.com/gopherjs/gopherjs/issues/374#issuecomment-170292266), I would swap the order of `len(id) > 1` and `id[0] == '$'` checks. The reason to do so is, code is the same, but it becomes easier for myself and others to read and be able to tell at a glance it won't panic.

The current code requires either being very familiar with the context already, or spending more time reading the surrounding code and trying to come to the conclusion that `id` will never be passed to `globalRef` as an empty string. I don't see any advantages to checking `id[0] == '$'` before `len(id) > 1`, are there any?

This is a very minor code review comment.

neelance: Is there a risk that `len(id) > 1 && id[0] == '$'` would look like a mistake which should rather be `len(id) >= 1 && id[0] == '$'`?

shurcooL: That's a good point, it may indeed look that way.

PR #480, opened 2016-06-26
compiler/natives/src/encoding/json: Add _test.go suffix to test file.

### LINKED ISSUE #481: Any package that imports "encoding/json" will needlessly import "testing" package and its deps.
The "testing" package imports a lot of stuff:

https://godoc.org/testing?import-graph

Currently, GopherJS has a bug where it will always pull in "testing" package and its dependencies whenever building a package that imports "encoding/json".

Fixing this will help reduce generated file size slightly.

shurcooL:
This way, the "testing" import isn't included when building package normally (not for tests).

Fixes #481.

---

Richard, there's one thing I wasn't sure about. Are there any special rules I need to be aware of when deciding on the name of native override files?

In this case, I went with the same name `stream_test.go` as the file being overridden (i.e., https://github.com/golang/go/blob/4fcb4eb27986e5fa7d26aeb85b04909e03d11ff9/src/encoding/json/stream_test.go#L389). Does that count as "additive"? Meaning the original is still included, and this new stream_test.go is additive on top? I.e., it only replaces the `TestHTTPDecoding` symbol. But keeps all others.

Based on output of `gopherjs test -v encoding/json` before and after the change, that seems to be the case... But I really wish this were documented so I wouldn't have to guess/verify/feel uncertain about it.


neelance:
LGTM. The file names do not matter except the usual `_test` suffix. The files for native overrides get added to the package. For all identifiers that exist in the original AND the overrides, the original identifier in the AST gets replaced by `_`.

shurcooL:
Thanks Richard. I've come to a similar conclusion after spending more time reading that code. Your overview confirms my understanding, and I love how succinct it is.

I'll add it to the documentation of that function so this is better documented for the future.

neelance:
Cool, thanks!

shurcooL:
> I'll add it to the documentation of that function so this is better documented for the future.

I've done that in 75fc501152ea51c15ab72b9e685e75e4152cda19.

PR #482, opened 2016-06-26
compiler/natives/src/time: Don't import strings package.

shurcooL:
The real time package doesn't import strings package. This keeps the
import graph of augmented time package consistent with the real
standard library.

Source: https://godoc.org/time?import-graph

This is a change analogous to #453.


neelance:
LGTM

neelance:
It's a really good improvement.

PR #483, opened 2016-06-26
build: Add test to verify that augmented packages don't import extras.

shurcooL:
Natives augment the standard library with GopherJS-specific changes. This test ensures that none of the standard library packages are modified in a way that adds imports which the original upstream standard library package does not already import. Doing that can increase generated output size or cause other unexpected issues (since the cmd/go tool does not know about these extra imports), so it's best to avoid it.

It checks all standard library packages. Each package is considered as a normal package, as a test package, and as an external test package.

Depends on #480 and #482 being merged. It catches both those issues, so the test will fail until those PRs are merged (and this branch is rebased).

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -135,10 +135,11 @@ func ImportDir(dir string, mode build.ImportMode) (*PackageData, error) {
 	return &PackageData{Package: pkg, JSFiles: jsFiles}, nil
 }
 
-// parse parses and returns all .go files of given pkg.
+// parseAndAugment parses and returns all .go files of given pkg.
 // Standard Go library packages are augmented with files in compiler/natives folder.
-// isTest is true when package is being built for running tests.
-func parse(pkg *build.Package, isTest bool, fileSet *token.FileSet) ([]*ast.File, error) {
+// If isTest is true and pkg.ImportPath has no _test suffix, package is built for running internal tests.
+// If isTest is true and pkg.ImportPath has _test suffix, package is built for running external tests.
+func parseAndAugment(pkg *build.Package, isTest bool, fileSet *token.FileSet) ([]*ast.File, error) {
 	var files []*ast.File
 	replacedDeclNames := make(map[string]bool)
 	funcName := func(d *ast.FuncDecl) string {
@@ -550,7 +551,7 @@ func (s *Session) BuildPackage(pkg *PackageData) (*compiler.Archive, error) {
 	}
 
 	fileSet := token.NewFileSet()
-	files, err := parse(pkg.Package, pkg.IsTest, fileSet)
+	files, err := parseAndAugment(pkg.Package, pkg.IsTest, fileSet)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


shurcooL:
Test is failing, as expected (see comment above):

```
--- FAIL: TestNativesDontImportExtraPackages (50.81s)
	build_test.go:85: augmented normal package "encoding/json" imports "testing" in file /src/encoding/json/json.go, but real "encoding/json" doesn't:
		realImports = ["github.com/gopherjs/gopherjs/nosync" "io" "runtime/internal/sys" "sync" "math" "reflect" "strings" "unicode/utf8" "encoding/base64" "bytes" "errors" "runtime" "sync/atomic" "runtime/internal/atomic" "internal/race" "syscall" "encoding" "strconv" "fmt" "time" "sort" "unicode/utf16" "unsafe" "unicode" "os"]
	build_test.go:85: augmented normal package "time" imports "strings" in file /src/time/time.go, but real "time" doesn't:
		realImports = ["sync/atomic" "syscall" "errors" "runtime" "runtime/internal/sys" "runtime/internal/atomic" "sync" "github.com/gopherjs/gopherjs/nosync" "unsafe" "internal/race"]
FAIL
FAIL	github.com/gopherjs/gopherjs/build	50.846s
```

neelance:
I like the idea of such test. However, the implementation is a bit hard to read for me. It might become easier if you make it a 2 step process:
1. gather dependency lists
2. check lists for violations
I also think that you don't have to fully parse the files. The `Imports`, `TestImports` and `XTestImports` fields of https://golang.org/pkg/go/build/#Package should already give you all the information you need.

shurcooL:
> I like the idea of such test.

Great!

> However, the implementation is a bit hard to read for me.

I agree.

Two comments. One, I have an idea for how to simplify the code, so I'll do that first. Second, I will try to make it more readable by splitting certain parts into helper functions and documenting them. Hopefully that'll help.

> I also think that you don't have to fully parse the files. The `Imports`, `TestImports` and `XTestImports` fields of https://golang.org/pkg/go/build/#Package should already give you all the information you need.

The general strategy (I'll add this as a comment to the test) is to first import each standard library package using the normal `build.Import`, which returns a `*build.Package`. That contains `Imports`, `TestImports`, and `XTestImports` values that are considered the "real imports".

That list of direct imports is then expanded to the transitive closure, meaning all packages that are indirectly imported are also added to the set.

Then, the result of GopherJS's `build.parse(*build.Package)` which returns `[]ast.File` with augmented parsed Go files of the package is compared, one file at at time, one import at a time. Each import is verified to belong in the set of allowed real imports.

> I also think that you don't have to fully parse the files.

So, as I described above, I only parse the files of the augmented package, because that's what `parse` outputs. I don't parse the files of the real Go package, I just use `*build.Package.{Imports,TestImports,XTestImports}` as you suggested.

shurcooL:
> One, I have an idea for how to simplify the code, so I'll do that first.

Nope, it's not going to work out. I can't make it simpler in the way that I was going to.

I think I'll have to rely on improved documentation. I'll push some commits, let me know if you think it gets to an acceptable level of readability.

I've also found a bug that caused external test files not to be checked correctly; I'll fix that.

shurcooL:
@neelance I've improved documentation, please see the 2 new commits. I hope it clears up what's going enough for this to be mergeable.

neelance:
Okay to merge. :-)

shurcooL:
Great, I've merged #480 and #482, and rebased this branch on top.

The tests should pass in CI now. I'll wait for that, and merge then.

Issue #440, opened 2016-04-15
Wrongly reusing cached precompiled code when changing build tags


simen:
I have a package that needs to run both on the native server and in a js client. I use build tags to omit code that rely on huge libraries like `reflect`, `time` and `fmt`. I had some very inconsistent results today where unused libraries would stick around even after I was completely sure I had excised them.

It turns out the problem was that gopherjs was reusing cached precompiled packages from `$GOPATH/pkg/darwin_js` and `$GOPATH/pkg/darwin_js_min`. When I deleted these directories and recompiled, everything worked as expected.

Easily worked around when you know it, but should probably be fixed at some point?

shurcooL:
Thanks for reporting this. I can confirm the issue.

shurcooL:
Hmm, this might be the same issue as #559, is that right? If not the same, definitely related.

shurcooL:
I've fixed #559 just now. Is it possible for you to check if this issue still occurs with latest GopherJS?

Issue #381, opened 2016-01-24
defer + go routine/channel + blocking = possible crash


flimzy:
When I uncommented the Example-executing code in #380, I ran into the following bug, which I have now condensed into a minimal reproduction case.  It may be that I'm running into GopherJS's limitation that the main Go routine can't block, but even in that case, some more informative and consistent error reporting would be nice.

Here's the code:

    package main

    import "fmt"

    func main() {
        foo()
    }

    func foo() {
        outC := make(chan string)
        go func() {
            outC <- ""
        }()

        defer func() {
            _ = <-outC
        }()

        fmt.Printf("Some output. Without this, GopherJS hangs.\n")
        return
    }

Go:

    Some output. Without this, GopherJS hangs

GopherJS:

    gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
    Some output. Without this, GopherJS hangs
    /home/jonhall/go/src/github.com/flimzy/extest/extest.go.328785025:1473
          throw err;
          ^

    TypeError: _r.$blk is not a function
        at Object.$packages.main.foo [as $blk] (/home/jonhall/go/src/github.com/flimzy/extest/extest.go.328785025:16211:137)
        at Object.$packages.main.main [as $blk] (/home/jonhall/go/src/github.com/flimzy/extest/extest.go.328785025:16186:66)
        at Object.$packages.main.$init [as $blk] (/home/jonhall/go/src/github.com/flimzy/extest/extest.go.328785025:16223:68)
        at fun (/home/jonhall/go/src/github.com/flimzy/extest/extest.go.328785025:1465:37)
        at $goroutine (/home/jonhall/go/src/github.com/flimzy/extest/extest.go.328785025:1463:19)
        at $runScheduled [as _onTimeout] (/home/jonhall/go/src/github.com/flimzy/extest/extest.go.328785025:1501:7)
        at Timer.listOnTimeout (timers.js:92:15)

Note in particular that in the deferred function, `err := recover()` happens before the blocking code `_ = <-outC`, which leads to the appearance that the error is not related to blocking code.  However, if I remove the blocking operation, by commenting out `_ = <-outC`, the code executes successfully in GopherJS as well.

Also, if I comment out the only `fmt.Printf` line (leaving `_ = <-outC` in place), the execution simply hangs, rather than generating an error.

flimzy:
I have made some progress on debugging this.  Still no solid solution yet.

[...]

PR #485, opened 2016-06-27
WIP: Use cmd/go with GopherJS support.

### LINKED ISSUE #382: Test multiple packages with ... wildcard
The "..." wildcard allows "go test" to test multiple sub packages, however this fails with "gopherjs test":

```
$ gopherjs test -v github.com/foo/bar/...
cannot find package "github.com/foo/bar/..." in any of:
	/usr/local/go/src/github.com/foo/bar/... (from $GOROOT)
	/Users/dave/dev/src/github.com/foo/bar/... (from $GOPATH)
```

It would be great to support this feature.

### LINKED ISSUE #388: proposal: Replace in-house implementation of `gopherjs` build tool with augmented forked `cmd/go` build tool.

### LINKED ISSUE #414: Circular imports lead to stack overflow crash
main.go:

    package main

    import "github.com/flimzy/jstest/foo"

foo/foo.go:

    package foo

    import "github.com/flimzy/jstest/bar"

bar/bar.go:

    package bar

    import "github.com/flimzy/jstest/foo"

`gopherjs build` crashes (after about 25 seconds):

    $ gopherjs build
    runtime: goroutine stack exceeds 1000000000-byte limit
    fatal error: stack overflow

    runtime stack:
    runtime.throw(0xaf0010, 0xe)
            /usr/local/go/src/runtime/panic.go:530 +0x90
    runtime.newstack()
            /usr/local/go/src/runtime/stack.go:940 +0xb11
    runtime.morestack()
            /usr/local/go/src/runtime/asm_amd64.s:359 +0x7f
    <snip>

### LINKED ISSUE #443: Not rebuilding main package
Hi,

When a main package imports another package the main package is not rebuilt as it should when using install (only the dependent package is being rebuilt). This is not how the "normal" go tool behaves.

Test for example:

    $ cat a/a.go 

package a

var F = 33

var D = map[string]string{
	"hej": "AAAAsadf",
}

    $ cat b/main.go 

package main

import (
	"a"

)

func main() {
	js.Global.Set("sdf", a.D)
}

    $ gopherjs -m -v install b

a
b

    $ ls -lah ../bin/ | grep b.js

-rw-rw-r--  1 vron vron  51K apr 24 19:04 b.js
-rw-rw-r--  1 vron vron  680 apr 24 19:04 b.js.map

    $ sed -i s/AAAA/BBBB/g a/a.go

    $ gopherjs -m -v install b

a      <-- Note that only package a is being rebuilt

    $ ls -lah ../bin/ | grep b.js

-rw-rw-r--  1 vron vron  51K apr 24 19:04 b.js
-rw-rw-r--  1 vron vron  680 apr 24 19:04 b.js.map

Note that the file (b.js) is not updated

### LINKED ISSUE #470: "Maximum call stack size exceeded" errors get swallowed
I don't know if there's anything GopherJS can do about this or not, but if so, it would have saved me several hours of debugging this afternoon.

In the case of infinite recursion, a GopherJS process will simply stop executing, without generating an error.

A simple reproduction case:

```
package main

import (
    "fmt"
)

func main() {
    recurse()
    fmt.Printf("This never gets printed\n")
}

func recurse() {
    recurse()
}
```

And when executed, there is no output, where I would expect a stack trace and an error message.

Another simple reproduction case (closer to how I found this bug in real life):

```
package main

import (
    "fmt"
    "testing"
)

func TestRecursion (m *testing.T) {
    recurse()
    fmt.Printf("You won't see this line\n")
}

func recurse() {
    recurse()
}
```

When I execute this one (`gopherjs test`), I get the following output:

```
FAIL    github.com/flimzy/jstest        0.299s
```

This gives me hope that GopherJS is somehow recovering from the error, and retaining control of execution long enough to output the FAIL, and thus that this issue may be resolvable.

When I ran the generated JS code in debugging mode, I was able to see the "Maximum call stack size" with a stack dump when the program crashed.

shurcooL:
DO NOT MERGE.

This is a very early unfinished WIP PR to use cmd/go with GopherJS support instead of own build tool (#388).

Don't try it on your computer yet! It won't work, because there are hardcoded paths specific to my setup. But that will be fixed soon, up next. I will update this when it's ready for first other people to try it.

Large parts of the code here are generated from https://github.com/gopherjs/go repository, [go1.6.2-v0.1](https://github.com/gopherjs/go/tree/go1.6.2-v0.1) or whatever's latest branch. That repo is used for development purposes as a cmd/go base with cherry-picked GopherJS changes on top.

Resolves #388.

Fixes #470.
Fixes #443.
Fixes #414.
Fixes #382.
## TODO
- [ ] Get rid of hardcoded paths in code, resolve remaining critical TODOs in code.
- [ ] Figure out a way to bootstrap (install) the `github.com/gopherjs/gopherjs/js` and `github.com/gopherjs/gopherjs/nosync` packages.
- [ ] `GOARCH=js go get -compiler=gopherjs github.com/goxjs/websocket` should work.
  
  The issue was missing "netgo" build tag. Need to fix that.
- [ ] Verify all "cmd/go would not have this issue" issues.
- [ ] Implement a linker flag to control minify/not minify. Currently it always minifies by default.
- [ ] Figure out a way to deal with sourcemap files. Right now they're not generated at all. Generating a 2nd file is kinda difficult and very different from all other Go compilers... maybe it should be embedded in to the same file? As a debug flag/option? Figure this out.
- [ ] Implement and test .inc.js stuff.
- [ ] Take care of 2 items in Setup section.
- [ ] Upstream or fork required fixes/changes to bundle command:
  - [x] Part 1 - https://golang.org/cl/23785
  - [ ] Part 2 - not yet done
- [ ] Document the complicated source package locations (and generated package locations). Explain why it's needed.
- [x] Fix `/usr/local/go/test/fixedbugs/issue10332.go` which is currently failing because `reflect.ValueOf(foo{}).Type().Field(0).PkgPath` is being "command-line-arguments" instead of "main" for some reason.
- [x] `gopherjs test fmt` should work, need to fix support for external tests.
- [x] There may be an issue with _test.go files being built for normal packages.
  
  ```
  gopherjs build -m -o /tmp/tictactoe.js github.com/shurcooL/tictactoe/cmd/tictactoe
  go build encoding/json: importFromDisk: archive for path "testing" missing: open /Users/Dmitri/Dropbox/Work/2013/GoLand/pkg/gopherjs_darwin_js/testing.a: no such file or directory
  ```
  
  This turned out to be #481.
### cmd/go would not have this issue

https://github.com/gopherjs/gopherjs/issues?q=is%3Aissue+is%3Aopen+label%3A%22cmd%2Fgo+would+not+have+this+issue%22
- [x] https://github.com/gopherjs/gopherjs/issues/470
- [x] https://github.com/gopherjs/gopherjs/issues/443
- [ ] https://github.com/gopherjs/gopherjs/issues/440
  -   not yet verified
- [x] https://github.com/gopherjs/gopherjs/issues/414
- [x] https://github.com/gopherjs/gopherjs/issues/382
- [ ] https://github.com/gopherjs/gopherjs/issues/239
  -   [x] blocked on https://github.com/gopherjs/gopherjs/issues/381
## Setup
-   Need to create a GOROOT/pkg subdirectory that is writeable, if GOROOT/pkg isn't.
  
  ```
  sudo mkdir /usr/local/go/pkg/gopherjs_darwin_js
  sudo chown $USER /usr/local/go/pkg/gopherjs_darwin_js
  ```
  
  Where `darwin` is your GOOS.
  
  Otherwise you get this when installing first package (which tries to install stdlib with GOARCH=js for first time, starting with runtime/internal/sys package):
  
  ```
  go install runtime/internal/sys: mkdir /usr/local/go/pkg/gopherjs_darwin_js: permission denied
  ```
-   Benchmarks seem to require System Calls (https://github.com/gopherjs/gopherjs/blob/master/doc/syscalls.md) to be installed, or otherwise you get:
  
  ```
  BenchmarkBasic  2000000000           0.82 ns/op
  fatal error: all goroutines are asleep - deadlock!
  exit status 2
  ```


shurcooL:
With https://github.com/gopherjs/gopherjs/pull/485/commits/ecc1b7a5ec0aeaf4e8d66b12d081f292eb8c6584 done, many tests are now passing, and many are failing.

Most of the failures are due to Examples now running, but they're still not supported due to #381. For CI to pass, either need to disable examples from running, or fix the issue causing them to be unsupported.

build/build_test.go
```diff
@@ -2,12 +2,12 @@ package build
...
 	"fmt"
-	gobuild "go/build"
 	"go/token"
 	"strconv"
 	"strings"
 	"testing"
 
+	gobuild "github.com/gopherjs/gopherjs/go/build"
 	"github.com/kisielk/gotool"
 	"github.com/shurcooL/go/importgraphutil"
```
shurcooL: This doesn't fix the issue completely because `github.com/shurcooL/go/importgraphutil` package still uses original `go/build`, so it's incompatible.

shurcooL: Fixed completely in a49c2ae6edfbdebd6cdec9aacffe7f06778f5592.

shurcooL:
https://golang.org/cl/23785 has been reviewed and merged, which brings us ever so closely forward, since this PR depended on that change. (See "Upstream or fork required fixes/changes to bundle command." item.) I still have to make and send out another improvement to `bundle` that this PR depends on.

shurcooL:
I'm in the process of updating/rebasing on top of `cmd/go` of Go 1.7.

Just finished updating the github.com/gopherjs/go repository (which is the source - everything in this repository is generated/copied from that one). There were quite a few minor changes in `cmd/go` and `go/build` packages between 1.6 and 1.7, so it took a few hours of work, but it's still logically straightforward.

I've pushed the result to https://github.com/gopherjs/go/commits/go1.7.0-v0.1. See individual commit messages and their changes - each commit should be logically separate and very easy to understand (that's the goal).

All of its tests are passing and things are building successfully (with noted limitations):

```
$ go test github.com/gopherjs/go/...
ok  	github.com/gopherjs/go/cmd/go	113.478s
ok  	github.com/gopherjs/go/go/build	0.072s
?   	github.com/gopherjs/go/internal/race	[no test files]
ok  	github.com/gopherjs/go/internal/singleflight	0.023s
?   	github.com/gopherjs/go/internal/testenv	[no test files]
```

I'm going to keep reusing this PR, and just squash, rebase, regenerate and force push on top. I'll keep a backup of the history of changes/work on the Go 1.6 branch for reference, in case we need to look something up later.

So, this PR will soon become 1 squashed commit that's rebased on latest GopherJS 1.7-1 and github.com/gopherjs/go repository on [branch go1.7.0-v0.1](https://github.com/gopherjs/go/commits/go1.7.0-v0.1).

neelance:
@shurcooL Do you have any plan on how we want to continue on this?

shurcooL:
Yep, I've been meaning to post an update here.

I will be closing this PR and abandoning this effort. I may choose to revisit the effort later, but until further notice, I won't be actively working on this.

There are 2 reasons I'll close the PR:

1. As of early Go 1.9 development, there are many changes being done to re-organize `cmd/go`. It's being split up from one huge package, into many packages. See https://github.com/golang/go/issues?q=label%3AGoCommand.

    GopherJS is now version 1.8-1 and supports Go 1.8. This PR was based on `cmd/go` from Go 1.7, and so it's out of date. It would need to be updated for Go 1.8. However, given the changes going into 1.9, it might even make sense to aim at start using that as base.

    So, this PR is out of date and not helpful.

2. The goal of this PR was to get it into a good "mergeable" shape where we could feel comfortable about merging it into `master`, confident that it would work as well as the current approach, but better (fix more issues), and lead to decreased maintenance effort in the future (as outlined in #388). This was unsuccessful.

Here's a little postmortem and explanations why this effort did not work out.

When I first prototyped this idea, I had early success. I was able to make a relatively easy-to-maintain go-gettable fork of `cmd/go`. I was also able to add support for a new compiler toolchain and get around all the difficulties of import paths, `internal` packages, and not being able to modify GOROOT for our needs.

At the time, I thought that was the hardest part, and once we'd gotten past that, we would be able to take advantage of the complicated logic that cmd/go has internally to resolve all the package build order, the logic for figuring out where to place built libraries, reusing non-stale built libraries, etc., etc. I also thought it'd be easier to simplify `gopherjs` itself to be just a dedicated compiler and linker, rather than everything-in-one.

However, it turned out that cmd/go internal code was actually very complicated, messy, and contained many hacks/weird fixes that are specific to `gc` and `gccgo` compilers (e.g., see https://github.com/golang/go/commit/b92a0a8969218c29a3300cc5ec1ba142343375c3), and various edge cases that needed to be handled. I had to both understand and re-implement those things for `gopherjsToolchain`. Its internal interfaces were too low level and not well documented, because they're not a part of any public API. It was also very hard to fully understand, especially when I had little spare time to dedicate to it after day job. I was hoping/wanting a clean interface to that would abstract all the messy logic, but that didn't exist (it may begin to exist with the current re-architecture of cmd/go, but I can't be sure without confirming that first; the re-architecture may have different goals and solve other problems). Many things would be easier if we could start by working on cmd/go to make it friendlier to expand towards other compilers first, but I don't know if I'd get permission to do that.

So, I gave up on it because the effort and difficulty and time needed were much higher than originally anticipated, and the benefits are not worth it.

There may be another opportunity to consider this in the future, given the cmd/go split into packages, and, for example, there were changes to make it possible to `go get` different versions of Go. But, that would need to evaluated. I'm happy to provide more detailed comments and participate in any future discussions on this topic, because of my experience.

shurcooL:
@neelance I will also delete the https://github.com/gopherjs/go WIP repository. It was created to support this PR, and is no longer needed.

neelance:
All right. It was worth a try. Thanks for investigating.

PR #486, opened 2016-07-04
compiler/natives/src/net/http: Set Fetch credentials to same-origin.

shurcooL:
This makes the behavior of http.DefaultTransport using Fetch API more similar as when using XHR API.

It's unfortunate to have to do this, since it's kinda magic, but this seems to be the most reasonable default. It's consistent with the previous behavior of XHR implementation.

There is no equivalent property in http.Request as far as I can tell. The only other way to set credentials would be to use req.AddCookie, but then one would need to get the cookie in JavaScript, which is not possible if that cookie has HttpOnly flag set.

It appears this is an unfortunate reality that we have to deal with to provide a useful http.DefaultTransport implementation on frontend, because of restrictions put into place due to security concerns on the web.

For reference, see:

https://fetch.spec.whatwg.org/#concept-request-credentials-mode
https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials

> A request has an associated credentials mode, which is "omit", "same-origin", or "include". Unless stated otherwise, it is "omit".

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials

> Setting withCredentials has no effect on same-site requests.

So default XHR behavior without withCredentials set is like Fetch with "same-origin" credentials mode.

Updates #454.

/cc @dominikh


neelance:
LGTM

PR #490, opened 2016-07-16
use new cobra help API.

### LINKED ISSUE #489: docker build failing
try running the following as a Dockerfile, with "docker run ."

````
FROM golang
RUN go get -u github.com/gopherjs/gopherjs
````

Output is:
````
Sending build context to Docker daemon  5.12 kB
Step 1 : FROM golang
 ---> 8ecba0e9bd48
Step 2 : RUN go get -u github.com/gopherjs/gopherjs
 ---> Running in 295a1bfe99eb
# github.com/gopherjs/gopherjs
src/github.com/gopherjs/gopherjs/tool.go:483: cmdServe.Help undefined (type *cobra.Command has no field or method Help)
The command '/bin/sh -c go get -u github.com/gopherjs/gopherjs' returned a non-zero code: 2


````

### PR DESCRIPTION
**Pull request by flimzy**
See https://github.com/spf13/cobra/pull/305

Fixes #489

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -480,7 +480,7 @@ func main() {
 		var root string
 
 		if len(args) > 1 {
-			cmdServe.Help()
+			cmdServe.HelpFunc()
 			return
 		}
 

### CODE REVIEW

File: tool.go [hunk 1]
```diff
@@ -480,7 +480,7 @@ func main() {
 		var root string
 
 		if len(args) > 1 {
-			cmdServe.Help()
+			cmdServe.HelpFunc()
```
shurcooL: This will fix the build, but I don't think it's right.

Look at the definition of `HelpFunc` method:

https://godoc.org/github.com/spf13/cobra#Command.HelpFunc

It returns a function with default help behavior `func(*Command, []string)`, which should be called.

See some usage examples of it being called here:

https://sourcegraph.com/github.com/spf13/cobra/-/info/GoPackage/github.com/spf13/cobra/-/Command/HelpFunc

flimzy: You're probably right.  I did some simple testing, but I in light of that, I don't think I was thorough enough. I'll close this PR, now that you're working on it as well.

PR #495, opened 2016-08-16
Add support for Go 1.7 (WIP).

### LINKED ISSUE #475: Go 1.7 support
Go1.7 beta is out now and would be awesome to support it (I think a lot of people will be upgrading quickly due to the performance and compiling improvements)

Not a huge hurry yet, just thought I would put in the ticket early :)

I started poking around and I think that changing runtime.go will fix half the issues
```
-const GOOS = sys.TheGoos
+const GOOS = sys.GOOS
```
And the other issue I don't fully understand (when trying to do a gopherjs build):
```
$GOROOT/src/runtime/error.go:70:9: invalid operation: e._type (variable of type *struct{_string *string}) has no field or method string
```

shurcooL:
This is the PR to add support for Go 1.7, which [has been released](https://blog.golang.org/go1.7) recently.

It's still WIP. Most tests are passing thanks to @neelance's work [so far](https://github.com/gopherjs/gopherjs/issues/475#issuecomment-234841873), but there are a few that are not.

Known issues:

- [ ] `archive/tar` `TestReader ` test is panicking because of #496.

Resolves #475 when merged.


shurcooL:
Great, as of f8e7e291cc5861e7f31506390dc2ef354c635221, the `go test -v -race ./...` step is passing (but of course we'll have more real issues to actually resolve after this is merged). So just the other one left. 😛

benechols-wf:
Quick check with go1.7 and it appears to be working for me!

PR #508, opened 2016-09-04
Add plumbing to support build tags in test mode.

### LINKED ISSUE #507: 'gopherjs test' mode ignores tags
It appears that tags are ignored by the 'test' mode.

Consider this `foo` package with two implementations:

`foo.go`:

    // +build foo

    package foo

    func Foo() string {
    	return "foo"
    }

`nonfoo.go`:

    // +build !foo

    package foo

    func Foo() string {
    	return "bar"
    }

And `foo_test.go`:

    package foo
    
    import (
    	"testing"
    )
    
    func TestFoo(t *testing.T) {
	if x := Foo(); x != "foo" {
    		t.Fatalf("Expected foo, got %s", x)
    	}
    }

Running gopherjs test, as normal, will fail as expected:

    $ gopherjs test
    --- FAIL: TestFoo (0.06s)
            jstest_test.go:9: Expected foo, got bar                                                                                                              
    FAIL
    FAIL    github.com/flimzy/jstest        0.429s

But specifying the 'foo' tag also fails:

    $ gopherjs test --tags=foo
    --- FAIL: TestFoo (0.06s)
            jstest_test.go:9: Expected foo, got bar                                                                                                              
    FAIL
    FAIL    github.com/flimzy/jstest        0.422s

In looking at `tool.go`, I see that `cmdTest` is never told about `flagTags`, which seems like a confirmation that this functionality is not implemented.  I haven't dug much deeper.

flimzy:
This resolves #507.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -121,8 +121,8 @@ func importWithSrcDir(path string, srcDir string, mode build.ImportMode, install
 
 // ImportDir is like Import but processes the Go package found in the named
 // directory.
-func ImportDir(dir string, mode build.ImportMode) (*PackageData, error) {
-	pkg, err := build.ImportDir(dir, mode)
+func ImportDir(dir string, mode build.ImportMode, installSuffix string, buildTags []string) (*PackageData, error) {
+	pkg, err := NewBuildContext(installSuffix, buildTags).ImportDir(dir, mode)
 	if err != nil {
 		return nil, err
 	}
--- a/tool.go
+++ b/tool.go
@@ -287,13 +287,15 @@ func main() {
 	outputFilename := cmdTest.Flags().StringP("output", "o", "", "Compile the test binary to the named file. The test still runs (unless -c is specified).")
 	cmdTest.Flags().AddFlag(flagMinify)
 	cmdTest.Flags().AddFlag(flagColor)
+	cmdTest.Flags().AddFlag(flagTags)
 	cmdTest.Run = func(cmd *cobra.Command, args []string) {
+		options.BuildTags = strings.Fields(*tags)
 		os.Exit(handleError(func() error {
 			pkgs := make([]*gbuild.PackageData, len(args))
 			for i, pkgPath := range args {
 				pkgPath = filepath.ToSlash(pkgPath)
 				var err error
-				pkgs[i], err = gbuild.Import(pkgPath, 0, "", nil)
+				pkgs[i], err = gbuild.Import(pkgPath, 0, "", options.BuildTags)
 				if err != nil {
 					return err
 				}
@@ -310,7 +312,7 @@ func main() {
 					if err != nil {
 						return err
 					}
-					if pkg, err = gbuild.Import(pkgPath, 0, "", nil); err != nil {
+					if pkg, err = gbuild.Import(pkgPath, 0, "", options.BuildTags); err != nil {
 						return err
 					}
 				}
--- a/tool.go
+++ b/tool.go
@@ -317,7 +317,7 @@ func main() {
 					}
 				}
 				if pkg == nil {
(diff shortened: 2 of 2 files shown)


tool.go
```diff
@@ -310,7 +312,7 @@ func main() {
 					if err != nil {
 						return err
 					}
-					if pkg, err = gbuild.Import(pkgPath, 0, "", nil); err != nil {
+					if pkg, err = gbuild.Import(pkgPath, 0, "", options.BuildTags); err != nil {
 						return err
 					}
 				}
```
shurcooL: The very next line is an `if pkg == nil { ... gbuild.ImportDir(currentDirectory, 0) ... }` block.

If that path is taken, then build tags would be ignored, right?

To fix that, you'd have to modify https://github.com/gopherjs/gopherjs/blob/45518c130e5bd1525f20110830a4986365a153de/build/build.go#L124 somewhat significantly, it looks like. It's fine by me to leave that out of this PR.

shurcooL:
It'd also be really nice to find a way to add tests for this functionality, so that we can be sure it doesn't break. However, it's unlikely to regress since this code is unlikely to be refactored, up until #485 supersedes and replaces it completely. I'll be sure to test this stuff there, but tests would still be nice and make it easier (assuming they work well).

The 4 lines changed by this PR so far fixes a clear oversight and resolves the majority of the problem. The two comments above are optional, and can be split off into future PRs. That way we can merge this part sooner.

So, LGTM. @neelance, does that SGTU?

@flimzy, let me know if you'd like to work on fixing `gbuild.ImportDir` and adding tests.

Thanks a lot for reporting the issue, looking into it and making a PR! :)

tool.go
flimzy: Good catch, @shurcooL. It's actually not a very complex fix for that case.

flimzy:
I'm always a fan of tests.  I'm not sure of the best way to go about that in this case.  If that's considered important, I'll spend some time on it.  But as you say, this code is likely to go away soon anyway, so that does nudge me to the side of laziness. :)

build/build.go
```diff
@@ -121,8 +121,8 @@ func importWithSrcDir(path string, srcDir string, mode build.ImportMode, install
 
 // ImportDir is like Import but processes the Go package found in the named
 // directory.
-func ImportDir(dir string, mode build.ImportMode) (*PackageData, error) {
-	pkg, err := build.ImportDir(dir, mode)
+func ImportDir(dir string, mode build.ImportMode, installSuffix string, buildTags []string) (*PackageData, error) {
+	pkg, err := NewBuildContext(installSuffix, buildTags).ImportDir(dir, mode)
 	if err != nil {
 		return nil, err
 	}
```
shurcooL: Okay, so this ended up being not as big of a change as I originally thought. I thought you'd have to copy all of [this logic](https://github.com/gopherjs/gopherjs/blob/45518c130e5bd1525f20110830a4986365a153de/build/build.go#L70-L112) too. But I see now that you don't. That logic was not a part of `ImportDir` before, but it's only needed for building the standard Go library with GopherJS. That means `build.ImportDir` can't be used to build stdlib packages. Oh well.

shurcooL:
LGTM. @neelance, do you want to review this?

build/build.go
flimzy: I think it would need to be added also to [BuildDir](https://github.com/gopherjs/gopherjs/blob/45518c130e5bd1525f20110830a4986365a153de/build/build.go#L402) to cover all cases.

That functionality might be worth adding, but it seems to me like it probably deserves its own issue/PR. What do you think?

shurcooL: Yes, that's definitely out of scope of this PR.

neelance:
LGTM

Issue #511, opened 2016-09-06
FreeBSD: gopherjs build "undeclared name: sysAF_LINK"


steveoc64:
Q:  Might this be something that #485 will address ?

Been running through available combinations on DigitalOcean, to ensure coverage for gopherjs across the different platforms that are available.

NOTE:   I have another droplet on FreeBSD 10.3 with the same application codebase, using go1.6, and that all works fine.

Platform:  FreeBSD 10.3 

freshly created droplet, with freshly installed go 1.7

```
digitalocean [freebsd@testbox4 ]$ go env
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="freebsd"
GOOS="freebsd"
GOPATH="/home/freebsd/go"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/freebsd_amd64"
CC="clang"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -gno-record-gcc-switches"
CXX="clang++"
CGO_ENABLED="1"
```

gives : 

```
digitalocean [freebsd@testbox4 ]$ make
cd app && gosimple
gopherjs build app/*.go -o dist/public/app.js
../../../../../../../local/go/src/vendor/golang_org/x/net/route/address.go:25:42: undeclared name: sysAF_LINK
../../../../../../../local/go/src/vendor/golang_org/x/net/route/address.go:31:35: undeclared name: sysAF_LINK
../../../../../../../local/go/src/vendor/golang_org/x/net/route/address.go:91:43: undeclared name: sysAF_INET
../../../../../../../local/go/src/vendor/golang_org/x/net/route/address.go:100:43: undeclared name: sysAF_INET6
../../../../../../../local/go/src/vendor/golang_org/x/net/route/address.go:105:7: undeclared name: sysAF_INET
../../../../../../../local/go/src/vendor/golang_org/x/net/route/address.go:112:7: undeclared name: sysAF_INET6
../../../../../../../local/go/src/vendor/golang_org/x/net/route/address.go:181:13: undeclared name: sysAF_INET6
../../../../../../../local/go/src/vendor/golang_org/x/net/route/address.go:223:10: undeclared name: sysRTAX_MAX
[...]

PR #513, opened 2016-09-08
add networking constant declarations for freebsd

steveoc64:
Simple addition of BSD networking constants, which is needed for gopherjs build on FreeBSD. 

As per DS. recommendation in #511

### DIFF
--- a/compiler/natives/src/net/route/zsys_freebsd.go
+++ b/compiler/natives/src/net/route/zsys_freebsd.go
@@ -1,0 +1,119 @@
+// Created by cgo -godefs - DO NOT EDIT
+// cgo -godefs defs_freebsd.go
+
+// +build js
+
+package route
+
+const (
+	sysAF_UNSPEC = 0x0
+	sysAF_INET   = 0x2
+	sysAF_ROUTE  = 0x11
+	sysAF_LINK   = 0x12
+	sysAF_INET6  = 0x1c
+
+	sysNET_RT_DUMP     = 0x1
+	sysNET_RT_FLAGS    = 0x2
+	sysNET_RT_IFLIST   = 0x3
+	sysNET_RT_IFMALIST = 0x4
+	sysNET_RT_IFLISTL  = 0x5
+)
+
+const (
+	sysCTL_MAXNAME = 0x18
+
+	sysCTL_UNSPEC   = 0x0
+	sysCTL_KERN     = 0x1
+	sysCTL_VM       = 0x2
+	sysCTL_VFS      = 0x3
+	sysCTL_NET      = 0x4
+	sysCTL_DEBUG    = 0x5
+	sysCTL_HW       = 0x6
+	sysCTL_MACHDEP  = 0x7
+	sysCTL_USER     = 0x8
+	sysCTL_P1003_1B = 0x9
+)
+
+const (
+	sysRTM_VERSION = 0x5
+
+	sysRTM_ADD        = 0x1
+	sysRTM_DELETE     = 0x2
+	sysRTM_CHANGE     = 0x3
+	sysRTM_GET        = 0x4
+	sysRTM_LOSING     = 0x5
+	sysRTM_REDIRECT   = 0x6
+	sysRTM_MISS       = 0x7
+	sysRTM_LOCK       = 0x8
+	sysRTM_RESOLVE    = 0xb
+	sysRTM_NEWADDR    = 0xc
+	sysRTM_DELADDR    = 0xd
+	sysRTM_IFINFO     = 0xe
+	sysRTM_NEWMADDR   = 0xf
+	sysRTM_DELMADDR   = 0x10
+	sysRTM_IFANNOUNCE = 0x11
+	sysRTM_IEEE80211  = 0x12
+
+	sysRTA_DST     = 0x1
+	sysRTA_GATEWAY = 0x2
+	sysRTA_NETMASK = 0x4
+	sysRTA_GENMASK = 0x8
+	sysRTA_IFP     = 0x10
+	sysRTA_IFA     = 0x20
+	sysRTA_AUTHOR  = 0x40
+	sysRTA_BRD     = 0x80
+
+	sysRTAX_DST     = 0x0
+	sysRTAX_GATEWAY = 0x1
+	sysRTAX_NETMASK = 0x2
+	sysRTAX_GENMASK = 0x3
+	sysRTAX_IFP     = 0x4
+	sysRTAX_IFA     = 0x5
+	sysRTAX_AUTHOR  = 0x6
+	sysRTAX_BRD     = 0x7
+	sysRTAX_MAX     = 0x8
+)
+
+const (
+	sizeofIfMsghdrlFreeBSD10        = 0xb0
+	sizeofIfaMsghdrFreeBSD10        = 0x14
+	sizeofIfaMsghdrlFreeBSD10       = 0xb0
+	sizeofIfmaMsghdrFreeBSD10       = 0x10
+	sizeofIfAnnouncemsghdrFreeBSD10 = 0x18
+
+	sizeofRtMsghdrFreeBSD10  = 0x98
+	sizeofRtMetricsFreeBSD10 = 0x70
+
+	sizeofIfMsghdrFreeBSD7  = 0xa8
+	sizeofIfMsghdrFreeBSD8  = 0xa8
+	sizeofIfMsghdrFreeBSD9  = 0xa8
+	sizeofIfMsghdrFreeBSD10 = 0xa8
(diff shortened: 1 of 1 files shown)


compiler/natives/src/net/route/zsys_freebsd.go
```diff
@@ -0,0 +1,119 @@
+// Created by cgo -godefs - DO NOT EDIT
+// cgo -godefs defs_freebsd.go
+
+// +build js
```
shurcooL: This line would make sense inside GOROOT, but it's not needed in this context, since this is a file inside:

```
gopherjs/compiler/natives/src/...
```

So it's clear this is meant for GopherJS.

compiler/natives/src/net/route/zsys_freebsd.go
```diff
@@ -0,0 +1,119 @@
+// Created by cgo -godefs - DO NOT EDIT
+// cgo -godefs defs_freebsd.go
```
shurcooL: Let's get rid of this, since it's potentially confusing, and maybe just write a comment like:

```
// This is a copy of https://raw.githubusercontent.com/golang/go/go1.7.1/src/vendor/golang_org/x/net/route/zsys_freebsd_amd64.go.
```

This is so that it'll be easier to remember/figure out where this file came from in the future.

shurcooL:
Looks good, minor suggestions.

Once this has LGTM and is ready to merge, we'll need to regenerate this package. That can be done by running `go generate github.com/gopherjs/gopherjs/compiler/natives`, but you'll need the latest version of [`vfsgendev`](https://github.com/shurcooL/vfsgen#vfsgendev-usage) installed.

You can do it after addressing review comments, or not. We can do it when merging, it's really easy for us.

Thanks!

shurcooL:
The tests are passing, so that's good. But I'd still like to verify something before this goes in. The filename is `zsys_freebsd.go`, which under normal Go build tool acts like a `freebsd` build constraint.

However, the rules for how `compiler/natives` files are used are kinda custom, and I don't know if they have this behavior. See:

https://github.com/gopherjs/gopherjs/blob/05b9d708f1c9b4cbe71da718f3ce23665031021a/build/build.go#L143-L148

Basically, we want to make sure this file is included on `GOOS=freebsd` only, not in other cases.

shurcooL:
Actually, that comment suggests that:

> The file names do not matter except the usual `_test` suffix.

So I think that suggests the `_freebsd` is actually ignored.

This whole process made me realize that another way (not necessarily better, but worth considering) of resolving #511 and similar issues would be if GopherJS chose to use a constant value for `GOOS` rather than adopting the host OS value. We could fix it to the most popular/standard one, such as `GOOS=linux GOARCH=js`. That would mean building with GopherJS for `GOARCH=js` would be consistent on any platform. I doubt GOOS variations are actually meaningful or helpful. /cc @neelance However, this is a much bigger change.

compiler/natives/src/net/route/zsys_freebsd.go
steveoc64: Applied those comment clarifications.

steveoc64:
Was thinking something similar earlier .. since this sort of error could manifest in other strange ways that we havent seen yet, the ideal solution would be for the compiler to recognise the situation, and act accordingly.

I like the idea of applying a constant value of GOOS based on the host.  Although in this particular case it may not do the right thing, since the FreeBSD filenames have that extra architecture extension on files and no build constraint helpers inside the file  ... a deadly combination which would be hard for the compiler to spot.

No rush - I might take a bit of time to read the code a bit deeper, digest things, and have a good think about it as well. Good fun.

steveoc64:
One other thing to consider, that may be quite important too : 

It would be desirable that for any given set of front end GO code, it should compile to an identical set of JS code, regardless of the platform that the compiler is running on.

... might be worth thinking about the implications of that for a while.

shurcooL:
> It would be desirable that for any given set of front end GO code, it should compile to an identical set of JS code, regardless of the platform that the compiler is running on.

Yep, being able to attain that would be one of the benefits of a fixed GOOS value.

It's not currently the case even when compiling on the same platform, so that needs to be fixed first. I'm not sure if there's already an issue tracking it or not. If not, we should make one. Looks like there [aren't any open ones](https://github.com/gopherjs/gopherjs/issues?utf8=%E2%9C%93&q=is%3Aissue%20deterministic%20is%3Aclosed%20), but it is an issue, so I'll report that.

neelance:
You should be able to set the environment variable `GOOS=linux`. We have support for system calls on Linux and OS X, so I think using a fixed `GOOS` is not ideal. Do system calls work on BSD? We could also consider defaulting to `linux` on unsupported platforms, including Windows.

neelance:
@steveoc64 Does the `GOOS=linux` approach work for you?

steveoc64:
Apologies for the delay ... been moving offices, and I completely missed seeing this thread for the last week.  In the meantime, the existing patch does the job, but it does add a bit of complexity, and leaves the question open about other similar cases that may come up in future.

I think that the GOOS=linux approach is a really elegant solution that should clean up a range of uncertainties.

I can try this again today on my FreeBSD instance, and let you know how that goes. Be good to close this bug report off.

steveoc64:
@neelance @shurcooL 

Summary:  GOOS=linux works for FreeBSD, we can drop the PR and close the initial bug report. No code changes needed.   Some other interesting things come up from this though, so please read on.

Detail:
I have set up a new droplet with FreeBSD 10.3  amd64
Install go1.7.1
go get -u everything
compile the app front end with : 

GOOS=linux gopherj build app/*.go -o dist/public/app.js -m
(compiles with no errors)

.. and then run the app. All works fine.

Also, repeated the same steps on a Linux server with the same amd64 arch.  Installed fresh go1.7.1, and repeat the same steps that I used on the FreeBSD machine.

The generated binaries from gopherjs come out to be different, despite both using GOOS=linux : 
FreeBSD   -> 5456199 bytes
Linux       -> 5468604 bytes

Of course, both generated app.js files work fine in the browser, regardless of which server is being used to serve them up.  I have no problem serving the Linux-generated JS from FreeBSD, and vise versa, but I understand that is a meaningless comparison, as its the browser that must execute the code.

Investigating that difference further, the Linux generated app.js file does contain a slightly different set of functions in the compiled $pkg array ... which is to be expected.  That in turn creates a lot of insignificant diffs to be reported, as the generated variable names are different in each case.

I did notice a fair bit of this sort of thing going on as well,  which might be worth a closer look ?

(comparison of uncompressed output from gopherjs on the same code base, Linux on top, FreeBSD underneath)
```
7597c7600
<                       this.Blksize = new $Int64(0, 0);
---
>                       this.Blksize = 0;
7599,7602c7602,7605
<                       this.Atim = new Timespec.ptr(new $Int64(0, 0), new $Int64(0, 0));
<                       this.Mtim = new Timespec.ptr(new $Int64(0, 0), new $Int64(0, 0));
<                       this.Ctim = new Timespec.ptr(new $Int64(0, 0), new $Int64(0, 0));
<                       this.X__unused = arrayType$18.zero();
---
>                       this.Atim = new Timespec.ptr(0, 0);
>                       this.Mtim = new Timespec.ptr(0, 0);
>                       this.Ctim = new Timespec.ptr(0, 0);
>                       this.Ino = new $Uint64(0, 0);
```

The FreeBSD compiled code manages to get by without the overhead of the extra function call ?  Its inserting integer literals in the output instead of going through $int64(<int literal>) 


and some differences in handling pointers which are worth noting (Linux on top, FreeBSD underneath)  : 
```
130803c130770
<                       form$1[0] = new formulate.EditForm.ptr("", "", 0, sliceType.nil, $throwNilPointerError, $throwNilPointerError, $throwNilPointerError, $t
hrowNilPointerError, $throwNilPointerError);
---
>                       form$1[0] = new formulate.EditForm.ptr("", "", 0, sliceType.nil, $throwNilPointerError, $throwNilPointerError, $throwNilPointerError, $t
hrowNilPointerError);
```
Keeping in mind that either output works fine in the browser, there might be some room for optimization in the compiler, as highlighted by this output ?  Not sure.

LINK: Complete diff of the uncompressed output, if of any interest 
[GopherJS output Diff](http://map.wargaming.io/gopherjs.diff)

If you would like a live link to the compiled test app, send me a PM and Ill get back to you.

shurcooL:
> Summary: GOOS=linux works for FreeBSD, we can drop the PR and close the initial bug report. No code changes needed.

Great to hear, thanks for the update!

> The generated binaries from gopherjs come out to be different, despite both using GOOS=linux : 
> FreeBSD -> 5456199 bytes
> Linux -> 5468604 bytes

This sounds like it belongs to the same general category as #514.

@neelance said he resolved it, but it's possible it still occurs in some other cases.

If you have a smaller reproducible case, I think we should re-open that issue and add it there.

PR #521, opened 2016-09-19
Add support for source maps that use the local disk path of go source

nullstyle:
This PR adds a simple command line flag to the gopherjs tools that configures how the sourcemaps are generated during a build.  When using the `--localmap` flag, the sourcemaps will use the original local disk location instead of the default behavior.

This improves the development experience when serving gopherjs compiled code off the local disk, such as, in my case running an electron app powered by gopherjs.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -673,7 +673,7 @@ func (s *Session) WriteCommandPackage(archive *compiler.Archive, pkgObj string) 
 	return compiler.WriteProgramCode(deps, sourceMapFilter)
 }
 
-func NewMappingCallback(m *sourcemap.Map, goroot, gopath string, localmap bool) func(generatedLine, generatedColumn int, originalPos token.Position) {
+func NewMappingCallback(m *sourcemap.Map, goroot, gopath string, localMap bool) func(generatedLine, generatedColumn int, originalPos token.Position) {
 	return func(generatedLine, generatedColumn int, originalPos token.Position) {
 		if !originalPos.IsValid() {
 			m.AddMapping(&sourcemap.Mapping{GeneratedLine: generatedLine, GeneratedColumn: generatedColumn})
@@ -683,7 +683,7 @@ func NewMappingCallback(m *sourcemap.Map, goroot, gopath string, localmap bool) 
 		file := originalPos.Filename
 
 		switch hasGopathPrefix, prefixLen := hasGopathPrefix(file, gopath); {
-		case localmap:
+		case localMap:
 			// no-op:  keep file as-is
 		case hasGopathPrefix:
 			file = filepath.ToSlash(file[prefixLen+4:])
--- a/build/build.go
+++ b/build/build.go
@@ -322,15 +322,16 @@ func parseAndAugment(pkg *build.Package, isTest bool, fileSet *token.FileSet) ([
 }
 
 type Options struct {
-	GOROOT        string
-	GOPATH        string
-	Verbose       bool
-	Quiet         bool
-	Watch         bool
-	CreateMapFile bool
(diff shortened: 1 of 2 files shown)


build/build.go
```diff
@@ -672,21 +673,26 @@ func (s *Session) WriteCommandPackage(archive *compiler.Archive, pkgObj string)
 	return compiler.WriteProgramCode(deps, sourceMapFilter)
 }
 
-func NewMappingCallback(m *sourcemap.Map, goroot, gopath string) func(generatedLine, generatedColumn int, originalPos token.Position) {
+func NewMappingCallback(m *sourcemap.Map, goroot, gopath string, localmap bool) func(generatedLine, generatedColumn int, originalPos token.Position) {
```
shurcooL: Style-wise, this should be `localMap`. See https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps.

nullstyle: Duh, totally.  Thanks for the catch.  I'll update it momentarily.

nullstyle:
@shurcooL On 1.7, I'm getting the error below when trying to use to go tool to build against gopherjs.  Is there something I'm missing? 

```shell
$ GOARCH=js go build -compiler=js .
invalid value "js" for flag -compiler: unknown compiler "js"
```

shurcooL:
Sorry if I mislead you, but we're not in a post-#388 world yet. That issue is still an open proposal. It's not complete. See #485 for a WIP PR to do that, but it's still open.

nullstyle:
Ah, I see, thanks for the heads up.  

IMO, in a post-#388 world the most appropriate place to put the "local source maps" option would be as a `-gcflags` option.  I don't have any experience using gcflags myself, but have found evidence that it's the appropriate place to modify compiler behavior:  http://dave.cheney.net/2012/10/07/notes-on-exploring-the-compiler-flags-in-the-go-compiler-suite

shurcooL:
Yep, that's what I was thinking, `-gcflags` or something along those lines.

@neelance, what do you think about this PR?

tool.go
```diff
@@ -588,7 +592,7 @@ func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
 
 				sourceMapFilter := &compiler.SourceMapFilter{Writer: buf}
 				m := &sourcemap.Map{File: base + ".js"}
-				sourceMapFilter.MappingCallback = gbuild.NewMappingCallback(m, fs.options.GOROOT, fs.options.GOPATH)
+				sourceMapFilter.MappingCallback = gbuild.NewMappingCallback(m, fs.options.GOROOT, fs.options.GOPATH, false)
```
neelance: For consistency, please also add the option to the `serve` command and the others where it applies.

nullstyle: Thanks for the heads up.  Fixed.

neelance: I think `AddFlag` is still missing for `serve` and the other commands.

nullstyle: doy.  sorry, I didn't read very well last time.

I've pushed a change that adds `--localmap` support to other commands:  specifically `install`, `test`, `serve`, and `get`.  Being a gopherjs noob, I wasn't quite sure how this flag would apply to `gopherjs get`, but it mentioned removing a `.map` file when browsing the source so I included it in this change.  Please advise whether I should remove it.

shurcooL: `gopherjs get` is equivalent to `go get -d -tags=js` followed by `gopherjs install`. If `gopherjs install` needs it, then so does `gopherjs get`.

neelance:
LGTM. Thanks, @nullstyle!

PR #522, opened 2016-09-20
Exits with non-success exit code upon root command failure

nullstyle:
This PR improves the usability of gopherjs in shell scripts by ensuring that a failure due to unknown command line flags results in a non-success exit code.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -530,7 +530,7 @@ func main() {
 	rootCmd.AddCommand(cmdBuild, cmdGet, cmdInstall, cmdRun, cmdTest, cmdServe, cmdVersion)
 	err := rootCmd.Execute()
 	if err != nil {
-		os.Exit(-1)
+		os.Exit(2)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


tool.go
```diff
@@ -528,7 +528,10 @@ func main() {
 		Long: "GopherJS is a tool for compiling Go source code to JavaScript.",
 	}
 	rootCmd.AddCommand(cmdBuild, cmdGet, cmdInstall, cmdRun, cmdTest, cmdServe, cmdVersion)
-	rootCmd.Execute()
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(-1)
```
shurcooL: Let's follow the behavior of the `go` tool here.

```bash
~ $ go unknowncommand
go: unknown subcommand "unknowncommand"
Run 'go help' for usage.
~ $ echo $?
2
```

So, let's use `os.Exit(2)` here.

nullstyle: Makes sense.  I've pushed an update.

shurcooL:
Thanks!

PR #523, opened 2016-09-20
Fix an issue with watching command packages.

jgennis:
Watching a command package specified as a non-relative package name would only catch changes made to the imported packages, not the command package itself. This fixes the issue by watching the command package directory returned from Import rather than the package name as given on the command line.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -121,12 +121,21 @@ func main() {
 
 				for _, pkgPath := range args {
 					pkgPath = filepath.ToSlash(pkgPath)
-					if s.Watcher != nil {
-						s.Watcher.Add(pkgPath)
-					}
 					pkg, err := gbuild.Import(pkgPath, 0, s.InstallSuffix(), options.BuildTags)
 					if err != nil {
 						return err
+					}
+					if s.Watcher != nil {
+						s.Watcher.Add(pkg.Dir)
+
+						// We want to have the package dir watched when calling Import to avoid
+						// races where we could miss a change made and not re-trigger a build.
+						// However, we need to call Import to get the directory to watch, so we
+						// just call it again here to be sure about the race.
+						pkg, err = gbuild.Import(pkgPath, 0, s.InstallSuffix(), options.BuildTags)
+						if err != nil {
+							return err
+						}
 					}
 					archive, err := s.BuildPackage(pkg)
 					if err != nil {


tool.go
```diff
@@ -121,13 +121,22 @@ func main() {
...
-					}
 					pkg, err := gbuild.Import(pkgPath, 0, s.InstallSuffix(), options.BuildTags)
 					if err != nil {
 						return err
 					}
+					if s.Watcher != nil {
+						s.Watcher.Add(pkg.Dir)
+
+						// We want to have the package dir watched when calling Import to avoid
+						// races where we could miss a change made and not re-trigger a build.
```
shurcooL: Just want to make sure I understand this part.

Is the reason for the above because _calling Import may take a while_, so if a file is changed there while `gbuild.Import` is called, you want to be able to detect that. Or is there another reason?

If that's true, then why is it okay to not have package dir watched when the "first" `gbuild.Import` runs. Surely, it would take as long.

tool.go
```diff
@@ -121,13 +121,22 @@ func main() {
...
+						// However, we need to call Import to get the directory to watch, so we
+						// just call it again here to be sure about the race.
+						pkg, err = gbuild.Import(pkgPath, 0, s.InstallSuffix(), options.BuildTags)
+						if err != nil {
+							return err
+						}
+					}
```
shurcooL: Should the same change be applied in the similar code for install command on [line 199](https://github.com/gopherjs/gopherjs/blob/7600e3d439bd24b1a637bb6b67e9936c7d8736b3/tool.go#L199)?

tool.go
jgennis: Yes, the reason for calling import a second time is because a file may change after Import has processed it but before the file is watched. The reason the first Import is fine is that the build isn't being done based on it (since it's replaced by the second Import). If a file changes during the first import then the second import will use the changed file, and if a file changes during the second Import, then the Watcher will trigger another build.

tool.go
jgennis: Yes, I think so. Do you guys prefer I squash it into this change or add another change to the pull request?

jgennis:
I've been using `gopherjs serve` for a couple months with some client-only code, and I like it. Now I'm adding some server-side functionality, so I decided to write a little watch-and-restart command for the server code that I'm now running alongside `gopherjs build -w`, and it seems to be working alright so far, though I haven't done much with it yet.

tool.go
shurcooL: Either way is fine. We can squash it at the end when merging, so separate commits are better, since they're easier to review. Thanks!

shurcooL:
Thanks for explaining your motivation for using the watcher functionality, and how it compares to `gopherjs serve` mode. That's useful information for me.

Unfortunately, it's hard to support, so it's currently not included as part of a PR #485 I'm working on. But it helps to better understand how people use it, and it may come back there in another way. That PR is still far from being ready.

tool.go
shurcooL: This stuff is a little hard to reason about, but I _think_ that makes sense. Thanks for confirming.

neelance: Can we turn this around? Put the `if` where it was before, but add `pkg, err := gbuild.Import(pkgPath, 0, s.InstallSuffix(), options.BuildTags)` above the `s.Watcher.Add(pkg.Dir)`.

Also, please try using `gbuild.NewBuildContext(...).Import(...)` with `ImportMode` set to `FindOnly` instead of `gbuild.Import`, cause we're only interested in locating the package.

neelance: I'd still like to have the cleaner version of this code before merging. @jgennis Are you going to update your PR?

neelance:
Superseded by b3580af7536de2db7fd11da92972931f243c54b1 due to inactivity.

PR #528, opened 2016-09-29
Add support for executing examples

### LINKED ISSUE #239: Feature Request: `gopherjs test` support running and verifying example code.

flimzy:
This fixes #239. It is a followup to #380, now that #381 is fixed to add support for executing example code along with tests.

I think this is complete, but I want to do some additional testing before declaring it so.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -780,9 +780,10 @@ type testFuncs struct {
 }
 
 type testFunc struct {
-	Package string // imported package name (_test or _xtest)
-	Name    string // function name
-	Output  string // output, for examples
+	Package   string // imported package name (_test or _xtest)
+	Name      string // function name
+	Output    string // output, for examples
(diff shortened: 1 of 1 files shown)


flimzy:
Examples are working, in general. But there is still some corner case where GopherJS hangs (see #529). I don't expect this is directly related to Examples, and as such it might be appropriate to merge this before solving #529, but running all upstream examples automatically in CircleCI is not an option before solving #529. But it's probably worth at least investigating #529 more fully before making that call (which I intend to do as time permits).

neelance:
LGTM, thanks @flimzy!

Issue #529, opened 2016-09-29
os.Read() appears to block other goroutines from executing


flimzy:
In some situation I have yet to specifically identify, GopherJS hangs.  I'm creating this issue to track my debugging progress.

I found, while working on #528, that the first example for the upstream 'html/template' package causes a freeze.  I have narrowed this down to a stand-alone reproduction case below.  But I still need to work on narrowing it down further.

When I execute the following code in Go:

    $ go run jstest.go -test.v
    === RUN   Example
    --- PASS: Example (0.00s)
    PASS

Or in GopherJS:

    $ gopherjs run jstest.go -- -test.v
    === RUN   Example

(and never exits)

	package main

	import (
		"html/template"
		"log"
		"os"

		"regexp"
		"testing"
	)

	var tests = []testing.InternalTest{}
	var benchmarks = []testing.InternalBenchmark{}

	var examples = []testing.InternalExample{
		{"Example", Example, "<!DOCTYPE html>\n<html>\n\t<head>\n\t\t<meta charset=\"UTF-8\">\n\t\t<title>My page</title>\n\t</head>\n\t<body>\n\t\t<div>My photos</div><div>My blog</div>\n\t</body>\n</html>\n<!DOCTYPE html>\n<html>\n\t<head>\n\t\t<meta charset=\"UTF-8\">\n\t\t<title>My another page</title>\n\t</head>\n\t<body>\n\t\t<div><strong>no rows</strong></div>\n\t</body>\n</html>\n", false},
	}

	var matchPat string
	var matchRe *regexp.Regexp

	func matchString(pat, str string) (result bool, err error) {
		if matchRe == nil || matchPat != pat {
			matchPat = pat
			matchRe, err = regexp.Compile(matchPat)
			if err != nil {
				return
			}
		}
		return matchRe.MatchString(str), nil
	}

	func main() {
		m := testing.MainStart(matchString, tests, benchmarks, examples)

		os.Exit(m.Run())
	}

	func Example() {
		const tpl = `
	<!DOCTYPE html>
	<html>
		<head>
			<meta charset="UTF-8">
			<title>{{.Title}}</title>
		</head>
		<body>
			{{range .Items}}<div>{{ . }}</div>{{else}}<div><strong>no rows</strong></div>{{end}}
		</body>
	</html>`

		check := func(err error) {
			if err != nil {
				log.Fatal(err)
			}
		}
		t, err := template.New("webpage").Parse(tpl)
		check(err)

[...]

PR #530, opened 2016-09-30
Document that system calls block other goroutines.

flimzy:
This addresses #529.


flimzy:
I hope my explanation is technically accurate.

PR #540, opened 2016-10-23
README: Remove google group link.

shurcooL:
The google group for GopherJS has been deprecated (it's read-only now). It has bad UI, and it was used very infrequently. Messages posted there were often not seen for weeks by anyone. It's better to have them posted elsewhere.

Please use the issue tracker and slack channel instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@ If you include an argument, it will be the root from which everything is served.
 
 ### Community
 - [#gopherjs Channel on Gophers Slack](https://gophers.slack.com/messages/gopherjs/) (invites to Gophers Slack are available [here](http://blog.gopheracademy.com/gophers-slack-community/#how-can-i-be-invited-to-join:2facdc921b2310f18cb851c36fa92369))
-- [Google Group](https://groups.google.com/d/forum/gopherjs)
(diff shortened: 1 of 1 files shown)


dominikh:
I strongly recommend having *some* form of forum that is not the issue tracker. Using the issue tracker for conversations tends to be a very bad idea, and while real-time communication can be nice, it's not easily searchable.

shurcooL:
> I strongly recommend having some form of forum that is not the issue tracker. Using the issue tracker for conversations tends to be a very bad idea, and while real-time communication can be nice, it's not easily searchable.

I agree, and I propose we use reddit for that purpose. It has much nicer UI, and I think it's a great replacement for Google Groups (and its awful, complicated, outdated UI).

The Go project uses reddit too (see https://www.reddit.com/r/golang/), and I like it quite a bit.

shurcooL:
Another possible replacement for Google Groups would be [Discourse](https://www.discourse.org/). I've never used it myself, but I heard good things about it...

shurcooL:
Well, I'll merge this since this PR updates README to better match reality.

Since it's unclear which alternative forum to consider adding, we can discuss and do that after this PR, as needed.

PR #555, opened 2016-12-17
Fixed file leak for serving

itsmontoya:


tool.go
```diff
@@ -636,12 +636,14 @@ func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
 		if err == nil {
 			return f, nil
 		}
+		defer f.Close()
```
shurcooL: This is not valid for two reasons:

1. The `if` statement above checks if `err == nil`. If so, it returns `f, nil`.

    Otherwise, execution continues and gets to this point. We know that `err != nil` must be the case. That means `f` is not a valid file and there's no need to close it. In fact, it may be nil, so this could panic with nil pointer dereference here.

2. We are inside the `serveCommandFileSystem.Open(requestName string) (http.File, error)` method. It implements [`http.FileSystem`](https://godoc.org/net/http#FileSystem) interface. The `Open(name string) (File, error)` method, similar to [`os.Open`](https://godoc.org/os#Open), opens a file and returns it. It's the user's responsibility to close it when they are done using it. `Open` both cannot and should not try to close the file (otherwise, it's already closed by the time it's returned, which defeats the purpose).

tool.go
```diff
@@ -636,12 +636,14 @@ func (fs serveCommandFileSystem) Open(requestName string) (http.File, error) {
...
 
 		// source maps are served outside of serveRoot
 		f, err = dir.Open(requestName)
 		if err == nil {
 			return f, nil
 		}
+		defer f.Close()
```
shurcooL: Same comment here.

tool.go
neelance: `f` should be `nil` in case `err != nil`, right?

shurcooL: Yes, @neelance, it doesn't have to be, but it could be (and most likely is). See my review comments above.

shurcooL:
@itsmontoya, see my review comments above.

What prompted you to make this PR, did you see a leak and attempt to find its source? Or did you look at the code and think this would be a source of leaks?

If there is a leak, it would have to be elsewhere. But first I want to find out if there's any known evidence of a leak.

itsmontoya:
After talking with @shurcooL we realized that the issue wasn't located where I thought. I'm closing this PR as it's not fixing the issue. 

I left some notes with @shurcooL which should aid in fixing the leak issue eventually. It appears the issue is with the "-w" flag while serving. I will be testing without that flag to see if the problem persists.

shurcooL:
> After talking with @shurcooL

For posterity, see https://gophers.slack.com/archives/gopherjs/p1481938673000166 and messages below.

Issue #554, opened 2016-12-16
Slow crypto operations (WebCryptoAPI support)


r-l-x:
Crypto operations are really slow with gopherjs.
Here is a small benchmark, consisting of hashing a 10MB buffer through SHA-256: http://www.gopherjs.org/playground/#/X98vxh8Y97
- native go: about 120ms
- gopherjs: about 1400ms under Chrome, 1900ms under Firefox 

Depending on the type of operation, the difference can be even bigger. Issue #536 mentions 300x slower operations for RSA signatures.

I think it would be really great if crypto operations were implemented by calling the WebCryptoAPI.
As an exemple, here is a simple POC, with the same operation implemented with WebCrypto: http://www.gopherjs.org/playground/#/-_Ed7QocdZ
The execution times are:
- Chrome (you have to load it over HTTPS): about 50ms
- Firefox: about 70ms

**That's about 25x faster than the current version, and even twice faster than a native go program!**

r-l-x:
A lot of work has already been done by @kothar here: https://bitbucket.org/kevalin-p2p/subtlecrypto
But one of the problem is that the various features of WebCrypto are not supported in every browser (See https://diafygi.github.io/webcrypto-examples), and there are some specific restrictions like with Chrome: WebCryptoAPI support is only enabled for pages loaded over HTTPS, so it would be necessary to have a fallback using the current go implementation when the WebCryptoAPI is not available.
Currently with gopherjs, it's not easy to override a function of the standard library while being able to call the original function, as original overridden functions are renamed "_".
So we need to have a way to override functions, while being able to call the original function. I'm not sure what's the best way to do this.

kothar:
I had similar thoughts - my initial plan was to create an isomorphic library that could fall back to the stdlib implementation if it couldn't find a browser implementation, but that doesn't help any of the other standard library code which uses crypto. Specifically net/http2 for what I was working on, although I feel like that's a bizarre use case anyway.

flimzy:
A general method for fallbacks like that would probably be useful in many areas, not just crypto.

r-l-x:
I guess one way to do it would be to rename the original overridden function e.g. by adding a suffix, allowing it to be called from the overriding function.

shurcooL:
I was considering that too. For example, by always adding a prefix such as `_` or `_GopherJS_` to the identifier being replaced.

[...]

PR #558, opened 2016-12-21
WIP: WebcryptoAPI support

r-l-x:
DO NOT MERGE

See #554


shurcooL:
> I didn't notice any problem with dead code elimination (on my current project, the generated javascript code has slightly the same size), but I'm not sure if there are other points we can check ?

Nice, that's promising! I will try it out on a larger project (that imports more of standard library) and see how it does.

It makes sense that the difference wouldn't be huge, because GopherJS overrides a modest, not large, amount of standard library identifiers. So even if they're now included, it's likely not to be a big deal. And if nothing references them, it's possible DCE still eliminates them. I don't know how advanced the current DCE algorithm is and how this will affect it, perhaps @neelance would know.

compiler/natives/src/crypto/sha256/sha256.go
```diff
@@ -0,0 +1,67 @@
...
+
+package sha256
+
+import (
+	"errors"
+
+	"github.com/gopherjs/gopherjs/js"
+)
+
+var ErrWebCryptoFailed = errors.New("crypto: Web Crypto API call failed")
```
shurcooL: Should this be an exported variable? I don't think we want user code to be able to use it.

compiler/natives/src/crypto/sha256/sha256.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	subtle := crypto.Get("subtle")
+	if subtle == js.Undefined {
+		subtle = crypto.Get("webkitSubtle") // for Safari
+		if subtle == js.Undefined {
+			return nil
+		}
+	}
+	return subtle
+}
```
shurcooL: Would it make more sense to execute this code once, at package init time, and save results.

Instead of calling it once for every `crypto.Sum256` call?

Similar to how `net/http` package does it for figuring out the transport to use.

compiler/natives/src/crypto/sha256/sha256.go
```diff
@@ -0,0 +1,67 @@
...
+	})
+	promise.Call("catch", func(*js.Object) {
+		errCh <- ErrWebCryptoFailed
+	})
+	select {
+	case err = <-errCh:
+	case resSlice:= <-resCh:
+		copy(res[:], resSlice)
+	}
+	return
```
shurcooL: Style-wise, I highly suggest avoiding naked returns because they're less readable.

> Naked returns are okay if the function is a handful of lines. Once it's a medium-sized function, be explicit with your return values. Corollary: it's not worth it to name result parameters just because it enables you to use naked returns.

Source: https://github.com/golang/go/wiki/CodeReviewComments#naked-returns

shurcooL:
Overall, this looks really good! Thanks for working on this @r-l-x.

I've left some comments that came to mind looking at it, but this is a great start.

compiler/natives/src/crypto/sha256/sha256.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func Sum256(data []byte) [Size]byte {
+	res, err := jsSum256(data)
+	if err != nil {
+		// The WebCryptoAPI call failed: fallback by calling the implementation from
+		// the Go standard library
+		return _gopherjs_overriden_Sum256(data)
+	}
```
r-l-x: That's a good point.
Generally speaking for crypto functions, the only cases I can think of where I would expect this to happen are:

- When the function or one of the parameters is not supported (but the function is still defined). I'm not sure which exception we will get though, I guess a NotSupportedError.
- With Chrome when the origin is insecure (NotSupportedError).
- When there is a problem with the arguments, e.g. (depending on the algorithm) decryption with a wrong key or wrong authentication tag  (According to the W3C spec, we should get an OperationError in this case).

We could try to check if the error code is NotSupportedError and then remember that the webcrypto implementation of this function shouldn't be called, but NotSupportedError may also mean that while the function itself is supported, a parameter, typically a key length or mode of operation is not supported (see https://diafygi.github.io/webcrypto-examples/), so that would be counterproductive in this case to globally avoid the JS implementation of the function.

The performance impact of calling the JS implementation when we get a NotSupportedError is about 5ms under Chrome on insecure origin: http://www.gopherjs.org/playground/#/jlVJjSdJeh

flimzy: 5ms doesn't sound like enough to worry about to me. At least not at this stage.  Micro-optimizations like that should come much later :)

r-l-x:
I guess the next step is to implement a generic way to call a JS function returning a promise, so that we don't have to create a channel, implement the goroutines and call select for each crypto function we want to wrap. The idea is to have something similar to what @kothar did [here](https://bitbucket.org/kevalin-p2p/subtlecrypto/src/6b6ae02d84a3baed2d22b8be3c1cf9e242cf314a/subtlecrypto.go?at=master&fileviewer=file-view-default#subtlecrypto.go-23:51).

@shurcooL do you think we could add a `CallAsync` function in `js/js.go` ? This kind of construction would be useful globally, e.g. for the [fetch transport](https://github.com/gopherjs/gopherjs/blob/master/compiler/natives/src/net/http/fetch.go#L22-L44).

compiler/natives/src/crypto/sha256/sha256.go
r-l-x: This is right with the current code, but in the near future this error code will be useful for each of the modules under crypto.

r-l-x:
Another point that would be really useful is to be able to check the error type when a promise fails. Typically, for a decryption function, depending on the mode (let's say AES-GCM), an `OperationError` will be thrown if the tag check fails, indicating that the key is incorrect or the message was corrupted. In this case, it's useless to call the native go function. But if the exception thrown is `NotSupportedError`, then it's totally appropriate to fallback to the go standard library.

So I would typically like to call instanceof in order to check the JS error type, but apparently it's not supported as is in gopherjs, it seems that [jsbuiltin](https://github.com/gopherjs/jsbuiltin) is needed for that ?

What would you recommend in order to do this in gopherjs ?

flimzy:
> So I would typically like to call instanceof in order to check the JS error type, but apparently it's not supported as is in gopherjs, it seems that jsbuiltin is needed for that ?

If you don't want to depend on `jsbuiltin` (probably should avoid a dependency on that, for the standard lib), you could copy the functionality, by including your own wrapper around `instanceOf` in an included .js file, the same way `jsbuiltin` does it.

flimzy:
> I guess the next step is to implement a generic way to call a JS function returning a promise, so that we don't have to create a channel, implement the calllbacks and call select() for each crypto function we want to wrap. The idea is to have something similar to what @kothar did here.

This can also be done with `sync.WaitGroup`, but I think a channel is the right approach to avoid a dependency on the `sync` package.

No reason you can't abstract that into your own helper function(s) for use in this lib, though.

r-l-x:
> No reason you can't abstract that into your own helper function(s) for use in this lib, though.

Yes, I agree. I was thinking that it would be convenient to have a method with a `*js.Object` receiver (like [Call()](https://github.com/gopherjs/gopherjs/blob/master/js/js.go#L50)), but I can do something similar in the crypto package for now.

flimzy:
> Yes, I agree. I was thinking that it would be convenient to have a method with a *js.Object receiver (like Call()), but I can do something similar in the crypto package for now.

That does indeed sound generally useful.

r-l-x:
Another thing that would be really helpful: Is there an easy way to launch the tests of the Go standard library in the browser ?

kothar:
Not sure if it helps, but this was the best way I could find to run tests in the browser. Very useful when I was working on the crypto stuff with different node/browser behaviour.
https://bitbucket.org/mikehouston/browsertests/src

It may be possible to use by importing that package into the stdlib tests, but I've not tried it.

compiler/natives/src/crypto/subtle/subtle.go
```diff
@@ -36,3 +38,37 @@ func getSubtle() *js.Object {
...
 	SubtleCrypto = getSubtle()
 }
+
+// SubtleCall calls the object's method with the given arguments assuming that it returns a promise, wait for it to be settled, and return the result/error.
+func SubtleCall(method string, args ...interface{}) (*js.Object, error) {
+	if SubtleCrypto == nil {
+		return nil, ErrCryptoNotFound
+	}
+
+	// First check that the method exists (if it doesn't, Call() would panic, so it's better to check first)
```
flimzy: Not saying this is a bad approach, but an alternative would be to catch the panic with a `recover` in a deferred function. Doing so might also be a bit more robust, as it would catch any other runtime panics in an odd JS implementation or something.

flimzy:
> Another thing that would be really helpful: Is there an easy way to launch the tests of the Go standard library in the browser ?

You can compile any test to a .js file, then load it in a browser and watch the JS console for output.  It's a bit cumbersome, but it works.

I spent some time about a year ago on trying to make this work better, but the GopherJS runtime is missing some features that I think are essential to automate the process.

compiler/natives/src/crypto/sha256/sha256.go
shurcooL: Can you elaborate on how it will be used?

I'm concerned about this because I don't think we should be adding to the API surface of `crypto/...` packages in standard library, as that would make user code incompatible with Go.

shurcooL:
I haven't tried and don't know if it's possible, but can you change the type of a top-level package identifier?

E.g., the real `sha256.Sum256` in https://godoc.org/crypto/sha256#Sum256 is a `func`:

```Go
func Sum256(data []byte) [Size]byte {
	...
}
```

If we can change the type while overriding, we could make it a variable, like this:

```Go
var Sum256 = func() func(data []byte) [Size]byte {
	if webCryptoAvailable {
		return webCryptoImplementationOfSum256
	}
	return _gopherjs_overridden_Sum256
}()
```

Have you already considered doing it like that? Thoughts?

r-l-x:
> ```
> var Sum256 = func() func(data []byte) [Size]byte {
> 	if webCryptoAvailable {
> 		return webCryptoImplementationOfSum256
> 	}
> 	return _gopherjs_overridden_Sum256
> }()
> ```

I see your point, but one of the problems with this kind of approach is to know when we should call the webcrypto implementation. It doesn't seem feasible to me to determine this a priori. Let's take a few examples:

- If the code is run from an insecure origin in Chrome, the SubleCrypto interface is available, but whenever a function is called, it will fail with a NotSupportedError.
- The same thing will happen (SubtleCrypto available, but NotSupportedError thrown) when you call a method that's not supported by the browser (and there are a few, even with recent Chrome/Firefox, see https://diafygi.github.io/webcrypto-examples/).
- And it can be even more complicated because a method may be supported by the browser but the specific algorithm or key length you requested may be unsupported (e.g. AES-GCM 192bits key generation with Chrome).

So it makes sense to me to first call the webcrypto implementation and then, if it failed because the requested operation was not supported, call the go implementation.
I guess the ideal behavior would be to remember that this specific method (including algorithm/key length) failed, so that the next time the program calls it we don't have to try the webcrypto implementation, but this seems quite complicated and error prone.

compiler/natives/src/crypto/sha256/sha256.go
r-l-x: > Can you elaborate on how it will be used?

The idea is to have a specific error code returned when a webcrypto method fails with a `NotSupportedError`. Then most cryptographic functions implementation (in each `crypto/*` package) would look like:
```
func foo() error {
  err := webcrypto_foo()
  if err == ErrWebCryptoUnsupported {
    err = _gopherjs_overriden_foo()
  }
  return err
}
```
In order to make this error code visible from every package under `crypto`, the only way I can think of is to make it public.

> that would make user code incompatible with Go.

I didn't think about that. In the last version of the code, I moved the the non standard public stuff in a new package [crypto/subtle](https://github.com/gopherjs/gopherjs/blob/985925063ca092307686be8555800313ad780190/compiler/natives/src/crypto/subtle/subtle.go
). Is that ok? Should we put this kind of code somewhere else (like under the `js` package) ?

r-l-x:
I'm looking at how to run the standard library tests in the browser. With gopherjs test,  it's really easy to run in the command line:
```
$ gopherjs test -v crypto/sha256
=== RUN   TestGolden
--- PASS: TestGolden (0.17s)
=== RUN   TestSize
--- PASS: TestSize (0.00s)
=== RUN   TestBlockSize
--- PASS: TestBlockSize (0.00s)
=== RUN   TestBlockGeneric
--- PASS: TestBlockGeneric (0.01s)
PASS
ok  	crypto/sha256	0.595s

```
But when I generate the code using `gopherjs test -v -c crypto/sha256`, and load the generated javascript in a browser, I only see in my console:
```
syscall.go:43 PASS
syscall.go:25 warning: system calls not available, see https://github.com/gopherjs/gopherjs/blob/master/doc/syscalls.md

```

@kothar 
> Not sure if it helps, but this was the best way I could find to run tests in the browser. Very useful when I was working on the crypto stuff with different node/browser behaviour.
https://bitbucket.org/mikehouston/browsertests/src
It may be possible to use by importing that package into the stdlib tests, but I've not tried it.

If I get it right, the trick is to inject the desired arguments in `os.Args` at init time, is that right ?

kothar:
@r-l-x Yep, looks like you got it working, in case it is helpful, here's the original discussion with @shurcooL about tests: https://gophers.slack.com/archives/gopherjs/p1472378655000172

build/build.go
```diff
@@ -290,7 +290,7 @@ func parseAndAugment(pkg *build.Package, isTest bool, fileSet *token.FileSet) ([
 			switch d := decl.(type) {
 			case *ast.FuncDecl:
 				if replacedDeclNames[funcName(d)] {
-					d.Name = ast.NewIdent("_")
+					d.Name.Name = "_gopherjs_overriden_" + d.Name.String()
```
shurcooL: You've addressed point 2 but still not 1. Did you miss it, or are you not making the change because you disagree it's an improvement?

r-l-x: Sorry, that was a mistake from me. I'll update the code.

r-l-x: Point 1 is now fixed as suggested.

r-l-x:
Ok the current implementation seems good to me, with great performance improvements. We could of course add more cryptographic fonction implementations later, but it would be actually quite tricky for most of them.
For example, in order to implement RSA OAEP encryption, I think we need to use the reflection to recognize what the provided hash fonction is. This is feasible, but I guess that could come in a second (probably more controversial) PR.

What do you think ?

compiler/natives/src/crypto/sha256/sha256.go
```diff
@@ -26,9 +26,25 @@ func webCryptoSum256(data []byte) (res [Size]byte, err error) {
...
+		// |        |    Go     | webcrypto |    Go     | webcrypto |
+		// +--------+-----------+-----------+-----------+-----------+
+		// | 1B     | 83ns*     | 2.004ms   | 116ns*    | 149ns     |
+		// | 1kB    | 818ns*    | 2.003ms   | 334ns     | 152ns*    |
+		// | 2.5kB  | 2.072ms*  | 2.011ms   | 672ns     | 159ns*    |
+		// | 2.6kB  | 2.175ms   | 2.020ms*  | 689ns     | 159ns*    |
+		// | 2.7kB  | 2.400ms   | 2.022ms*  | 710ns     | 160ns*    |
+		// | 100kB  | 774ms     | 24ms*     | 215ms     | 16ms*     |
+		// | 10MB   | 1.632s    | 41ms*     | 1.76s     | 74ms*     |
+		// +--------+-----------+-----------+-----------+-----------+
```
flimzy: Do you have an easy way to run these benchmarks?  I'd be curious to see how these results compare on different hardware configurations.

shurcooL:
> This is feasible, but I guess that could come in a second (probably more controversial) PR. What do you think ?

I agree, let's leave out additional enhancements for future PRs. It's better to get this one fully reviewed and merged first.

compiler/natives/src/crypto/sha256/sha256.go
neelance: Why is the webcrypto version slower for small buffers? Where does it spend the time?

compiler/natives/src/crypto/rand/rand.go
```diff
@@ -5,6 +5,7 @@ package rand
 import (
 	"errors"
 
+	"github.com/gopherjs/gopherjs/compiler/natives/src/crypto/webcrypto"
```
neelance: I think `github.com/gopherjs/gopherjs/js/webcrypto` might be better, since `natives` is not meant to be imported directly.

compiler/natives/src/crypto/rand/rand.go
```diff
@@ -19,20 +20,16 @@ func (r *rngReader) Read(b []byte) (n int, err error) {
...
 
 	// browser
-	crypto := js.Global.Get("crypto")
-	if crypto == js.Undefined {
-		crypto = js.Global.Get("msCrypto")
-	}
-	if crypto != js.Undefined {
-		if crypto.Get("getRandomValues") != js.Undefined {
+	if webcrypto.Crypto != nil {
+		if webcrypto.Crypto.Get("getRandomValues") != js.Undefined {
```
neelance: Please merge the `if` statements.

compiler/natives/src/crypto/sha256/sha256.go
r-l-x: @flimzy I mostly used the gopherJS playground, I don't have the link to the script right now, but I can add it to the discussion when I'm back on my PC.

@neelance I don't know where the time is spent, but from the measurements I did, there is definitely something specific to Chrome, it seems like a webcrypto promise cannot give an answer in less than 2ms. It would be interesting to see if it's the same for any promise. Firefox doesn't have this problem. I can try to profile it.

r-l-x: @flimzy Here is a rather clean way to launch the benchmarks: http://www.gopherjs.org/playground/#/dPF9AaIxVg. Works as is with Firefox, but with Chrome in order to avoid the insecure origin problem, you have to copy the code, load the playground over HTTPS while accepting the security warning and paste the code.

r-l-x: @neelance From the profiling I did (using this in Chrome: http://www.gopherjs.org/playground/#/dPF9AaIxVg), the main thread is idle for most of the time, waiting for the response from the webcrypto API. I'm not sure how to profile the webcryptoAPI thread.

flimzy: On my machine (Lenovo T430s, Linux 4.2.0), webcrypto outperforms Go for any buffer > 2kb in both Chromium (Debian-branded Chrome) and Firefox. See here: https://gist.github.com/flimzy/7732aba13ca83e6aa3b4c615d7e0c1b0

compiler/natives/src/crypto/rand/rand.go
r-l-x: I did the move: https://github.com/gopherjs/gopherjs/pull/558/commits/02e03500ee7cd665b1dfad6b47240e4dc808f46a

compiler/natives/src/crypto/rand/rand.go
r-l-x: [done](https://github.com/gopherjs/gopherjs/pull/558/commits/ccf8b1e64be9bb4c968d3cfa2d390df20473f235).

compiler/natives/src/crypto/sha256/sha256.go
r-l-x: @neelance My link was wrong, [this ](http://www.gopherjs.org/playground/#/7uJwxm1UYd)is what I used to profile.

r-l-x: @neelance FYI I don't reproduce this "slow webcrypto start" using native Javascript. Here is the code I used: https://jsbin.com/xukuqaqesa/1/edit?js,console. An iteration runs in 2µs vs 2ms for gopherjs. So it seems like it's gopherjs related.

neelance: @r-l-x Please debug where exactly the latency happens. How long does it take for the promise to return vs. the Go code to continue running? I think this may be related to the blocking and channel operations.

r-l-x: @neelance I think you are right about the channel operation. This example seams to show that the latency occurs between the channel write (in the promise callback) and the channel read outside the promise: http://www.gopherjs.org/playground/#/btm0GV_mQY

neelance: Hey @r-l-x, could you please try the benchmarks again with the `scheduler` branch of GopherJS? I think it may resolve the 2ms latencies you were seeing.

neelance: The branch got merged into `master`, please test that one instead. :-)

neelance:
@r-l-x Any plan to test this again on latest master to see if we can remove the `if len(data) >= 2600`?

build/build.go
```diff
@@ -289,8 +289,10 @@ func parseAndAugment(pkg *build.Package, isTest bool, fileSet *token.FileSet) ([
 		for _, decl := range file.Decls {
 			switch d := decl.(type) {
 			case *ast.FuncDecl:
+				// Allow overridden function calls
+				// The standard library implementation of foo() becomes _gopherjs_overriden_foo()
 				if replacedDeclNames[funcName(d)] {
-					d.Name = ast.NewIdent("_")
+					d.Name.Name = "_gopherjs_overridden_" + d.Name.Name
```
flimzy: The comment for this function is no longer accurate after these changes.

compiler/natives/src/crypto/sha256/sha256.go
lologarithm: @neelance is latest master deployed to gopherjs playground?

Running https://gopherjs.github.io/playground/#/xN-qiC0TQk on gopherjs playground right now results in len 2000 being slightly faster for webcrypto, 1750 being about equal, and <=1500 being faster for the go implementation 

(running on ubuntu16.04 x64 with chrome)

lologarithm: Also ran the 'stop benchmark after result, before sending result in channel' test and am seeing webcrypto faster until 64 bytes and less

len: 1 
go           20000	     58100 ns/op	   0.02 MB/s 
webcrypto     1000	   1267000 ns/op	   0.00 MB/s 
webcrypto No Chan       133800 ns/op
 
len: 64 
go           10000	    103200 ns/op	   0.62 MB/s 
webcrypto     1000	   1255000 ns/op	   0.05 MB/s 
webcrypto No Chan       134000 ns/op

len: 128
go           10000	    142300 ns/op	   0.90 MB/s 
webcrypto     1000	   1248000 ns/op	   0.10 MB/s 
webcrypto No Chan       134800 ns/op

len: 256 
go            5000	    221800 ns/op	   1.15 MB/s 
webcrypto     1000	   1258000 ns/op	   0.20 MB/s 
webcrypto No Chan       137200 ns/op

len: 512 
go            3000	    402000 ns/op	   1.27 MB/s 
webcrypto     1000	   1255000 ns/op	   0.41 MB/s 
 
len: 1024 
go            2000	    705000 ns/op	   1.45 MB/s 
webcrypto     1000	   1256000 ns/op	   0.82 MB/s 
 
len: 1200 
go            2000	    824000 ns/op	   1.46 MB/s 
webcrypto     1000	   1246000 ns/op	   0.96 MB/s 
 
len: 1500 
go            2000	   1012500 ns/op	   1.48 MB/s 
webcrypto     1000	   1254000 ns/op	   1.20 MB/s

neelance:
@r-l-x I think this PR is valuable and I'd like to move forward with it. Unfortunately it is not ready to be merged.

Maybe someone else wants to pick up this PR and get it to the finish line?

r-l-x:
Sorry guys, I was really busy for the last 2 weeks and didn't have time to work on this, but I should be able to work on this in the next few days.

neelance:
All right, no problem at all. I just wanted to get some kind of feedback so I know that you're still on board with this. :+1:

build/build.go
```diff
@@ -307,8 +307,10 @@ func parseAndAugment(pkg *build.Package, isTest bool, fileSet *token.FileSet) ([
...
 			switch d := decl.(type) {
 			case *ast.FuncDecl:
+				// Allow overridden function calls
+				// The standard library implementation of foo() becomes _gopherjs_overridden_foo()
 				if replacedDeclNames[funcName(d)] {
-					d.Name = ast.NewIdent("_")
+					d.Name.Name = "_gopherjs_overridden_" + d.Name.Name
 				}
 			case *ast.GenDecl:
 				switch d.Tok {
```
flimzy: I'm mentioning this again so it doesn't get lost in the shuffle, now that GitHub has hidden my original note.

The comment for this function is outdated by the changes you've made.  In particular, this text is no longer correct:

     // The native packages are augmented by the contents of natives.FS in the following way.
     // The file names do not matter except the usual `_test` suffix. The files for
     // native overrides get added to the package (even if they have the same name
     // as an existing file from the standard library). For all identifiers that exist
     // in the original AND the overrides, the original identifier in the AST gets
     // replaced by `_`. New identifiers that don't exist in original package get added.

r-l-x: That's right, here is the updated comment: https://github.com/gopherjs/gopherjs/pull/558/commits/febab31b01360451d242cd77388b1aaac0f2c828

r-l-x:
The work on the scheduler brings great improvements: hashing a 1 byte buffer using webcrypto now takes about 60µs (vs 2ms before) with Chrome. For small buffers, the performances stay in the same order of magnitude between go and webcrypto implementations (and this is true both on Firefox and on Chrome).

Previously, the strategy in this branch was to use webcrypto for buffers bigger than 2.6kB, and native go for smaller buffers, but I think we can now change this to always try the webcrypto implementation if possible. It's now faster starting for buffers around 500B in my tests. It's still slower for smaller buffers, but another point in favor of the webcrypto implementation is that it doesn't block the javascript thread, so we can update the UI or execute another goroutine during the computation where this is impossible with the native go implementation.

What do you think ?

r-l-x:
[Tests fail](https://circleci.com/gh/gopherjs/gopherjs/1097) with the following error, which seems like the expected behavior to me:

`augmented normal package "crypto/rand" imports "github.com/gopherjs/gopherjs/js/webcrypto" in file /src/crypto/rand/rand.go, but real "crypto/rand" doesn't`

(See https://circleci.com/gh/gopherjs/gopherjs/1097)

Is it ok for me to whitelist `github.com/gopherjs/gopherjs/js/webcrypto` like it's done [here](https://github.com/gopherjs/gopherjs/blob/f2fc182abb28925c7aa40502db839c6ebd7f7171/build/build_test.go#L97-L98) for `github.com/gopherjs/gopherjs/js` ?

shurcooL:
Sure, we can treat `github.com/gopherjs/gopherjs/js/webcrypto` as an exception since it's not in standard library. The goal of that test is to ensure we don't modify the standard library in incompatible ways.

r-l-x:
The `reflect` test [is failing on circleCI](https://circleci.com/gh/gopherjs/gopherjs/1098) with `ReferenceError: VD is not defined`.

Does anyone has an idea why this occurs?

r-l-x:
I also realize that there is a problem with empty slices which are apparently not converted as expected in JavaScript:
[This exemple](https://gopherjs.github.io/playground/#/7jKNdYjy_N) outputs `TypeError: Failed to execute 'digest' on 'SubtleCrypto': The provided value is not of type '(ArrayBuffer or ArrayBufferView)' ` for the empty slice.

Any idea on this one ?

flimzy:
> I also realize that there is a problem with empty slices which are apparently not converted as expected in JavaScript:
This exemple outputs TypeError: Failed to execute 'digest' on 'SubtleCrypto': The provided value is not of type '(ArrayBuffer or ArrayBufferView)' for the empty slice.

> Any idea on this one ?

Are you using the latest version of GopherJS?  I just fixed a bug about two weeks ago relating to externalization of empty slices.  See #599 and #597.

shurcooL:
I was about to say exactly what @flimzy said above.

If it's still an issue on the latest master, please file an issue.

r-l-x:
@shurcooL @flimzy You are right, the problem doesn't occur anymore since the rebase (I was testing on the playground, which apparently is not up to date).
So I would say that the the last blocking point is [this problem](https://circleci.com/gh/gopherjs/gopherjs/1099) with the `reflect` test.

shurcooL:
> @shurcooL @flimzy You are right, the problem doesn't occur anymore since the rebase (I was testing on the playground, which apparently is not up to date).

I've updated the playground to latest GopherJS version locally, and I can confirm it fixes that issue. I'm going to push the playground update soon.

image

shurcooL:
> So I would say that the the last blocking point is [this problem](https://circleci.com/gh/gopherjs/gopherjs/1099) with the `reflect` test.

I can't see the error message in Circle CI. I just keep seeing a spinning circle, and an error in browser console:

image

Can you paste the failure log here if you can access it?

r-l-x:
> I can't see the error message in Circle CI. I just keep seeing a spinning circle, and an error in browser console:


Yes, I'm not sure why, but in order to see the error with CircleCI, I need to fold and unfold the red part.

Here is our error:
```
/home/ubuntu/gopherjs/test.790433663:39
$packages["reflect_test"]=(function(){var $pkg={},$init,E,F,S,G,A,T,H,B,I,J,C,K,L,M,N,O,D,P,Q,R,AP,AQ,AR,BQ,BS,BT,BU,BY,BZ,CA,CI,CK,CN,CZ,DG,DJ,DK,DV,DZ,EA,EB,EC,ED,EE,EF,EH,EI,EJ,EK,EN,EP,EQ,ER,ES,ET,EU,EV,EW,EX,EY,EZ,FA,FB,FC,FD,FE,FF,FG,FH,FI,FJ,FK,FR,FS,FU,FY,FZ,GB,GU,GV,GW,GZ,HI,HJ,HK,HL,HM,HN,HO,HR,HS,IX,JA,JB,JC,JD,JE,JF,JG,JH,JI,JJ,JK,JL,JM,JN,JO,JP,JQ,JR,JS,JT,JU,JV,JW,JX,JY,KB,KG,KH,KN,KO,KS,LA,LD,LP,LT,LU,MK,MT,MU,MV,VE,VG,XP,XR,XZ,YK,YM,YN,YR,YS,YT,YU,YV,YW,YX,YY,ZE,ZF,ZG,ZH,ZI,ZJ,ZL,ZN,ZO,ZP,ZQ,ZR,ZT,ZU,ZW,ZY,ZZ,AAC,AAI,AAQ,AAX,AAY,ABE,ABF,ABH,ABI,ABL,ABM,ABN,ABP,ABQ,ABR,ABS,ABU,ACE,ACF,ACL,ACV,AFB,AFE,AFF,AFG,AFH,AFI,AFJ,AFK,AFL,AFP,AFT,AFU,AFV,AFW,AGL,AGM,ND,NE,NF,NG,NH,NI,NJ,NK,NL,NM,NN,NO,NP,NQ,NR,NS,NT,NU,NV,NW,NX,NY,NZ,OA,OB,OC,OD,OE,OF,OG,OH,OI,OJ,OK,OL,OM,ON,OO,OP,OQ,OR,OS,OT,OU,OV,OW,OX,OY,OZ,PA,PB,PC,PD,PE,PF,PG,PH,PI,PJ,PK,PL,PM,PN,PO,PP,PQ,PR,PS,PT,PU,PV,PW,PX,PY,PZ,QA,QB,QC,QD,QE,QF,QG,QH,QI,QJ,QK,QL,QM,QN,QO,QP,QQ,QR,QS,QT,QU,QV,QW,QX,QY,QZ,

ReferenceError: VD is not defined
    at /home/ubuntu/gopherjs/test.790433663:39:601154
    at Object.<anonymous> (/home/ubuntu/gopherjs/test.790433663:39:711135)
    at Object.<anonymous> (/home/ubuntu/gopherjs/test.790433663:51:4)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.runMain (module.js:575:10)
    at run (node.js:348:7)
    at startup (node.js:140:9)
    at node.js:463:3
FAIL	reflect	0.418s
```

I guess running the same test without the minimify option should give a more exploitable log. I should have a bit of time today to try that.

r-l-x:
Here is the error without minify:

```
$ gopherjs test reflect
/home/rl/go/src/github.com/gopherjs/gopherjs/test.166851752:36502
	T1$1.init("reflect_test", [{prop: "T1inner", name: "", exported: true, typ: T1inner, tag: ""}, {prop: "f", name: "f", exported: false, typ: $Int, tag: ""}]);
	                                                                            ^

ReferenceError: T1inner is not defined
    at /home/rl/go/src/github.com/gopherjs/gopherjs/test.166851752:36502:78
    at Object.<anonymous> (/home/rl/go/src/github.com/gopherjs/gopherjs/test.166851752:37213:3)
    at Object.<anonymous> (/home/rl/go/src/github.com/gopherjs/gopherjs/test.166851752:43577:4)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
FAIL	reflect	0.386s
```

r-l-x:
The only reference to a `T1inner` object I can find is in [TestAlignment](https://github.com/golang/go/blob/release-branch.go1.8/src/reflect/all_test.go#L833-L835), which [should be disabled](https://github.com/gopherjs/gopherjs/blob/master/compiler/natives/src/reflect/reflect_test.go#L12) in our case.

shurcooL:
> Yes, I'm not sure why, but in order to see the error with CircleCI, I need to fold and unfold the red part.

Thanks for the tip, that works for me.

> The only reference to a `T1inner` object I can find is in [TestAlignment](https://github.com/golang/go/blob/release-branch.go1.8/src/reflect/all_test.go#L833-L835), which [should be disabled](https://github.com/gopherjs/gopherjs/blob/master/compiler/natives/src/reflect/reflect_test.go#L12) in our case.

It is indeed disabled from _running_ that test, as you said. However, due to this change in this PR:

```Go
// Allow overridden function calls
// The standard library implementation of foo() becomes _gopherjs_overridden_foo()
if replacedDeclNames[funcName(d)] {
	d.Name = ast.NewIdent("_")
	d.Name.Name = "_gopherjs_overridden_" + d.Name.Name
}
```

It's no longer disabled from being _compiled_.

I suspect there may be a bug with deadcode elimination not handling this case well. /cc @neelance

However, I'm not able to reproduce it by doing something [like this](https://gopherjs.github.io/playground/#/46BjqVHugr). Perhaps we should factor out that change and investigate it separately. We should also confirm that it doesn't have a significant negative effect on generated file sizes.

shurcooL:
I've confirmed locally that's the case:

```bash
gopherjs $ gopherjs test --short reflect
PASS
ok  	reflect	3.742s
gopherjs $ open build/build.go
gopherjs $ git diff
diff --git a/build/build.go b/build/build.go
index 47a889e..260d242 100644
--- a/build/build.go
+++ b/build/build.go
@@ -308,7 +308,7 @@ func parseAndAugment(pkg *build.Package, isTest bool, fileSet *token.FileSet) ([
                        switch d := decl.(type) {
                        case *ast.FuncDecl:
                                if replacedDeclNames[funcName(d)] {
-                                       d.Name = ast.NewIdent("_")
+                                       d.Name.Name = "_gopherjs_overridden_" + d.Name.Name
                                }
                        case *ast.GenDecl:
                                switch d.Tok {
gopherjs $ go install -v
github.com/gopherjs/gopherjs/build
github.com/gopherjs/gopherjs
gopherjs $ gopherjs test --short reflect

/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/gopherjs/gopherjs/test.730446537:37375
})();
  ^
ReferenceError: T1inner is not defined
    at /Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/gopherjs/gopherjs/test.730446537:36664:78
    at Object.<anonymous> (/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/gopherjs/gopherjs/test.730446537:37375:3)
    at Object.<anonymous> (/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/gopherjs/gopherjs/test.730446537:43739:4)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:422:7)
    at startup (bootstrap_node.js:143:9)
    at bootstrap_node.js:537:3
FAIL	reflect	0.659s
gopherjs $ 
```

shurcooL:
Looking into it further, I think it may be a DCE/miscompilation issue introduced by that change itself. If so, that means it'll be more complicated to support overriding existing funcs than the 1 line change above.

r-l-x:
Hello again. FYI i'm currently working to fix this by specifying which of the overridden functions should be kept, using a syntax in the function comments, like:

```
//gopherjs:keep_overridden
func foo() {
  ...
}
```
This is similar to the go generate directives.

shurcooL:
About `//gopherjs:keep_overridden`, is it possible to defer that until a future PR?

A lot of code GopherJS already deals with preventing DCE by assigning to `_` (search for "DCE" in the codebase). I worry it might be suboptimal to try to introduce a second mechanism in this PR, because then the project will be inconsistent about how it prevents DCE. That can be considered and applied in a followup PR.

Of course, if it's really unavoidable to make things work here, then it's ok.

r-l-x:
Ok, I just pushed #798 for this. The branch associated to the current PR is rebased on the branch from #798, so the merge should be easy.

flimzy:
@r-l-x Are you still interested in bringing this PR up to date for inclusion?  #798 looks like it's close to being ready, which I understand would make this easier.

@nevkontakte were discussiong, and we think it might be wise to guard this functionality behind a build tag, at least initially, as well.  We might want to discuss including that in this PR.  But before we get too far down that road, we'd like to know if you're interested in pursuing this further.

Thanks.

PR #567, opened 2017-01-07
Update packages doc

TrueFurby:
I used emojis in `Supported` column for better overview of package status.


shurcooL:
Actually, here's an idea. The current version looks nice when rendered as markdown, but it's not so nice in raw markdown form.

What about replacing the emoji with unicode symbols. So:

```
:white_check_mark:
:ballot_box_with_check:
:x:
```

Becomes:

```
✅
☑️
❌
```

flimzy:
> becomes:

> ✅
> ☑️
> ❌

My 2¢, I like the colored icons better, even though it renders the MD less readable in raw form.

Maybe a compromise would be something like:

    runtime            | :ballot_box_with_check: partially |
    -- cgo             | :x:                     no        |

or

    runtime            | :ballot_box_with_check: | partially |
    -- cgo             | :x:                     | no        |

flimzy:
(Disregard my comment; the same icons show up as colorized when not in a blockquote)

shurcooL:
Here's a gist of what I mean:

https://gist.github.com/shurcooL/26998a997d87c3d775be9fc0a691b440

See its raw format (as it would show up in one's text editor):

https://gist.githubusercontent.com/shurcooL/26998a997d87c3d775be9fc0a691b440/raw/731215df11c27c066a6ccb595c6297bfe2af13d7/packages.md

TrueFurby:
I have changed `☑️ partially` to `❓ partially` since that shows as emoji and the one before did not.

shurcooL:
> since that shows as emoji and the one before did not.

Check the [update to my gist](https://gist.github.com/shurcooL/26998a997d87c3d775be9fc0a691b440/revisions). Apparently there are two ways to encode that unicode symbol, and one shows up prettier than the other:

image

TrueFurby:
Alright. Then is question which one is more relevant the the status **partially**.

This: ☑️ or this: ❓

shurcooL:
I seem to strongly prefer ☑️, especially because of how it fits in between the "✅ yes" and "❌ no". Using❓ implies uncertainty to me, when in fact we're certain that those packages are partially supported.

TrueFurby:
To be honest as I think about it, neither fits the status for me. But I see your logic. 

I've committed the change.

shurcooL:
I've improved alignment slightly in 932e8b0422626af2bfd2ed5874f5eb1cb6008960, hope that's okay (let me know if not).

TrueFurby:
Seem of as well just to the right side now. It's probably because of the width of those symbols. 

I don't really think it matters at all, most of the people will display it here in github anyway where it doesn't make a difference after Markdown and in nobody will care about the extra space in raw format anyway.

I think it's good enough now.

shurcooL:
Yep, LGTM. I can merge if @neelance is okay with this.

shurcooL:
@neelance, why not squash these 5 commits into one on master? They were just iterations during this PR.

neelance:
Honestly I didn't check the commits. Rebase+merge is my default. Do you want to squash and force push or keep it as is?

dominikh:
Do not force push.

shurcooL:
Force pushing to master is not an option, I wanted to raise it for next time.

I think for most logical PRs that do 1 thing, squashing into one commit should be the default, **unless** the author has carefully split it up into logical commits with nice commit messages that are more valuable to keep than squash.

neelance:
@dominikh Haha yeah, I agree with that. I just didn't put it that boldly. ;-)

Issue #569, opened 2017-01-09
Endless unwinding call stack for seachJsObject


influx6:
Today while attempting to integrate a test JS wrapper for the cache API released for webkit, a previous compile work but all new compiles consistently errored out due to a stack overflow error. Having established it is not the Go code in question I hoped any could help.

The problem arises when goperhjs attempts to retrieving the functions argument type for a promise, but instead cycles out till stack memory is depleted

WebCache Doc: https://developer.mozilla.org/en-US/docs/Web/API/Cache
WebCache API: https://github.com/gu-io/gu/blob/develop/shell/cache/webcache/webcache.go#L62-L66

Example Code:

```go
package main

import (
	"honnef.co/go/js/dom"

	"github.com/gu-io/gu/shell/cache/webcache"
)

func main() {
	doc := dom.GetWindow().Document().(dom.HTMLDocument)
	body := doc.QuerySelector("body").(*dom.HTMLBodyElement)

	webCache, err := webcache.New()
	if err != nil {
		body.AppendChild(doc.CreateTextNode(err.Error()))
		body.AppendChild(doc.CreateElement("br"))
	}

	cache, err := webCache.Open("debi.v1") // issue arises here.
	if err != nil {
		body.AppendChild(doc.CreateTextNode(err.Error() + "\n"))
		body.AppendChild(doc.CreateElement("br"))
		return
	}

	err = cache.Add("http://localhost:8080/github.com/gu-io/gu/shell/")
	if err != nil {
		body.AppendChild(doc.CreateTextNode(err.Error()))
		body.AppendChild(doc.CreateElement("br"))
	}

	res, err := cache.MatchPath("http://localhost:8080/github.com/gu-io/gu/shell/", webcache.MatchAttr{})
	if err != nil {
		body.AppendChild(doc.CreateTextNode(err.Error()))
		body.AppendChild(doc.CreateElement("br"))
		return
	}

	item := doc.CreateElement("div")
	item.SetInnerHTML(string(res.Body))

	body.AppendChild(item)
	body.AppendChild(doc.CreateElement("br"))
}
```

Error Received:
```
main.js:2058 Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at searchJsObject (main.js:2058)
    at searchJsObject (main.js:2070)
    at searchJsObject (main.js:2067)
    at searchJsObject (main.js:2070)
    at searchJsObject (main.js:2067)
    at searchJsObject (main.js:2070)
[...]

PR #572, opened 2017-01-12
build: Detect vendored js package and return error.

shurcooL:
This follows #538.

Vendoring `github.com/gopherjs/gopherjs/js` package is not supported, and can cause hard to understand runtime errors. Better catch it earlier and print a clear error message.

Helps #569.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -79,6 +79,11 @@ func importWithSrcDir(path string, srcDir string, mode build.ImportMode, install
 		return nil, err
 	}
 
+	// TODO: Resolve issue #415 and remove this temporary workaround.
+	if strings.HasSuffix(pkg.ImportPath, "/vendor/github.com/gopherjs/gopherjs/js") {
+		return nil, fmt.Errorf("vendoring github.com/gopherjs/gopherjs/js package is not supported, see https://github.com/gopherjs/gopherjs/issues/415")
+	}
+
 	switch path {
 	case "runtime":
 		pkg.GoFiles = []string{"error.go"}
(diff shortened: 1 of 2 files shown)


build/build.go
```diff
@@ -79,6 +79,11 @@ func importWithSrcDir(path string, srcDir string, mode build.ImportMode, install
 		return nil, err
 	}
 
+	// TODO: Resolve issue #415 and remove this temporary workaround.
+	if strings.HasSuffix(pkg.ImportPath, "/vendor/github.com/gopherjs/gopherjs/js") {
```
neelance: Does this condition work? `path == "github.com/gopherjs/gopherjs/js" && pkg.ImportPath != path`

shurcooL: I know `pkg.ImportPath` will always be an absolute import path, but I'm not sure about `path`. It might be relative. E.g., consider:

```
cd $GOPATH/src/github.com/gopherjs/gopherjs
gopherjs install ./js
```

Is there some issue with this method, or do you dislike it because of a string comparison?

I _think_ this is the canonical way to detect vendored packages, but I could be wrong.

neelance: Yes, `path` might be relative in the case you mentioned, but I don't think that case is relevant here. Still, on second thought I think your solution is also good. Let's merge.

shurcooL:
> Still, on second thought I think your solution is also good.

If you look at 42e645943c55c7a36ccd46f479b8ee038fe51b2e, this solution is effectively taken from our existing code to detect `js` package and vendored versions thereof in `IsJsPackage`.

PR #588, opened 2017-02-15
compiler/natives: add missing runtime/pprof.Profile methods

fjl:
I need this change to build "github.com/rcrowley/go-metrics" as a dependency of my program.


fjl:
Unrelated question: I also need "os/user", would you support adding a noop implementation to gopherjs? Right now packages that need "os/user" can't be compiled because it needs cgo.

shurcooL:
What are you doing that needs os/user? Is it an option to not use it via build tags?

neelance:
LGTM, thanks for contributing.

Issue #499, opened 2016-08-21
Support reflect.StructOf function (new to Go 1.7).


dmitshur:
[`reflect.StructOf` function](https://godoc.org/reflect#StructOf) has been added in Go 1.7 and GopherJS currently does not support it, as can be seen on the [compatibility table](https://github.com/gopherjs/gopherjs/blob/65dd5e201a11ceb651774a2617c879722d3cbe31/doc/packages.md) as of commit https://github.com/gopherjs/gopherjs/commit/65dd5e201a11ceb651774a2617c879722d3cbe31.

This issue is to track its support status. /cc @neelance It is not the highest priority issue at this time and may take more time to get around to.

flimzy:
This has now been fixed with https://github.com/gopherjs/gopherjs/pull/1015

PR #590, opened 2017-02-17
Make runnable Examples work

flimzy:
This PR works around the issues discussed in #529 by using a temp file, rather than `os.Pipe`, when executing examples.


flimzy:
A question for @neelance or @shurcooL :

This PR is failing with the following output:

    go test -v -race ./...
    ?   	github.com/gopherjs/gopherjs	[no test files]
    # github.com/gopherjs/gopherjs/compiler/natives/src/testing
    compiler/natives/src/testing/example.go:29: undefined: chatty
    compiler/natives/src/testing/example.go:75: undefined: chatty

Previously I had even more undefined symbols, as they are defined in the native package, and I only overwrote the symbols I wanted to change in the 'compiler/natives/src/testing/*.go' files.

Setting a variable here, and in the native package seems not to work, though--it seems GopherJS's overriding logic doesn't work on variables(??)

What's the best approach here?

Do I need to do something differently?  Is the test too strict?

flimzy:
I see now that all the other GopherJS-specific natives use the `js` build tag, which answers my question. So sorry for the unnecessary ping!

neelance:
Looks good to me. Please rebase on latest master, then we can merge.

@shurcooL Looks good to you as well?

flimzy:
Done @neelance

shurcooL:
@flimzy, those 2 commits where you changed code to avoid extra imports, was that because a test was failing when you did that? Just want to check.

flimzy:
@shurcooL Yes, exactly.

circle.yml
```diff
@@ -126,4 +124,8 @@ test:
       unicode
       unicode/utf16
       unicode/utf8
+    - >
+      gopherjs test --short --minify --run='^Test'
+      net/http/cookiejar
+      reflect
```
shurcooL: It [looks like](https://circleci.com/gh/gopherjs/gopherjs/1062) only two examples are failing:

- `ExampleNew`, because it tries to perform network operations that GopherJS doesn't support (i.e., it tries to create a new HTTP server with `httptest.NewServer`).
- `ExampleStructOf` is failing because `reflect.StructOf` is not supported (see #499).

In the case of `net/http/cookiejar`, `ExampleNew` is the only example it has.

However, for `reflect`, there are [other examples](https://godoc.org/reflect#pkg-examples). Do you know if they work?

I think it'd be nice to simply disable the failing/unsupported examples, the way we already for tests, instead of outright disabling _all_ examples for the packages that contain examples that don't work. That'd be cleaner, and hopefully some examples in reflect packages do pass (and might catch regressions in the future).

compiler/natives/src/testing/ioutil.go
```diff
@@ -0,0 +1,63 @@
...
+	return
+}
+
+func readFile(filename string) (string, error) {
+	buf := new(bytes.Buffer)
+	f, err := os.Open(filename)
+	if err != nil {
+		return "", err
+	}
+	io.Copy(buf, f)
```
shurcooL: Why not check errors from `io.Copy`?

How about this:

```Go
func readFile(filename string) (string, error) {
	f, err := os.Open(filename)
	if err != nil {
		return "", err
	}
	defer f.Close()
	var buf bytes.Buffer
	_, err = io.Copy(&buf, f)
	if err != nil {
		return "", err
	}
	return buf.String(), nil
}
```

It would also be more efficient to do a stat on the file to get its size, then allocate a buffer of exactly that size. I'm guessing you chose not to do that because most test output is pretty small, so it's not worth it. Is that the reason?

circle.yml
flimzy: I believe that's the only failure in the reflect package.  I was essentially being lazy about creating yet another special case to run the other examples, but I can do that.

compiler/natives/src/testing/example.go
```diff
@@ -20,23 +19,12 @@ func runExample(eg InternalExample) (ok bool) {
...
-	go func() {
-		var buf bytes.Buffer
-		_, err := io.Copy(&buf, r)
-		r.Close()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "testing: copying pipe: %v\n", err)
-			os.Exit(1)
-		}
-		outC <- buf.String()
-	}()
```
shurcooL: Just trying to understand this better. Do you know why the original Go code did this here in a goroutine, instead of simply doing it at the end (original line 51, `out := <-outC`)?

Is it so that writes into the write-end of pipe don't block? I don't know how `io.Pipe` works in that sense. Do you need to read from the read-end, otherwise writing-end blocks?

Ok, reading [its documentation](https://godoc.org/io#Pipe) confirms that's the case:

> Pipe creates a synchronous in-memory pipe. It can be used to connect code expecting an io.Reader with code expecting an io.Writer.
>
> Reads and Writes on the pipe are matched one to one except when multiple Reads are needed to consume a single Write. That is, each Write to the PipeWriter blocks until it has satisfied one or more Reads from the PipeReader that fully consume the written data. The data is copied directly from the Write to the corresponding Read (or Reads); there is no internal buffering.

circle.yml
shurcooL: Well, it would allow you to simplify circle.yml back to the original version. I think it's worth doing, if it's indeed possible. Thanks!

flimzy: Ah yes, that is a much better approach. I'll make that change.

circle.yml
```diff
@@ -126,4 +124,9 @@ test:
...
+    - gopherjs test --short --minify --run='^Example(MakeFunc|StructTag|StructTag.Lookup|TypeOf)'
```
shurcooL: This isn't the approach I meant. Is it possible to override and skip examples similar to how we skip failing/unsupported tests?

Look at https://github.com/gopherjs/gopherjs/commit/3303ee8edfe1a368bdc2c500279f2ba56bf7c3ec for example.

There's no `t.Skip` for examples, but just leaving the func body blank would probably work, right?

My goal is to simplify the command to run tests to just `gopherjs test --short --minify ...` instead of adding more special cases. That way, it's easier to run it locally, not just in CI.

flimzy: I realized after I committed the change. Stand by for an update.

compiler/natives/src/testing/example.go
```diff
@@ -48,11 +36,16 @@ func runExample(eg InternalExample) (ok bool) {
 		// Close pipe, restore stdout, get output.
 		w.Close()
 		os.Stdout = stdout
-		out := <-outC
+		out, e := ioutil.ReadFile(w.Name())
+		if e != nil {
+			fmt.Fprintf(os.Stderr, "testing: reading stdout file: %v\n", e)
+			os.Exit(1)
+		}
+		_ = os.Remove(w.Name())
```
shurcooL: If there's any error during reading the example stdout file, `_ = os.Remove(w.Name())` line will never run.

Would the following be better?

```Go
out, readFileError := ioutil.ReadFile(w.Name())
os.Remove(w.Name())
if readFileError != nil {
	fmt.Fprintf(os.Stderr, "testing: reading stdout file: %v\n", readFileError)
	os.Exit(1)
}
```

compiler/natives/src/net/http/cookiejar/example_test.go
```diff
@@ -0,0 +1,5 @@
+// +build js
+
+package cookiejar
+
+func ExampleNew() {}
```
shurcooL: It's very helpful (for later) to put a comment inside that explains why a test/example is disabled:

```Go
func ExampleNew() {
    // Example relies on httptest.NewServer, which GopherJS does not implement.
}
```

compiler/natives/src/reflect/example_test.go
```diff
@@ -0,0 +1,5 @@
+// +build js
+
+package reflect
+
+func ExampleStructOf() {}
```
shurcooL: Same here.

```Go
func ExampleStructOf() {
    // Blocked on https://github.com/gopherjs/gopherjs/issues/499.
}
```

compiler/natives/src/testing/ioutil.go
flimzy: What is the best way of allocating a `bytes.Buffer` of a specific size?

    b := make([]byte,size)
    buf := bytes.NewBuffer(b)

?

compiler/natives/src/net/http/cookiejar/example_test.go
```diff
@@ -0,0 +1,7 @@
+// +build js
+
+package cookiejar
```
shurcooL: I think this needs to be `package cookiejar_test`. The example is in an external test package.

See https://github.com/golang/go/blob/go1.8/src/net/http/cookiejar/example_test.go#L5.

compiler/natives/src/reflect/example_test.go
```diff
@@ -0,0 +1,8 @@
+// +build js
+
+package reflect
```
shurcooL: Same here, `package reflect_test`.

See https://github.com/golang/go/blob/go1.8/src/reflect/example_test.go#L5.

compiler/natives/src/net/http/cookiejar/example_test.go
```diff
@@ -0,0 +1,7 @@
...
+
+func ExampleNew() {
+  // network access not supported by GopherJS, and this test depends on httptest.NewServer
```
shurcooL: gofmt?

compiler/natives/src/testing/ioutil.go
shurcooL: Depends on if you want a byte slice, then:

```
b := make([]byte, 0, capacity)
```

Or if you want a `bytes.Buffer`, then:

```
buf := bytes.NewBuffer(make([]byte, 0, capacity))
```

See [`bytes.NewBuffer` documentation](https://godoc.org/bytes#NewBuffer):

> NewBuffer creates and initializes a new Buffer using buf as its initial contents. It is intended to prepare a Buffer to read existing data. **It can also be used to size the internal buffer for writing. To do that, buf should have the desired capacity but a length of zero.**

shurcooL:
@flimzy I've pushed some fixes to this PR, I hope that's okay with you. See last 4 commits and their commit messages for details.

flimzy:
Thank you @shurcooL.  I have squished a number of my commits now as well.

I believe the only outstanding issue you brought up is that of reading the file size to pre-allocate an appropriate buffer. Do you think that's still worth doing? The official ioutils package has some pretty elaborate logic about how it handles that (specifically in that it doesn't trust the reported file size to always be accurate). I suspect there are cases of networked filesystems where the size might be under-reported, especially immediately after having written the file.  If you think it's worth optimising that bit of code, I can continue to work on it.  But as you mentioned, with our small file sizes (the vast majority only a few dozen bytes, I suspect), it may not be worth it.

shurcooL:
No problem. Don't hesitate to squash my commits too, there's no reason not to. This PR will likely be merged as one squashed commit anyway.

As for the read file preallocation issue, I don't think it's worth it. It's fine as is.

At the most, you can do a stat to get the file size and use that to set initial buffer size. `bytes.Buffer` will still grow if needed, if the initial size was underestimated. But I doubt it's worth it. Add some quick `timeTaken := time.Since(started)` printf statements and see how long it takes to the read the file (and how much of a difference it makes).

Leaving it as is is fine by me. I commented because I wanted us to be aware and make a conscious decision, that's all.

circle.yml
```diff
@@ -18,7 +18,7 @@ test:
     - go tool vet *.go                                                   # Go package in root directory.
     - for d in */; do echo $d; done | grep -v tests/ | grep -v third_party/ | xargs go tool vet # All subdirectories except "tests", "third_party".
     - >
-      gopherjs test --short --minify --run='^Test'
+      gopherjs test --short --minify
```
shurcooL: This change is so clean and beautiful now. 👍 😻

shurcooL:
@neelance Do you want to look this over again, since there've been some changes based on my feedback? I think by now it has both mine and @flimzy's approval.

compiler/natives/src/testing/example.go
```diff
@@ -0,0 +1,70 @@
...
+	os.Stdout = w
+
+	start := time.Now()
+	ok = true
+
+	// Clean up in a deferred call so we can recover if the example panics.
+	defer func() {
+		dstr := fmtDuration(time.Now().Sub(start))
+
+		// Close pipe, restore stdout, get output.
```
shurcooL: Minor, we've changed the code to use file rather than pipe, but the comment still says pipe.

I can fix it.

neelance:
Looks good to me, feel free to merge.

flimzy:
Thanks, by the way, for your help in getting this PR polished, @shurcooL . I've been without my normal development machine the last week, which made responding to your comments in a timely and efficient manner more difficult than usual.

shurcooL:
No problem @flimzy, I'm happy if I can I do something useful and help make progress!

Thanks for the PR! It's merged now.

PR #595, opened 2017-02-23
Handle GOARCH env var similarly to cmd/go.

### LINKED ISSUE #594: env GOARCH=js gopherjs test fails
I created a simple CLI wrapper to switch between go and gopherjs development in LiteIDE:

https://github.com/bep/go-or-gopherjs

I thought I'd be [clever](https://github.com/bep/go-or-gopherjs/blob/master/main.go#L22) and use `GOARCH=js` for this, but that fails on my `macOS` computer.

```
▶ env GOARCH=js gopherjs test
../../../../../../../../usr/local/go/src/syscall/syscall_unix.go:196:38: undeclared name: _Socklen
../../../../../../../../usr/local/go/src/syscall/bpf_bsd.go:16:28: undeclared name: BpfInsn
../../../../../../../../usr/local/go/src/syscall/bpf_bsd.go:21:36: undeclared name: BpfInsn
../../../../../../../../usr/local/go/src/syscall/bpf_bsd.go:82:9: undeclared name: IFNAMSIZ
../../../../../../../../usr/local/go/src/syscall/bpf_bsd.go:108:27: undeclared name: Timeval
../../../../../../../../usr/local/go/src/syscall/bpf_bsd.go:118:32: undeclared name: Timeval
../../../../../../../../usr/local/go/src/syscall/bpf_bsd.go:127:25: undeclared name: BpfStat
../../../../../../../../usr/local/go/src/syscall/bpf_bsd.go:146:25: undeclared name: BpfInsn
../../../../../../../../usr/local/go/src/syscall/flock.go:13:30: undeclared name: SYS_FCNTL
../../../../../../../../usr/local/go/src/syscall/flock.go:16:42: undeclared name: Flock_t
../../../../../../../../usr/local/go/src/syscall/flock.go:16:42: too many errors
```

It is not a big deal for my use case, but I thought I should give you a heads up.

shurcooL:
Externally, GOARCH=js is the only supported mode for gopherjs compiler. Make it so that setting GOARCH=js explicitly does what one would expect, which is normal gopherjs behavior instead of a strange error.

If GOARCH is set to any value other than js, which are not supported, exit with an error (similar to cmd/go behavior).

Set "GOARCH" env var to correct value for gopherjs in tests/run.go.

Fixes #594.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -91,6 +91,10 @@ func main() {
 	cmdBuild.Flags().AddFlag(flagTags)
 	cmdBuild.Flags().AddFlag(flagLocalMap)
 	cmdBuild.Run = func(cmd *cobra.Command, args []string) {
+		if err := verifyAndUnsetGOARCH(); err != nil {
+			printError(err, options, nil)
+			os.Exit(2)
+		}
 		options.BuildTags = strings.Fields(*tags)
 		for {
 			s := gbuild.NewSession(options)
@@ -172,6 +176,10 @@ func main() {
 	cmdInstall.Flags().AddFlag(flagTags)
 	cmdInstall.Flags().AddFlag(flagLocalMap)
 	cmdInstall.Run = func(cmd *cobra.Command, args []string) {
+		if err := verifyAndUnsetGOARCH(); err != nil {
+			printError(err, options, nil)
+			os.Exit(2)
+		}
 		options.BuildTags = strings.Fields(*tags)
 		for {
 			s := gbuild.NewSession(options)
@@ -238,6 +246,10 @@ func main() {
 		Short: "display documentation for the requested, package, method or symbol",
 	}
 	cmdDoc.Run = func(cmd *cobra.Command, args []string) {
+		if err := verifyAndUnsetGOARCH(); err != nil {
+			printError(err, options, nil)
+			os.Exit(2)
+		}
 		exitCode := handleError(func() error {
 			goDoc := exec.Command("go", append([]string{"doc"}, args...)...)
 			goDoc.Stdout = os.Stdout
@@ -270,6 +282,10 @@ func main() {
 		Short: "compile and run Go program",
 	}
 	cmdRun.Run = func(cmd *cobra.Command, args []string) {
+		if err := verifyAndUnsetGOARCH(); err != nil {
+			printError(err, options, nil)
+			os.Exit(2)
+		}
 		os.Exit(handleError(func() error {
 			lastSourceArg := 0
 			for {
@@ -321,6 +337,10 @@ func main() {
 	cmdTest.Flags().AddFlag(flagTags)
 	cmdTest.Flags().AddFlag(flagLocalMap)
 	cmdTest.Run = func(cmd *cobra.Command, args []string) {
+		if err := verifyAndUnsetGOARCH(); err != nil {
+			printError(err, options, nil)
+			os.Exit(2)
+		}
 		options.BuildTags = strings.Fields(*tags)
 		os.Exit(handleError(func() error {
 			pkgs := make([]*gbuild.PackageData, len(args))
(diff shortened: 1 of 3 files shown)


tool.go
```diff
@@ -772,6 +796,25 @@ func printError(err error, options *gbuild.Options, browserErrors *bytes.Buffer)
...
+		return nil
+	}
+	if goarch != "js" {
+		return fmt.Errorf("gopherjs: unsupported GOOS/GOARCH pair %s/%s", build.Default.GOOS, goarch)
+	}
+	err := os.Unsetenv("GOARCH")
+	if err != nil {
+		return err
+	}
+	build.Default.GOARCH = runtime.GOARCH // Reset GOARCH to its default value, what it would've been had the GOARCH environment variable not overridden it.
```
neelance: Where do we depend on GOARCH not being `js`? Can we fix the behavior there instead of having to unset it here?

tool.go
```diff
@@ -91,6 +91,10 @@ func main() {
 	cmdBuild.Flags().AddFlag(flagTags)
 	cmdBuild.Flags().AddFlag(flagLocalMap)
 	cmdBuild.Run = func(cmd *cobra.Command, args []string) {
+		if err := verifyAndUnsetGOARCH(); err != nil {
```
neelance: Imho this should only be referenced once in the `compiler` package, not for every command.

tool.go
shurcooL: I'll investigate, thanks for the improvement idea.

I worry that some packages in std lib might list all known architectures explicitly, and `js` wouldn't be one of them, so we would select no .go files, unless we pick a well known architecture such as `amd64`. But if it's only a few such packages, we can get around by making them special cases.

I'll see if it works out.

tool.go
shurcooL: cmd/go does this as part of cmd/go, rather than compiler. But only for `gc`, it seems.

https://github.com/golang/go/blob/master/src/cmd/go/internal/work/build.go#L1098-L1101

I'll try and see it if works out.

tool.go
shurcooL: > I worry that some packages in std lib might list all known architectures explicitly, and `js` wouldn't be one of them, so we would select no .go files, unless we pick a well known architecture such as `amd64`. But if it's only a few such packages, we can get around by making them special cases.

It turns out we were already doing that, just suboptimally. See:

https://github.com/gopherjs/gopherjs/blob/b9bcb1da229a59cc1e1d168401662cb6450aae08/build/build.go#L70

Using `build.Context.GOARCH` there is a mistake, because it's overridden by `GOARCH` env var. Instead, we should use `runtime.GOARCH` there.

> I'll see if it works out well.

Looking great so far.

tool.go
shurcooL: I've spent some time looking through https://godoc.org/github.com/gopherjs/gopherjs/build, https://godoc.org/github.com/gopherjs/gopherjs/compiler, https://godoc.org/github.com/gopherjs/gopherjs and the source, thinking about a better place where this could be moved.

I didn't find a great place.

There are multiple different entry points, depending on whether you're building import paths or specifying .go files on the command line, etc. Finding/catching them all is tricky and hard to be confident in.

`github.com/gopherjs/gopherjs/build.Session.BuildPackage` is pretty central and gets called by most entry points, but it seems a little too late. In some cases, a bunch of work loading and importing packages is done before `BuildPackage` is called.

`github.com/gopherjs/gopherjs/build.NewSession` seems more fitting, it's also central, and happens much earlier. But it doesn't return an error right now. We could change it to do that.

But it still happens inside a for loop in tool.go because of watcher support.

Also, nothing compilation-related happens inside `gopherjs doc` command. But I don't want it to do anything with non-`js` architectures (even if it happens to work, supporting that is out of scope).

In the end, checking that GOARCH is not some crazy value makes most sense to me inside `tool.go`, first thing. I don't want anything to happen if GOARCH=mips64 or something else. GopherJS only supports compiling GOARCH=js architecture, and there are no short term plans to change that.

Unless you have some other suggestion that I've missed, I'd prefer we keep this check inside tool, that happens right away, rather than later.

shurcooL:
I believe I've addressed all feedback (see https://github.com/gopherjs/gopherjs/pull/595#discussion_r102806426), PTAL.

shurcooL:
Squashed into 2 logical commits and merged as 62bca28b2835514b22712edb8104a624030a08c0 and 69499efe08d29030cc47842bf2a214ca2de4edc5.

PR #596, opened 2017-02-25
Minor refactor (no behavior change) to simplify and clean up.

shurcooL:
This PR is a minor refactor (no behavior change) to simplify and clean up some of the code I've recently seen/looked over, and found opportunities to improve. Some of the changes are ones I wanted to make while working on #595, and some others I just did on the fly when making this PR.

Please review commit-by-commit. Each commit contains a readable bite-sized changed, and the commit message describes the motivation and rationale. Please leave comments if you disagree with something or have better suggestions.

There should not be any provable behavior changes. If there are, that's a bug.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -168,8 +168,11 @@ func WriteProgramCode(pkgs []*Archive, w *SourceMapFilter) error {
 		}
 	}
 
-	_, err := w.Write([]byte("$synthesizeMethods();\nvar $mainPkg = $packages[\"" + string(mainPkg.ImportPath) + "\"];\n$packages[\"runtime\"].$init();\n$go($mainPkg.$init, []);\n$flushConsole();\n\n}).call(this);\n"))
-	return err
+	if _, err := w.Write([]byte("$synthesizeMethods();\nvar $mainPkg = $packages[\"" + string(mainPkg.ImportPath) + "\"];\n$packages[\"runtime\"].$init();\n$go($mainPkg.$init, []);\n$flushConsole();\n\n}).call(this);\n")); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, minify bool, w *SourceMapFilter) error {
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -168,11 +168,8 @@ func WriteProgramCode(pkgs []*Archive, w *SourceMapFilter) error {
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


compiler/compiler.go
```diff
@@ -168,11 +168,8 @@ func WriteProgramCode(pkgs []*Archive, w *SourceMapFilter) error {
 		}
 	}
 
-	if _, err := w.Write([]byte("$synthesizeMethods();\nvar $mainPkg = $packages[\"" + string(mainPkg.ImportPath) + "\"];\n$packages[\"runtime\"].$init();\n$go($mainPkg.$init, []);\n$flushConsole();\n\n}).call(this);\n")); err != nil {
-		return err
-	}
-
-	return nil
+	_, err := w.Write([]byte("$synthesizeMethods();\nvar $mainPkg = $packages[\"" + string(mainPkg.ImportPath) + "\"];\n$packages[\"runtime\"].$init();\n$go($mainPkg.$init, []);\n$flushConsole();\n\n}).call(this);\n"))
```
neelance: I don't like this simplification, I often even change my code to not do this. I prefer having the primary flow with `return nil` (all okay) at the toplevel and have `return err` only in branches for error handling. This also allows cleaner diffs when adding new stuff between this statement and the end of the function. Same goes for other such changes in this commit.

shurcooL: Up until about a year ago, I used to feel the same way. But I saw that the Go standard library did it, and I decided to try it after all. Then this simplification grew on me.

Now, when I read Go code, seeing `return err` at the end of a function is absolutely equivalent and interchangeable with `if err != nil { return err }; return nil`, except it avoids being unnecessarily longer.

I've also started to not mind diffs. Unless the code happens to be very very WIP, diffs are actually pretty rare. So I'd rather look at the shorter equivalent version for a year before a change, and when a chance is neccessary, the diff is not at all that much worse (i.e., I find it very readable too).

In some places, I still prefer to do `err := DoSomething(); return err` instead of just `return DoSomething()`, when I want to emphasize that doing the thing is more important than returning its error value. But even that is almost interchangeable for me now, and that's why I wrote `return s.BuildFiles(args, pkgObj, currentDirectory)` below, instead of `err = s.BuildFiles(args, pkgObj, currentDirectory); return err`.

But okay, I will back out this change for now since you prefer the `if err != nil { return err }; return nil` version and we can always change it later if you change your mind.

(/cc @dominikh FYI, this is related to S1013.)

neelance: I'm also open to changing my mind, that's why I suggested to talk about it.

shurcooL: I find these kind of things are best to do on your own time, in the background, and not let other code improvements be blocked on talking about it. Consider the points I made and see how you feel later on.

My main argument for why the simpler form is better is because it's equivalent, shorter, and done more commonly in the Go project (hence, more idiomatic).

Issue #597, opened 2017-02-26
0-length byte arrays not properly externalized? (Empty HTTP body causes errors with xhr2 polyfill.)


flimzy:
I'm trying to get an XHR polyfill to work for testing in node.js, and discovered that with the [xhr2](https://www.npmjs.com/package/xhr2) library, an empty body on a PUT request generates an error.  To reproduce:

`main.go`:

	package main

	import (
		"fmt"
		"net/http"
	)

	func main() {
		req, _ := http.NewRequest("PUT", "http://bogusurl.com/", nil)
		client := &http.Client{}
		_, err := client.Do(req)
		if err != nil {
			fmt.Printf("Error: %s\n", err)
		}
	}

`main.inc.js`:

    try {
        global.XMLHttpRequest = require('xhr2');
    } catch(e) {}

Then run these commands:

    npm install xhr2
    gopherjs run main.go main.inc.js

And observe the following output:


    /home/jonhall/go/src/github.com/flimzy/jstest/node_modules/xhr2/lib/xhr2.js:844                                                                                                                                                  
    throw new Error("Unsupported send() data " + data);                                                                                                                                                                      
    ^                                                                                                                                                                                                                  
    Error: Unsupported send() data                                                                                                                                                                                                   
    at XMLHttpRequestUpload._setData (/home/jonhall/go/src/github.com/flimzy/jstest/node_modules/xhr2/lib/xhr2.js:844:15)                                                                                                        
    at XMLHttpRequest._sendHttp (/home/jonhall/go/src/github.com/flimzy/jstest/node_modules/xhr2/lib/xhr2.js:375:19)                                                                                                             
    at XMLHttpRequest.send (/home/jonhall/go/src/github.com/flimzy/jstest/node_modules/xhr2/lib/xhr2.js:203:16)
    at $packages.net/http.XHRTransport.ptr.RoundTrip (/home/jonhall/go/src/github.com/flimzy/jstest/http.go:99:3)
    at send (/net/http/client.go:249:3)
    at $packages.net/http.Client.ptr.send (/net/http/client.go:173:3)
    at $packages.net/http.Client.ptr.Do (/net/http/client.go:595:7)
    at main (/github.com/flimzy/jstest/main.go:11:3)
    at $init (/home/jonhall/go/src/github.com/flimzy/jstest/main.go.863915836:99004:9)
    at $goroutine (/home/jonhall/go/src/github.com/flimzy/jstest/main.go.863915836:1474:19)
[...]

PR #599, opened 2017-02-26
Don't ignore array type for special JS array types

flimzy:
This PR should fix the problem mentioned in #597 .

Using this sample program:

    package main

    import "honnef.co/go/js/console"

    func main() {
    	x := []byte{}
    	var y []byte
    	z := []byte{0x01}
    	console.Log(x)
    	console.Log(y)
    	console.Log(z)
    }

Before:

    []
    []
    Uint8Array [ 1 ]

After:

    Uint8Array [  ]
    Uint8Array [  ]
    Uint8Array [ 1 ]

### DIFF
--- a/compiler/prelude/prelude.go
+++ b/compiler/prelude/prelude.go
@@ -119,9 +119,6 @@ var $substring = function(str, low, high) {
 };
 
 var $sliceToArray = function(slice) {
-  if (slice.$length === 0) {
-    return new slice.$array.constructor;
-  }
   if (slice.$array.constructor !== Array) {
     return slice.$array.subarray(slice.$offset, slice.$offset + slice.$length);
   }
--- a/compiler/prelude/prelude.go
+++ b/compiler/prelude/prelude.go
@@ -120,7 +120,7 @@ var $substring = function(str, low, high) {
 
 var $sliceToArray = function(slice) {
   if (slice.$length === 0) {
-    return [];
+    return new slice.$array.constructor;
   }
   if (slice.$array.constructor !== Array) {
     return slice.$array.subarray(slice.$offset, slice.$offset + slice.$length);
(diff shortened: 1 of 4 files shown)


js/js_test.go
```diff
@@ -558,3 +558,17 @@ func TestSurrogatePairs(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func instanceof(x interface{}, y *js.Object) bool {
+	return js.Global.Call("$instanceof", x, y).Bool()
+}
```
flimzy: There may be a better way to do this test; I'm not sure what it would be.

compiler/prelude/prelude.go
```diff
@@ -120,7 +120,7 @@ var $substring = function(str, low, high) {
 
 var $sliceToArray = function(slice) {
   if (slice.$length === 0) {
-    return [];
+    return new slice.$array.constructor;
```
neelance: Can we simply remove this special case? `subarray` and `slice` with a length of `0` should give the same result.

js/js_test.go
neelance: This is fine I think.

compiler/prelude/prelude.go
flimzy: I tested with that entirely removed at first, and it worked just as well. I wondered if the special case served as a meaningful optimization, so left it in there.

If you're in favor of removing it, that is of course fine with me. I can make that change.

js/js_test.go
neelance: I just realized that you added a new file for this. I'd prefer the `eval` method of line 15 for this. Also, could you maybe try to add a function to `dummys` that simply returns the parameter (thus externalizing the value) and then try `.Get("constructor") != uint8Array`?

flimzy: A clever solution. Done.

neelance:
Cool. Thanks the updates and for contributing in general. Merging.

PR #600, opened 2017-02-26
Small optimization/improvement for XHR implementation

flimzy:
While sending an empty body is probably compliant with the spec, it's unnecessary, and it causes at least one known implementation (the nodejs xhr2 polyfill library) to output a bunch of extraneous warnings such as:

    Discarding entity body for GET requests
    Discarding entity body for HEAD requests
    Discarding entity body for HEAD requests
    Discarding entity body for GET requests
    Discarding entity body for GET requests

This is arguably a bug/misfeature of that library, but fixing it here makes sense to me, as it may be considered a small optimization, regardless of the XHR implementation.


neelance:
What difference does this make?

flimzy:
I should have put the explanation in the PR, not just the commit message. (I've updated the description on the PR now).

But TL;DR; a null value is distinct from an empty array in JS, and at least one implementation of XMLHttpRequest thinks its an error to have a body, even if it's 0 bytes long, on GET and HEAD requests, and warns rather annoyingly about it.

neelance:
Hmm okay, yeah `GET` and `HEAD` don't have a body. All right, merging. Thx.

PR #605, opened 2017-03-14
Revert "Handle GOARCH env var similarly to cmd/go."

### LINKED ISSUE #601: gopherjs can no longer be used from go generate
Setup:
$ cat generate.go
//go:generate gopherjs build
//(other go:generate stuff using the .js output)

Expected behaviour:
Running go generate runs gopherjs, which builds the .js and .js.map as if running gopherjs build directly.

Actual output:
gopherjs: unsupported GOOS/GOARCH pair darwin/amd64

go generate sets $GOOS and $GOARCH (https://golang.org/src/cmd/go/generate.go).
go generate has no way of overriding $GOOS or $GOARCH from the generate directive: //go:generate GOARCH=js gopherjs build just makes go generate try to exec "GOARCH=js".

shurcooL:
This reverts commit 62bca28b2835514b22712edb8104a624030a08c0.

We've originally discussed two possible fixes in issue #594:

> The gopherjs tool does not use the GOARCH environment variable
> in any way. Internally, GopherJS always uses js as GOARCH value
> when compiling user code, but it does other internal stuff when
> compiling standard library.
>
> 1. One fix is to make the gopherjs tool really ignore GOARCH env
> var. In other words, setting it to any value should have no effect;
> it shouldn't cause this failure.
>
> 2. We could also consider making it use the value of GOARCH, and
> error out if it's set to anything other than js (which is the
> default and the only supported value). Since gopherjs can't compile
> for any other architectures.

Upon further consideration, after seeing issue #601 and gopherjs/gopherjs.github.io#66,
we've decided (see discussion in https://github.com/gopherjs/gopherjs.github.io/pull/66#issuecomment-286519322)
to revisit this issue and change gopherjs to completely ignore GOARCH
environment variable, instead of requiring it to be unset or equal to js.

Closes #601.
Updates #594.

### DIFF
--- a/tests/run.go
+++ b/tests/run.go
@@ -620,9 +620,7 @@ func (t *test) run() {
 		os.Setenv("GOOS", runtime.GOOS)
 	}
 	if os.Getenv("GOARCH") == "" {
-		//os.Setenv("GOARCH", runtime.GOARCH)
-		// GOPHERJS.
-		os.Setenv("GOARCH", "js") // We're running this script natively, but the tests are executed with js architecture.
+		os.Setenv("GOARCH", runtime.GOARCH)
 	}
 
 	useTmp := true
--- a/tool.go
+++ b/tool.go
@@ -91,10 +91,6 @@ func main() {
 	cmdBuild.Flags().AddFlag(flagTags)
 	cmdBuild.Flags().AddFlag(flagLocalMap)
 	cmdBuild.Run = func(cmd *cobra.Command, args []string) {
-		if err := verifyGOARCH(); err != nil {
-			printError(err, options, nil)
-			os.Exit(2)
-		}
 		options.BuildTags = strings.Fields(*tags)
 		for {
 			s := gbuild.NewSession(options)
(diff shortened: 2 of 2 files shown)


tests/run.go
```diff
@@ -620,9 +620,7 @@ func (t *test) run() {
 		os.Setenv("GOOS", runtime.GOOS)
 	}
 	if os.Getenv("GOARCH") == "" {
-		//os.Setenv("GOARCH", runtime.GOARCH)
-		// GOPHERJS.
-		os.Setenv("GOARCH", "js") // We're running this script natively, but the tests are executed with js architecture.
+		os.Setenv("GOARCH", runtime.GOARCH)
```
neelance: Why do we still need this?

shurcooL: On the contrary, now that "GOARCH must equal js" is no longer enforced, we _no longer_ need this, which is why this change is being reverted.

shurcooL: The commit message of 62bca28b2835514b22712edb8104a624030a08c0 described this change as:

> Set "GOARCH" env var to correct value for gopherjs in tests/run.go.

Doing that is no longer needed, since GopherJS now ignores GOARCH. I'd rather revert it and keep tests/run.go more similar to upstream, rather than continuing to deviate it unnecessarily.

neelance: Ah I see, this is from upstream.

shurcooL: Yes, tests/run.go is a copy of https://github.com/golang/go/blob/go1.8/test/run.go, with modifications to support GopherJS. Sorry I didn't make that more clear.

PR #607, opened 2017-03-15
Make flags local to the commands that actually support them.

### LINKED ISSUE #606: gopherjs serve -w makes no sense and should not be possible.
`-w` flag has some meaning for `build` and `install` commands, but it's not needed for `serve`. We should make it error if used with `serve` to avoid confusion. **Edit:** Better yet, simply remove the flag from `serve` command.

I've seen this before in some reported issue, and I saw it again at https://github.com/myitcv/gopherjs/pull/20/files#r106279586. /cc @myitcv

shurcooL:
This PR consists of 2 commits.

First one touches the `-w` (aka `--watch`) flag only, fixes #606 and modifies the behavior slightly (the `-w` flag is removed from `gopherjs get` command, kept in `gopherjs build` and `gopherjs install` only).

Second commit refactors all other flags to make them local to commands that actually use them, instead of being global flags. That improves the output of `gopherjs --help` and `gopherjs <command> --help`, because it more accurately maps to the reality of flags and how they're used. Flags that are unused by a certain command do not show up in that command's `--help` output.

### Example --help Output

```
$ gopherjs --help
GopherJS is a tool for compiling Go source code to JavaScript.

Usage:
  gopherjs [command]

Available Commands:
  build       compile packages and dependencies
  doc         display documentation for the requested, package, method or symbol
  get         download and install packages and dependencies
  help        Help about any command
  install     compile and install packages and dependencies
  run         compile and run Go program
  serve       compile on-the-fly and serve
  test        test packages
  version     print GopherJS compiler version

Use "gopherjs [command] --help" for more information about a command.
```

```
$ gopherjs build --help
compile packages and dependencies

Usage:
  gopherjs build [packages] [flags]

Flags:
      --color           colored output (default true)
      --localmap        use local paths for sourcemap
  -m, --minify          minify generated code
  -o, --output string   output file
  -q, --quiet           suppress non-fatal warnings
      --tags string     a list of build tags to consider satisfied during the build
  -v, --verbose         print the names of packages as they are compiled
  -w, --watch           watch for changes to the source files
```

```
$ gopherjs serve --help
compile on-the-fly and serve

Usage:
  gopherjs serve [root] [flags]

Flags:
      --color         colored output (default true)
      --http string   HTTP bind address to serve (default ":8080")
      --localmap      use local paths for sourcemap
  -m, --minify        minify generated code
  -q, --quiet         suppress non-fatal warnings
      --tags string   a list of build tags to consider satisfied during the build
  -v, --verbose       print the names of packages as they are compiled
```

```
$ gopherjs doc --help
display documentation for the requested, package, method or symbol

Usage:
  gopherjs doc [arguments] [flags]
```


shurcooL:
Doing this refactored surfaced the fact that none of the flags were added locally to the `gopherjs run` command. This is contrary to all other commands that had the flags added locally.

It worked in the past because `gopherjs run` simply used the global flags. It should continue to work after b8afab4bc2b86e4d755eb15a9717ffee58d21e9c, because the flags are now explicitly added to that command.

shurcooL:
@neelance Do you have a chance to review this?

tool.go
```diff
@@ -266,6 +258,13 @@ func main() {
 		Use:   "run [gofiles...] [arguments...]",
 		Short: "compile and run Go program",
 	}
+	cmdRun.Flags().AddFlag(flagVerbose)
```
neelance: This seems weird, creating this flag on `cmdBuild` and then adding it to `cmdRun`. I think we're approaching this wrong. What we probably should do instead is not use `Lookup` at all, but https://godoc.org/github.com/spf13/pflag#NewFlagSet, define flags on it and then https://godoc.org/github.com/spf13/pflag#FlagSet.AddFlagSet to add it to a command.

shurcooL: Ideally, what I want to do is create all the flags, then add them to the commands that need/use them.

However, I don't see an API to create a standalone flag, only to create one within a flag set, then look it up.

The only change that I see being possible is to create all flags within a dummy flag set, then look them up, and add to individual commands.

I don't see how https://godoc.org/github.com/spf13/pflag#FlagSet.AddFlagSet can be useful. It adds an entire flag set, but most commands need a variety of specific flags added.

shurcooL: > The only change that I see being possible is to create all flags within a dummy flag set, then look them up, and add to individual commands.

I've done that in fe2b2c69b8d23430b4a6c8240ad44aa0b15fddc7, PTAL.

Let me know if you have ideas for how to improve it further. I don't see any other ways.

neelance: Create multiple flag sets and add them. Flag sets consisting of only a single flag are totally fine. That way you can avoid `Lookup` completely.

shurcooL:
> What do you think about grouping the flags a bit? For example `minify`, `color`, `tags` and `localmap`are always added together.

My goal with this PR is to resolve #606. I don't have strong opinions on this, and don't mind changing it.

I'm not sure if you meant to group multiple flags into a single flag set, or just group them visually by adding a blank line separator?

I'll done a little visual grouping. Is that okay? Feel free to follow up on this to arrange them in a way you think is better.

neelance:
I would combine them into a single flag set, so the code is not repeated that often. Also this shows that the fields are related, so when adding a new flag it is less likely to forget to add it to some command.

shurcooL:
Ok, I'll group `minify`, `color`, `tags` and `localmap` flags into one set, since it's a noop refactor. But I find it a little strange that --color flag is in the same category as --tags or --minify.

shurcooL:
Done in 1c6e08e, PTAL.

neelance:
`minifyColorTagsLocalMapFlags` Wow, that's a 🐍  name. When you add a new flag now, then you have to change all occurrences. What about `commonTags`?

shurcooL:
> What about `commonTags`?

What does "tags" refer to? Did you mean `commonFlags`?

I can do that, but it's kinda misleading. Not all commands get those flags, so they're not very common. They're more common to commands that are related to compiling. So how about `compilerFlags` instead?

neelance:
Yeah flags, not tags. ;-) `compilerFlags` sounds good.

shurcooL:
PTAL.

neelance:
lgtm

shurcooL:
Merged. Thanks for reviewing!

Issue #580, opened 2017-01-28
Using `btoa` and `atob` for faster Base64 encoding and decoding


flimzy:
Modern browsers support the [atob](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/atob) and [btoa](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa) functions for Base64 encoding and decoding respectively.

I just threw together [some benchmarks](https://gopherjs.github.io/playground/#/S0y0imUzKg) that demonstrate this could be an area for improved performance with GopherJS, which show that the native JS functions are consistently 30-40% faster for encoding, and 130-1050% faster for decoding.  Test results from Firefox:

```
BenchmarkStandardEnc10b:   300000	      6350 ns/op 
BenchmarkStandardEnc1kb:    30000	     54466 ns/op 
BenchmarkStandardEnc1mb:       30	  46933333 ns/op 
BenchmarkJSEnc10b      :   300000	      4610 ns/op 
BenchmarkJSEnc1kb      :    30000	     40933 ns/op 
BenchmarkJSEnc1mb      :       50	  33480000 ns/op 
BenchmarkStandardDec10b:   200000	      7105 ns/op 
BenchmarkStandardDec1kb:     3000	    421333 ns/op 
BenchmarkStandardDec1mb:        3	 420333333 ns/op 
BenchmarkJSDec10b      :   500000	      3108 ns/op 
BenchmarkJSDec1kb      :    30000	     38400 ns/op 
BenchmarkJSDec1mb      :       30	  40000000 ns/op 
```

Naturally, the native JS functions only work for standard encoding, and the Go standard library supports arbitrary encodings, as well as optional padding.  Therefore, to maintain compatibility, the JS native functions could only be used for the standard encoding, and standard padding, with everything else falling back to Go's implementation. But that should still cover the vast majority of use cases.

Is this worth perusing?

shurcooL:
This is a cool finding. In general, it's pretty awesome when there are fast low-level implementations available in the browser that map 1:1 to something the Go standard library exposes.

> Is this worth perusing?

My personal recommended answer would be... It's worth perusing if you have a need yourself. I.e., if doing this would optimize some code you care about being faster, and you're bottlenecked on base64 encoding/decoding.

[...]

PR #615, opened 2017-03-24
Improve []byte to string conversion performance on NodeJS

Mikayex:
Using `Buffer.from(array).toString("binary")` gives a 13x performance boost when doing `[]byte` to `string` conversion.

`Buffer` is only supported on NodeJS AFAIK (since v5.10.0). If `Buffer` is not supported it falls back to default implementation.

### Benchmark
The benchmark function was:
```go
var str string

func BenchmarkBytesToString(b *testing.B) {
	x := []byte("abcdefghijklmnopqrstuvwxyz")
	for i := 0; i < 15; i++ {
		x = append(x, x...)
	}
	b.SetBytes(int64(len(x)))
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		str = string(x)
	}
}
```

```
benchmark                  old ns/op     new ns/op     delta
BenchmarkBytesToString     35366666      2734000       -92.27%

benchmark                  old MB/s     new MB/s     speedup
BenchmarkBytesToString     24.09        311.62       12.94x
```

### DIFF
--- a/compiler/prelude/prelude.go
+++ b/compiler/prelude/prelude.go
@@ -209,6 +209,11 @@ var $bytesToString = function(slice) {
   if (slice.$length === 0) {
     return "";
   }
+
+  if ($global.Buffer !== "undefined") {
+    return Buffer.from(slice.$array).toString("binary", slice.$offset, slice.$offset + slice.$length);
+  }
+
   var str = "";
   for (var i = 0; i < slice.$length; i += 10000) {
     str += String.fromCharCode.apply(undefined, slice.$array.subarray(slice.$offset + i, slice.$offset + Math.min(slice.$length, i + 10000)));


compiler/prelude/prelude.go
```diff
@@ -209,6 +209,11 @@ var $bytesToString = function(slice) {
   if (slice.$length === 0) {
     return "";
   }
+
+  if ($global.Buffer !== "undefined") {
+    return Buffer.from(slice.$array).toString("binary", slice.$offset, slice.$offset + slice.$length);
+  }
+
   var str = "";
   for (var i = 0; i < slice.$length; i += 10000) {
```
myitcv: @neelance @shurcooL - what's special about the value `10,000`?

I note the Google Closure compiler seems to have a rather specific `CHUNK_SIZE` value... no idea where they got the value from however (iterative solution based on some performance tests perhaps?):

https://github.com/google/closure-library/blob/e877b1eac410c0d842bcda118689759512e0e26f/closure/goog/crypt/crypt.js#L53-L71

flimzy:
I noticed the same behavior while working on #580, but didn't consider it worth creating a PR for three reasons:

1. Node.js is supported primarily for testing, and testing code that's different than that run in the browser seems like a bad idea. Perhaps there's a use case for targeting Node.js as a primary target though? 
2. I didn't want to increase build sizes for an optimization that would, by the previous definition, help 0% of the target environments. 
3. Buffer polyfills exist, and I wouldn't want to hurt browser performance in a case where a Buffer polyfill was loaded out of necessity (to run a node-centric module). 

Just my thoughts.

theclapp:
> Node.js is supported primarily for testing

FWIW, I didn't realize this.  I figured GopherJS was for anybody that preferred Go over JS but had to run on a JS-only platform.  Just my own $0.02.

flimzy:
Quoting from the top of the README:

> GopherJS compiles Go code (golang.org) to pure JavaScript code. Its main purpose is to give you the opportunity to write front-end code in Go which will still run in all browsers.

Now I don't think that has to exclude running Go code in Node.js, but a number of fundamental design decisions do seem to preclude running meaningful production code in Node.js. In particular, the lack of true syscall support in Node.js, and limited emulation of syscalls, would be blockers for all but the most trivial use cases, I suspect.

But that's not my call, so I'm only offering my observations for perspective and consideration.

neelance:
Yes, I currently don't consider Node.js a primary target. I don't see any scenario where you would need it. Just run plain Go instead.

Mikayex:
About the code size increase, it is only 127 bytes minified and 32 bytes minified/gzipped.

Anyway, I understand your points and close this PR.
I'll try to find other optimizations, supported in browsers this time.

Issue #617, opened 2017-04-05
Bug with *js.Object-special types' zero values


myitcv:
This is really picking up conversation from #236, to register an issue for a specific problem, namely ([playground](https://gopherjs.github.io/playground/#/EzBPjTpGjn)):

```go
package main

import (
	"fmt"

	"github.com/gopherjs/gopherjs/js"
)

type S struct {
	*js.Object

	A string `js:"A"`
}

func main() {
	o := js.Global.Get("Object").New()
	s := S{Object: o}

	var str string

	str = s.A

	exp := ""

	fmt.Printf("%q == %q: %v", str, exp, str == exp)
	js.Global.Get("console").Call("log", "\"%s\" == \"%s\": %s", str, exp, str == exp)
}
```

The output of this is:

```
"undefined" == "": false
```

in both the browser console and the onscreen-Playground output.

To my mind this is a bug because the compiler generated code is failing to properly internalise an external value which we have defined to be of type `string`. That internalisation should handle the Javascript `undefined` value when we "read" the field value and convert it to the GopherJS `""` value. The output should be:

```
"" == "": true
```

Similarly (and I note the discussion in #504) if the type of `A` were instead `*string` then undefined would be converted to `nil`. 

And in any situation where the conversion is not well defined (i.e. if the underlying field had a `number` value) then we panic.

Furthermore the same conversion concept should apply for fields of all types on `*js.Object`-special types.

shurcooL:
Thanks for reporting this and providing concrete details, @myitcv.

I'll share some quick thoughts that come to me immediately after reading this, but I'll need more time to think about it further and perhaps I'll change my mind.

> To my mind this is a bug because the compiler generated code is failing to properly internalise an external value which we have defined to be of type string.

In my mind, it's not so clear whether it's a bug or not. It may, or may not be. It really depends on the following. Given this struct:

```Go
type S struct {
	*js.Object

	A string `js:"A"`
}
```

I know that <code>A string \`js:"A"\`</code> is really just a shortcut to accessing the JavaScript object's A property.

And when you do:

```Go
o := js.Global.Get("Object").New()
s := S{Object: o}
```

Which is translated to something like `var s = new Object();` in JavaScript.

It's very expected to me that `s.A` (JavaScript syntax) will be `undefined`.

The question is whether GopherJS should do a lightweight mapping/rewriting that is JavaScript-like, which it seems it currently does, or if it should do something more/different that is more Go-like (as you're expecting).

[...]

PR #618, opened 2017-04-06
[WIP] js: tighten up internalize and externalize

myitcv:
**!! DO NOT REVIEW !!**

**NEEDS REBASE AFTER #687 IS MERGED**

### Status

Step 1 and 2 (see below) review.

### Description

A work-in-progress PR to complement the discussion in #617 

Tighten up `$internalize` and `$externalize` such that:

* zero val translation of `undefined` and `null` is well defined
* the domain of valid values for a given target type is well defined
* runtime errors are thrown for invalid values

See impact of this PR on `honnef.co/go/js/dom` in [this comparison](https://github.com/dominikh/go-js-dom/compare/master...myitcv:internalize_zero_val?expand=1)

### Documentation

Proposed human-readable definitions of `$internalize` and `$externalize` can be seen [here](https://github.com/myitcv/gopherjs-compiler/blob/internalize_zero_val/js/js.go).

### Scope of `$internalize` testing

The domain of source values we need to consider is:

* `null`
* `undefined` (there is no need to distinguish between a field/variable being explicitly assigned the value of `undefined` and it being so by virtue of not having been initialised)
* Primitive and non-primitive values (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#Distinction_between_string_primitives_and_String_objects))

### Overall PR TODO list

1. Agree on structure of code, tests and docs for a couple of types, e.g. `string` and `bool`
2. Work out and fix reflection (manifests via `TestReflection` failing in `js` tests)
3. Complete revised implementation of `$internalize` and `$externalize` for all types to allow for testing that code in real-life code bases (not just tests)
4. Finish docs for all types
5. Tighten up wording of `*js.Object`-special struct types in `js` package (unless we work this into the tables for `$internalize` and `$externalize`
6. Document ways in which you can fallback to Javascript "loose" type conversions

### Consideration for later PRs

* Splitting `$internalize` and `$externalize` into separate, smaller type-specific functions 
* Fixing the fact that `null` and `undefined` do not internalize as `(*T)(nil)` for a `*js.Object`-special struct type `T`


js/js_test.go
```diff
@@ -571,3 +571,20 @@ func TestUint8Array(t *testing.T) {
 		t.Errorf("Non-empty byte array is not externalized as a Uint8Array")
 	}
 }
+
+func TestInternalizeZeroVal(t *testing.T) {
+	type S struct {
+		*js.Object
+
+		string `js:"string"`
```
shurcooL: Since this is a test of internalization logic (as I understand from our conversation in #617), I would suggest rewriting it to avoid using `js` struct field tags. I think it's simpler to use `Object.String` method, it's lower level and more explicit, making it more clear what the test is testing.

myitcv: @shurcooL I think for completeness we'll need to test `String()` as well, but in order that we can cover all Go types I think we need to follow the struct field tags approach... unless I'm missing something?

shurcooL: > but in order that we can cover all Go types I think we need to follow the struct field tags approach...

Are methods like `String`, `Float`, `Bool` `Uint64`, etc., not sufficient? What's not possible without struct field tags?

myitcv: All the types that do not have corresponding methods on `Object`. e.g. the specific numeric types like `int8`.

shurcooL: Can `Interface().(int8)` be done for those, or is that not equivalent?

myitcv: I think `Interface()` is another test case altogether because it [uses the Javascript type of the value to decide which Go type to resolve to](https://github.com/gopherjs/gopherjs/blob/3496c6f94e1b945f3fc9580f6982675c0a74cd6c/compiler/prelude/jsmapping.go#L258-L308). So in the context of this discussion, I suspect use of `Interface()` wouldn't actually be a problem per se.

What you've proposed transpiles to:

```javascript
$assertType($internalize(o.A, $emptyInterface), $Int8)
```

So we still need to have test cases for all the Go types... because it's `$internalize` that we're ultimately testing here.

shurcooL: Got it, thanks. Then it's okay to use the struct field tags, but my request is to document that what's being tested is internalize behavior, not the struct field tag behavior.

myitcv: Ah I see; 100% agree on that point.

js/js.go
```diff
@@ -1,28 +1,47 @@
...
+//
+// Source values annotated with † are generally applicable to all valid
+// values of the target type. e.g. for target type string, "ok" represents
+// all valid string values.
+//
+// Externalization
+//
+// When values pass from Go to Javascript, a process known as externalization,
+// the following conversion table is applied:
+//
```
flimzy: Table missing?

js/js.go
```diff
@@ -1,28 +1,47 @@
...
+// target type if a conversion is attempted, e.g. a Javascript number value
+// being assigned to a string type Go variable.
+//
+// Source values annotated with † are generally applicable to all valid
+// values of the target type. e.g. for target type string, "ok" represents
+// all valid string values.
+//
+// Externalization
+//
+// When values pass from Go to Javascript, a process known as externalization,
```
flimzy: Incomplete sentence

myitcv: Thanks @flimzy - I should have mentioned this entire PR is still very much WIP.

In particular, I haven't even started on the `$externalize` side of things. Had some fairly interesting edge cases on the `$internalize` side of things

js/js.go
shurcooL: See next line.

myitcv:
@flimzy @shurcooL - thanks both for taking a look at this.

I should have been clearer in my update message in #617 because I value your time... This is still WIP so I will give you a shout when it's essentially "ready" for review. There are still too many rough edges. 

My plan is to get a couple of types (probably `string` and `bool`) done for both `$internalize` and `$externalize` and then get some feedback on how things look, from code and tests through to docs etc.

Just wanted to post an update to confirm I'm still "on" this, albeit after a significant period of inactivity!

shurcooL:
> I should have been clearer in my update message in #617 because I value your time... This is still WIP so I will give you a shout when it's essentially "ready" for review. There are still too many rough edges.

No problem.

I suggest using Go code review convention for achieving this, put a "DO NOT REVIEW" text anywhere in the PR description to signal that. (See https://groups.google.com/d/msg/golang-dev/YU56eKYiXJg/K1CkXo7Iy0gJ.)

myitcv:
> I suggest...

Done. I've switched from using my `[WIP]` convention in the title to your suggestion in the PR description.

Thanks again

myitcv:
@shurcooL @flimzy et al.

Given there's quite a lot to work through in this PR, I've updated the description in this PR with, amongst other things, the following:

* Status
* Overall PR TODO list

Would welcome your thoughts on the TODO list, in terms of content but also approach (i.e. using the TODO list as a means of communicating how I plan to go about things).

At this point I think I've got to the end of step 1, hence am marking this for a review.

All thoughts/feedback welcomed.

Thanks

myitcv:
Worth pointing out that at this stage, not least because of point 2 from the TODO list, I'm not expecting the build to pass.

js/js.go
```diff
@@ -1,28 +1,60 @@
...
+//  |                |               | undefined               | ""     |
+//  |                |               | ""                      | ""     |
+//  |                |               | new String("")          | ""     |
+//  |                |               | "ok" †                  | "ok"   |
+//  |                |               | new String("ok") †      | "ok"   |
+//  |----------------+---------------+-------------------------+--------|
+//  | bool           | none          | null                    | false  |
+//  |                |               | undefined               | false  |
+//  |                |               | false †                 | false  |
+//  |                |               | new Boolean(false) †    | false  |
```
flimzy: Is it worth including `true` examples here, for the sake of completeness, similar to the non-zero value examples for strings?

js/js_test.go
```diff
@@ -467,7 +467,7 @@ func TestCallWithNull(t *testing.T) {
 func TestReflection(t *testing.T) {
 	o := js.Global.Call("eval", "({ answer: 42 })")
 	if reflect.ValueOf(o).Interface().(*js.Object) != o {
-		t.Fail()
+		t.Fatal()
```
flimzy: AFAIK, `Fatal()` is for reporting a failure message.  Maybe `FailNow()` would be more semantically correct?

js/js_test.go
```diff
@@ -571,3 +572,207 @@ func TestUint8Array(t *testing.T) {
...
+
+	string string `js:"string"`
+	bool   bool   `js:"bool"`
+}
+
+func TestInternalizeString(t *testing.T) {
+	fieldName := "string"
+	zero := ""
+
+	s := &Internalize{Object: js.Global.Get("Object").New()}
```
flimzy: Stylistic note:  I wonder if a table-based test would make sense here?  It may make the tests more readable, reduce the need for the `// ***` comment blocks, and if coupled with `t.Run(...)` calls for each test, could make the output more readable.

js/js_test.go
shurcooL: Is it not possible to include a helpful error message in the `t.Fatal` call? If possible, that's better.

js/js_test.go
myitcv: You raise a good point. I half considered this myself, but postponed giving it much thought until we've confirmed we're actually testing the right stuff... at which point extracting to table tests should be relatively easy. 

The little thought I did give it however amount to the following: I'm not convinced we _can_ actually use table tests. We need to address specific fields of the `Internalize` struct and do so in such a way that the `gopherjs` compiler sees us doing so and compiles to the corresponding Javascript. Put another way, the code we write in these tests needs to correspond _exactly_ to the code that will be written in the wild.

Therefore the one approach I have considered fairly seriously is code generating the tests. Because they do follow a formula that's for sure.

But as I said, I haven't given this much time/though so would very much welcome any suggestions.

js/js_test.go
myitcv: To be honest I was being a little lazy here in just flipping the existing `t.Fail()` to `t.Fatal()` to help with debugging step 2 from the TODO. 

I'm sure we could add helpful messages to the `t.Fatal()` calls... because the existing `t.Fail()` calls are equally lossy. Could even be something we do in another PR.

js/js.go
myitcv: We would end up adding two lines: `true` and `new Boolean(true)` - hence why I opted for the † callout.

js/js_test.go
shurcooL: It's usually easier to do these minor things as you're working in a relevant context than later. But this is completely up to you, and not critical for getting the PR into a mergeable state.

js/js_test.go
myitcv: @flimzy just pushed up a code-generated version of the tests which seems to work well. Let me know what you think.

myitcv:
Just pushed up a new commit per https://github.com/gopherjs/gopherjs/issues/617#issuecomment-324688005

flimzy:
Nearly 4 years later, @nevkontakte and I have decided to close this PR.  We like the effort here, and some of it will be very informative if we address these issues in the future, but with limited bandwidth, and the author having left GopherJS development, we feel this PR is at a dead end.

PR #623, opened 2017-04-10
js: Point to new location of Interface conversion details.

### LINKED ISSUE #622: Interface() undocumented
[The comment for Object.Interface](https://godoc.org/github.com/gopherjs/gopherjs/js#Object.Interface) just says "Interface returns the object converted to interface{}. See GopherJS' README for details." But the README doesn't seem to mention Interface (any more?).

shurcooL:
When the "See GopherJS' README for details." phrase was added in 88f12aa6b316ade3347e7723510a954e09edff1d, the JavaScript <-> Go conversion table was in README. It was moved in c6e849f88527ab0a93f7d4ee4d9c67a65db5285c to the [package comment of `js` package](https://github.com/gopherjs/gopherjs/blob/3496c6f94e1b945f3fc9580f6982675c0a74cd6c/js/js.go#L5-L23).

Fixes #622.

### DIFF
--- a/js/js.go
+++ b/js/js.go
@@ -73,7 +73,7 @@ func (o *Object) Uint64() uint64 { return o.object.Uint64() }
 // Float returns the object converted to float64 according to JavaScript type conversions (parseFloat).
 func (o *Object) Float() float64 { return o.object.Float() }
 
-// Interface returns the object converted to interface{}. See GopherJS' README for details.
+// Interface returns the object converted to interface{}. See table in package comment for details.
 func (o *Object) Interface() interface{} { return o.object.Interface() }
 
 // Unsafe returns the object as an uintptr, which can be converted via unsafe.Pointer. Not intended for public use.


js/js.go
```diff
@@ -73,7 +73,7 @@ func (o *Object) Uint64() uint64 { return o.object.Uint64() }
 // Float returns the object converted to float64 according to JavaScript type conversions (parseFloat).
 func (o *Object) Float() float64 { return o.object.Float() }
 
-// Interface returns the object converted to interface{}. See GopherJS' README for details.
+// Interface returns the object converted to interface{}. See table in package comment for details.
```
neelance: Would "package documentation" be better? Still approved.

shurcooL: I'm worried that package documentation means all of the documentation, it's not very specific.

I wanted to refer to the package comment - the comment on package declaration - which is just the top-most section of package documentation, where the table is contained.

From https://blog.golang.org/godoc-documenting-go-code:

> **Comments on package declarations** should provide **general package documentation**. These comments can be short, like the sort package's brief description:
>
> ```Go
> // Package sort provides primitives for sorting slices and user-defined
> // collections.
> package sort
> ```
>
> They can also be detailed like the gob package's overview. That package uses another convention for packages that need large amounts of introductory documentation: the **package comment** is placed in its own file, doc.go, which contains only those comments and a package clause.
>
> When writing **package comments** of any size, keep in mind that their first sentence will appear in godoc's package list.

Emphasized relevant phrases used. What do you think?

shurcooL: @neelance Ping. Is the above convincing enough?

shurcooL: It's been a few days, and I still haven't heard back on my comment above. Since this has approval, I'll assume it's convincing and merge this in order to make progress.

If there's anything that can be improved or needs to be followed up, another PR can be made.

PR #629, opened 2017-04-18
Improve "source maps disabled" warning text.

### LINKED ISSUE #627: "Source maps disabled" warning message is confusing/misleading.
When the source-map-support Node module is not installed, it prints a warning:

```
```

It's the "**with source-map-support module**" part that's important. I think mentioning the Node.js version is not very relevant, and can be misleading.What it really means is Node.js 4.x or newer, and most Node versions are much newer than that by now.

See how @slimsag interpreted it in https://github.com/gopherjs/gopherjs/issues/624#issuecomment-294386098:

> I don't know why I still get the Node.js 4.x warning when my `node --version` reports `v7.9.0` though.

shurcooL:
Be more specific that it's the source-map-support module that needs to be installed. Point to the instructions for source-map-support module installation.

Fixes #627.

/cc @slimsag Can you review this text? Do you think it's an improvement in readability over what you saw before?

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -771,7 +771,7 @@ func runNode(script string, args []string, dir string, quiet bool) error {
 		allArgs = []string{"--require", "source-map-support/register"}
 		if err := exec.Command("node", "--require", "source-map-support/register", "--eval", "").Run(); err != nil {
 			if !quiet {
-				fmt.Fprintln(os.Stderr, "gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.")
+				fmt.Fprintln(os.Stderr, "gopherjs: Source maps disabled. Install source-map-support module for nice stack traces. See https://github.com/gopherjs/gopherjs#installation-and-usage.")
 			}
 			allArgs = []string{}
 		}
--- a/tool.go
+++ b/tool.go
@@ -771,7 +771,7 @@ func runNode(script string, args []string, dir string, quiet bool) error {
 		allArgs = []string{"--require", "source-map-support/register"}
 		if err := exec.Command("node", "--require", "source-map-support/register", "--eval", "").Run(); err != nil {
 			if !quiet {
-				fmt.Fprintln(os.Stderr, "gopherjs: Source maps disabled. Install source-map-support module for nice stack traces. See https://github.com/gopherjs/gopherjs#installation-and-usage.")
+				fmt.Fprintln(os.Stderr, "gopherjs: Source maps disabled. Install source-map-support module for nice stack traces. See https://github.com/gopherjs/gopherjs#gopherjs-run-gopherjs-test.")
 			}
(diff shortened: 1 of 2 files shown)


tool.go
```diff
@@ -771,7 +771,7 @@ func runNode(script string, args []string, dir string, quiet bool) error {
 		allArgs = []string{"--require", "source-map-support/register"}
 		if err := exec.Command("node", "--require", "source-map-support/register", "--eval", "").Run(); err != nil {
 			if !quiet {
-				fmt.Fprintln(os.Stderr, "gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.")
+				fmt.Fprintln(os.Stderr, "gopherjs: Source maps disabled. Install source-map-support module for nice stack traces. See https://github.com/gopherjs/gopherjs#installation-and-usage.")
```
flimzy: Is Node 4.x a minimum requirement for that module? If so, it's probably worth still mentioning that, but possibly not in the error message itself.

Many environments, notably in some CI setups, still use Node < 4.x.

flimzy: I wonder if it would also be better to link directly to the syscalls page (https://github.com/gopherjs/gopherjs/blob/master/doc/syscalls.md)?  Or make it more prominent on the page linked to (possibly with an anchor tag)?

shurcooL: > Is Node 4.x a minimum requirement for that module? If so, it's probably worth still mentioning that, but possibly not in the error message itself.

It is still mentioned at the linked page. See https://github.com/gopherjs/gopherjs#installation-and-usage:

> If you want to use `gopherjs run` or `gopherjs test` to run the generated code locally, install Node.js 4.x and the module `source-map-support`:
>
> ```
> npm install --global source-map-support
> ```

> I wonder if it would also be better to link directly to the syscalls page (https://github.com/gopherjs/gopherjs/blob/master/doc/syscalls.md)?

But that page doesn't talk about `source-map-support` module at all. Pointing to it would be very confusing.

> Or make it more prominent on the page linked to (possibly with an anchor tag for direct linking)?

The Installation and Usage section is only 4 paragraphs long. Does it really need to be broken down into smaller subsections, or can we expect people to read the whole thing?

flimzy: > But that page doesn't talk about source-map-support module at all. How would it be better?

You're right.  Brainfart.  I was thinking of the _other_ reason to install a node module.

shurcooL: Well, after looking at it more, I thought making a small subsection could work well after all.

Take a look at https://github.com/gopherjs/gopherjs/pull/629/commits/4e30b5aa07e19a395bb551c01bc7607a8b0bffbc. How's it look?

flimzy: Looks great! :+1:

shurcooL:
People above seem to like this change. @neelance, does this change have your approval?

PR #635, opened 2017-04-27
build: Fix shadowing bug in staleness calculation.

### LINKED ISSUE #559: GopherJS incorrectly uses stale archive files when latest source code fails to build.
1. Create the following three files.
2. Run `gopherjs serve` and load `alltom/` in your browser.
3. Edit iface.go to change `Bar() int` to `Bar2() int`.
4. Edit main.go to change `impl.MakeFoo().Bar()` to `impl.MakeFoo().Bar2()`.
5. Reload your browser.

Instead of seeing a compile error about FooImpl no longer implementing Foo, you get a runtime error saying "Uncaught TypeError: impl.MakeFoo(...).Bar2 is not a function". It looks like `gopherjs serve` doesn't realize that impl.go needs to be recompiled.

## alltom/iface/iface.go
```go
package iface

type Foo interface {
	Bar() int
}
```

## alltom/impl/impl.go
```go
package impl

import "alltom/iface"

type FooImpl int

func MakeFoo() iface.Foo {
	return FooImpl(0)
}

func (f FooImpl) Bar() int {
	return 42
}
```

## alltom/main.go
```go
package main

import "alltom/impl"

func main() {
	impl.MakeFoo().Bar()
}
```

shurcooL:
The parent scope `pkg` variable was inadvertently shadowed in #411. As a result, the value of `impModeTime` was wrong, and the if statement could never be true, and `pkg.SrcModTime` was never updated to a later time.

The shadow happened at https://github.com/gopherjs/gopherjs/commit/e3a9838fe85c461e4cd869da5065ea1f21b1759e#commitcomment-21940173.

Fixes #559 (possibly other similar issues, but they'll need to be verified).

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -532,11 +532,11 @@ func (s *Session) BuildPackage(pkg *PackageData) (*compiler.Archive, error) {
 			if importedPkgPath == "unsafe" || ignored {
 				continue
 			}
-			pkg, _, err := s.buildImportPathWithSrcDir(importedPkgPath, pkg.Dir)
+			importedPkg, _, err := s.buildImportPathWithSrcDir(importedPkgPath, pkg.Dir)
 			if err != nil {
 				return nil, err
 			}
-			impModeTime := pkg.SrcModTime
+			impModeTime := importedPkg.SrcModTime
 			if impModeTime.After(pkg.SrcModTime) {
 				pkg.SrcModTime = impModeTime
 			}


johanbrandhorst:
That is a really obscure bug and really shows how nasty go variable shadowing can be -  I thought `go vet` had a check for this kind of thing though?

myitcv:
Sadly not by default:

https://golang.org/cmd/vet/#hdr-Shadowed_variables

```bash
$ go vet -help
usage: vet [-n] [-x] [build flags] [packages]

Vet runs the Go vet command on the packages named by the import paths.

For more about vet, see 'go doc cmd/vet'.
For more about specifying packages, see 'go help packages'.

To run the vet tool with specific options, run 'go tool vet'.

The -n flag prints commands that would be executed.
The -x flag prints commands as they are executed.

For more about build flags, see 'go help build'.

See also: go fmt, go fix.
```

You need to do something like:

```bash
go list -f '{{.Dir}}' ./... | xargs go tool vet -shadow
```

flimzy:
It might be worth running this in our CI script, along with the `go fmt` check: https://github.com/gopherjs/gopherjs/blob/master/circle.yml#L17

myitcv:
@shurcooL great find!

I've just tried this out. For a particular staleness/caching problem I'm seeing this fix unfortunately doesn't help... will continue to investigate.

shurcooL:
@flimzy We already run `go tool vet` on [all GopherJS packages](https://github.com/gopherjs/gopherjs/blob/9659c814f1d54d63f9c623449a7111d3864c1361/circle.yml#L18-L19), just without `-shadow`. Sadly, `-shadow` gives many false positives like:

```
build/build.go:110: declaration of "err" shadows declaration at build/build.go:77
build/build.go:535: declaration of "err" shadows declaration at build/build.go:506
build/build.go:546: declaration of "err" shadows declaration at build/build.go:506
...
```

It did also spot the `pkg` shadow (when run on previous commit). Maybe if we filter out instances of "err" being shadowed, it'd be viable to include it?

---

> I've just tried this out. For a particular staleness/caching problem I'm seeing this fix unfortunately doesn't help... will continue to investigate.

@myitcv If you can make a minimal reproduce sample and report that, I'd be happy to investigate. I didn't expect this fix to necessarily resolve all problems, this was just the first step before others.

flimzy:
@shurcooL Are those really false positives? I know I've seen bugs (in code @ work) where `err` being shadowed was the culprit.  My personal preference is to avoid all shadowed variables, even when they seem innocuous (as most shadowed `err`s do).

shurcooL:
I've never made it a goal for myself to not shadow `err` variables, so it'd be a false positive in most code I write. For me, if a variable is named `err`, its value **must** be consumed on next immediate line, or the one after (to allow inserting `foo.Close()`), and after that `err` becomes available for reuse only.

If I need to use a value of `err` in later code, I always call it `somethingError` instead. See https://dmitri.shuralyov.com/idiomatic-go#error-variable-naming.

So, I don't know whether they're false positives or not in this codebase, but I expect them to be if other people follow similar practice (and as far as I know, most people do).

flimzy:
> For me, if a variable is named err, its value must be consumed on next immediate line

I agree with this, and follow the same rule. But of course the compiler doesn't know or enforce this rule. And I've seen it violated (almost certainly by accident, due to copy-paste errors, or similar, leaving both unchecked, or double-checked `err` values).  I don't really expect this sort of problem in a codebase like this one, though, where good coding practices, including peer review, are routine.

neelance:
Ha, there wasn't much peer review on most of the code. Honestly I did not
follow the rule that Dmitri suggested, but I like it now that I hear it.
I'll keep it in mind for the future.

Jonathan Hall <notifications@github.com> schrieb am Fr., 28. Apr. 2017 um
17:35 Uhr:

neelance:
In fact, GopherJS was my first big open source Go project. The first commit
is from 2013. I've learned a lot since then.

Richard Musiol <mail@richard-musiol.de> schrieb am Fr., 28. Apr. 2017 um
17:37 Uhr:

PR #644, opened 2017-05-20
Add support to gopherjs tool for specifying import path patterns and relative import paths.

### LINKED ISSUE #302: `gopherjs test ./with/relative/path` failing

shurcooL:
This change improves the `gopherjs build`, `gopherjs install`, `gopherjs test` commands to behave more like the equivalent `go` commands. Namely, while previously not possible, one can now:

- Use import path [patterns](https://golang.org/cmd/go/#hdr-Description_of_package_lists) when specifying packages:

    ```
    gopherjs test github.com/user/repo/...
    ```

- Use relative import paths when specifying packages:

    ```
    gopherjs test ./foo/bar
    ```

- Use relative patterns when specifying packages:

    ```
    gopherjs test ./...
    ```

It also includes other minor improvements and simplifications to the code and behavior of the `gopherjs` tool.

The change is somewhat large, but broken up into 4 individual commits with detailed description in their commit message.

Fixes #302. /cc @slimsag FYI, since you recently felt the pain from it not being resolved.

Reviewing and/or testing this PR would be helpful.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -62,7 +62,14 @@ func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
 // If an error occurs, Import returns a non-nil error and a nil
 // *PackageData.
 func Import(path string, mode build.ImportMode, installSuffix string, buildTags []string) (*PackageData, error) {
-	return importWithSrcDir(path, "", mode, installSuffix, buildTags)
+	wd, err := os.Getwd()
+	if err != nil {
+		// Getwd may fail if we're in GOARCH=js mode. That's okay, handle
+		// it by falling back to empty working directory. It just means
+		// Import will not be able to resolve relative import paths.
+		wd = ""
+	}
+	return importWithSrcDir(path, wd, mode, installSuffix, buildTags)
 }
 
 func importWithSrcDir(path string, srcDir string, mode build.ImportMode, installSuffix string, buildTags []string) (*PackageData, error) {
--- a/compiler/version_check.go
+++ b/compiler/version_check.go
@@ -6,4 +6,4 @@ package compiler
 const ___GOPHERJS_REQUIRES_GO_VERSION_1_8___ = true
 
 // Version is the GopherJS compiler version string.
-const Version = "1.8-1"
+const Version = "1.8-2"
--- a/tool.go
+++ b/tool.go
@@ -128,7 +128,6 @@ func main() {
 				pkgs := (&gotool.Context{BuildContext: *patternContext}).ImportPaths(args)
 
 				for _, pkgPath := range pkgs {
-					pkgPath = filepath.ToSlash(pkgPath)
 					if s.Watcher != nil {
 						pkg, err := gbuild.NewBuildContext(s.InstallSuffix(), options.BuildTags).Import(pkgPath, "", build.FindOnly)
 						if err != nil {
@@ -208,8 +207,6 @@ func main() {
 					}
 				}
 				for _, pkgPath := range pkgs {
-					pkgPath = filepath.ToSlash(pkgPath)
-
 					pkg, err := gbuild.Import(pkgPath, 0, s.InstallSuffix(), options.BuildTags)
 					if s.Watcher != nil && pkg != nil { // add watch even on error
 						s.Watcher.Add(pkg.Dir)
@@ -331,7 +328,6 @@ func main() {
 
 			pkgs := make([]*gbuild.PackageData, len(args))
 			for i, pkgPath := range args {
-				pkgPath = filepath.ToSlash(pkgPath)
 				var err error
 				pkgs[i], err = gbuild.Import(pkgPath, 0, "", options.BuildTags)
 				if err != nil {
--- a/tool.go
+++ b/tool.go
@@ -95,11 +95,8 @@ func main() {
 			s := gbuild.NewSession(options)
 
 			err := func() error {
-				if len(args) == 0 {
-					return s.BuildDir(currentDirectory, currentDirectory, pkgObj)
-				}
-
-				if strings.HasSuffix(args[0], ".go") || strings.HasSuffix(args[0], ".inc.js") {
+				// Handle "gopherjs build [files]" ad-hoc package mode.
+				if len(args) > 0 && (strings.HasSuffix(args[0], ".go") || strings.HasSuffix(args[0], ".inc.js")) {
 					for _, arg := range args {
 						if !strings.HasSuffix(arg, ".go") && !strings.HasSuffix(arg, ".inc.js") {
 							return fmt.Errorf("named files must be .go or .inc.js files")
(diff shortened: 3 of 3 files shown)


build/build.go
```diff
@@ -62,7 +62,14 @@ func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
 // If an error occurs, Import returns a non-nil error and a nil
 // *PackageData.
 func Import(path string, mode build.ImportMode, installSuffix string, buildTags []string) (*PackageData, error) {
-	return importWithSrcDir(path, "", mode, installSuffix, buildTags)
+	wd, err := os.Getwd()
+	if err != nil {
+		// Getwd may fail if we're in GOARCH=js mode. That's okay, handle
+		// it by falling back to empty working directory. It just means
+		// Import will not be able to resolve relative import paths.
+		wd = ""
```
flimzy: Is it worth it to detect the GOARCH=js case, and only fall back then, and otherwise fail?

shurcooL: > and otherwise fail?

That's what'll happen anyway. If a relative import path is specified, and `wd` is left blank, the next function will return an error like "cannot resolve relative import path".

So it doesn't seem to be worth it to me, since the behavior wouldn't change.

neelance: This should be at the tool level. It already has https://github.com/gopherjs/gopherjs/blob/master/tool.go#L40 which you can use. Either export `importWithSrcDir` or consider renaming `importWithSrcDir` to `Import`. I wouldn't mind breaking the API in this way, it is easy to detect and fix for the users and then it would also fit better to `go/build.Context.Import`.

shurcooL: @neelance That's exactly what I want to do next, this was just a working intermediate solution. Just checking, but did you consider the commit message of f79744a233766fd8897d5f2dc8f05b6ee0e72d8c before making the comment?

As long as there are no functional issues with the current implementation (I'm not aware of any at this time), I'd prefer to proceed with a smaller scope PR and follow up with breaking the `gopherjs/build` API in future changes, rather than doing everything at once just to resolve #302.

Would it help if I copied the future steps from commit message into a TODO comment?

Or do you feel strongly all that needs to be done before this can be merged?

neelance: I did not see the commit message before. Yes, that's exactly what I'm talking about. What's the advantage of having this intermediate state?

shurcooL: > What's the advantage of having this intermediate state?

There are many.

It keeps the scope of this PR smaller. To resolve #302, that's it.

Having a smaller PR/smaller scope has many benefits and I think it's a good thing to aim for.  A smaller change to review carefully. It means it can be merged sooner. Then, it can be tested by more people sooner, and if there are any regressions, they can be reported sooner. A fix for an unexpected issue can be simple, or it may turn out that there's a big flaw and the best course of action is to revert the entire PR and rethink the approach. That way, the revert only affects #302 and doesn't change any public APIs. It doesn't create ripple effect, forcing reverts in external repos.

It also means the future changes that involve breaking APIs can be more carefully thought out, planned, and implemented, without holding back and being married to the fix to #302. I wanted to take the following next steps (one step at a time):

1. Refactor internal code of `gopherjs/build` package. It's all in a single huge .go file, and the order of identifiers doesn't follow a very logical pattern. I find it's hard to read/understand. To implement this fix, I first performed this change to better understand what was going on. Then, I applied and tested the fix. Then, once I understood it, I backported it to `master` and made this pretty small PR with an easy to read diff. But before I do anything else, I want to follow up by applying this refactor next.

2. Next, I want to mark the `build`/`compiler`/etc. APIs as not subject to compatibility promise and may change. Similar to how https://godoc.org/golang.org/x/build/gerrit API is documented:

    > The API is not subject to the Go 1 compatibility promise and may change at any time.

    This is unfortunate, but feels neccessary to make progress. The current APIs were unofficially always meant to be private and able to change, but not explicitly documented that way. This would be about making that more clear, and then being able to take advantage of it to improve the code and APIs.

3. Next, I want to move https://godoc.org/github.com/shurcooL/gopherjslib and https://godoc.org/github.com/shurcooL/go/gopherjs_http into gopherjs repo (also without promise of compatibility and ability to change at any time). The goal is to de-duplicate and clean up that functionality once it's all in one repo, where atomic operations can be done, and everything can be tested via Circle CI. (This was something I asked if you'd be okay in the past, and you told me you were fine with it at that time.)

4. Then, once it's in the same repo, I want to refactor/clean up the `build`/`compiler`/`gopherjslib`/`gopherjs_http` APIs, which would certainly include having the working directory be passed to `Import` (just like `go/build.Import`) instead of `build` package trying to figure it out on its own. I want to make it possible for `gopherjs serve` to reuse the same functionality from `gopherjs_http` instead of an unexported implementation within `tool.go`.

That's the general direction I wanted to move, and it's not something I can realistically make happen as part of 1 PR.

If there are sudden pressing needs (for functionality reasons), we can take a shortcut and make API changes sooner, but that should be done in context of a specific requirement.

In summary, if it's possible to fix #302 in an **acceptable way** (i.e., no problems with functionality) without changing public APIs, I want this PR not to change public APIs. If there are functionality problems (i.e., some sort of problem with behavior) that require public APIs changed in order to fix #302, then I have absolutely no problem with doing that, since it's neccessary. But I don't want to increase the scope of this PR to include improvements to the API without a good reason - I don't think that's neccessary nor a good idea.

Does that sound reasonable, or do you think a different approach is better (if so, why)?

neelance: Fine with me. ;-)

shurcooL: Great, thanks!

One other benefit I didn't explicitly mention is that this PR removes more code than it adds. The code it removes contained partially incorrect and hard to understand logic. It also reduces the number of different code paths taken. Being able to merge this PR sooner makes future changes easier to make, because there'll be less code to read/understand/deal with.

shurcooL:
I plan to test that the GopherJS Playground continues to work successfully with this change. I will also continue to test this change locally for some more time, to gain confidence, before merging.

If other people try it out and confirm there are no issues, that's still helpful.

shurcooL:
This PR is a good time to bump the version string. It'll help to tell what version was used if there are bug reports in the future.

shurcooL:
I've regenerated GopherJS Playground with this version, and didn't spot any issues. Merging.

PR #647, opened 2017-05-20
Include more information in error message when fetch() fails

albrow:
I recently ran into a frustrating issue when running browser tests for [go-humble/rest](https://github.com/go-humble/rest). The only error message I was seeing was:

```
net/http: fetch() failed
```

After I made a small tweak to GopherJS, the error message turns into:

```
net/http: fetch() failed: TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body
```

With this additional information I was able to quickly fix the bug in go-humble/rest and move on.

I can understand if you prefer to keep the error message short, but it would have saved me a lot of time to have this longer error message available from the beginning. I suspect this will save other people some time too.


shurcooL:
No, there wasn't an intention to keep the message short, so this is a welcome improvement, thanks.

compiler/natives/src/net/http/fetch.go
```diff
@@ -116,7 +117,7 @@ func (t *fetchTransport) RoundTrip(req *Request) (*Response, error) {
 		},
 		func(reason *js.Object) {
 			select {
-			case errCh <- errors.New("net/http: fetch() failed"):
+			case errCh <- fmt.Errorf("net/http: fetch() failed: %s", reason.String()):
```
shurcooL: Can you be sure that `reason` will never be nil? Otherwise it'd be a nil pointer dereference panic, right?

shurcooL: I guess it should never be `nil` for the same reason `result` is never nil in the success case.

I looked at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then, and don't see an explicit confirmation that reason won't be nil...

But maybe it would be? Is there a way we can be sure?

albrow: Done. I'm not 100% sure and it's pretty easy to add this check.

shurcooL: It'd be nice if could find out. That way, the code could be simpler.

compiler/natives/src/net/http/fetch.go
```diff
@@ -115,8 +116,15 @@ func (t *fetchTransport) RoundTrip(req *Request) (*Response, error) {
 			}
 		},
 		func(reason *js.Object) {
+			var jsErr error
```
shurcooL: Style-wise, just name this `err`. I don't see why it should be `jsErr`.

compiler/natives/src/net/http/fetch.go
```diff
@@ -115,8 +116,15 @@ func (t *fetchTransport) RoundTrip(req *Request) (*Response, error) {
...
+			if reason != nil {
+				jsErr = fmt.Errorf("net/http: fetch failed: %s", reason.String())
+			} else {
+				jsErr = errors.New("net/http: fetch() failed")
```
shurcooL: Why is one message "net/http: fetch" but the other "net/http: fetch()"? It'd be better to make them consistent.

shurcooL:
I found that it's indeed possible to pass `null` as a rejection reason with JavaScript promises:

```JavaScript
var p1 = new Promise( (resolve, reject) => {
  //resolve('Success!');
  // or
  reject (null);
} );

p1.then( value => {
  console.log(value); // Success!
}, reason => {
  console.log(reason === null); // true
  console.log(reason);          // "null"
} );
```

(On [JSFiddle](https://jsfiddle.net/gr5mhL6v/).)

That prints `true` and `null`.

So the question becomes, does any implementation of Fetch API ever pass `null` as a rejection reason (i.e., is it allowed by its specification).

albrow:
I addressed the style comments.

As far as I can tell, neither the [ECMAScript 2015 Spec for Promises](http://www.ecma-international.org/ecma-262/6.0/#sec-promise.reject) nor the [Living Standard for the Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) explicitly forbid a null value showing up here.

A quick test shows that Chrome allows you to do things like this:

```javascript
Promise.reject(null).catch(function(e) {console.log(e)})
Promise.reject(undefined).catch(function(e) {console.log(e)})
```

I'm not sure what the best course of action is, so I'll defer to you or anyone else who can chime in.

shurcooL:
Ah, well, I just remembered one GopherJS-specific thing. It's actually safe to call the `js.Object` methods on a nil value of `*js.Object`. It's a way of converting it to that type.

E.g., this is fine:

https://gopherjs.github.io/playground/#/kY9ne7yWeM

It's effectively equivalent to:

```Go
func (o *Object) String() string {
    if o == nil {
        return "null"
    }
    return string(*o) // not really, but you get the idea
}
```

So, no panic is possible. You can revert to your original code, since it's shorter and better.

That said, we might want to improve the documentation of `js` package to mention this. It's also related to #617 (/cc @myitcv).

albrow:
Cool. I reverted back to the original state of this PR (without the nil check).

Issue #649, opened 2017-06-07
Bug: `float64(a - b) / float64(c)` can be treated as `float64(a) - float64(b) / float64(c)`


hajimehoshi:
https://gopherjs.github.io/playground/#/wOjEpp--2V

The calculation is wrong when float64 cast is used.

hajimehoshi:
I think translateConversion is the culprit (sorry if I'm wrong). I'm trying to make a PR.

PR #650, opened 2017-06-07
Use parenthesis for numeric expressions

hajimehoshi:
This PR fixes https://github.com/gopherjs/gopherjs/issues/649.

The problem was that a numeric expression like `float64(a - b) / c` was converted to `a - b / c` when `a -
 b` was float64. This PR adds parenthesis to all numeric expressions at `fixNumber`.

I think there was a potential bug that `int(a - b) / c` was converted to `a - b >> 0 / c`, but I couldn't find a actual problematic case.

### DIFF
--- a/tests/misc_test.go
+++ b/tests/misc_test.go
@@ -612,3 +612,14 @@ func TestReceiverCapture(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestParenthesis(t *testing.T) {
+	i1, i2, i3 := 4, 2, 2
+	if (i1 - i2) / i3 != int(i1 - i2) / int(i3) {
+		t.Fail()
+	}
+	f1, f2, f3 := 4.0, 2.0, 2.0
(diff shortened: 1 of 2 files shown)


hajimehoshi:
`TestNumGoroutine` failed, which I couldn't reproduce. Is this a CI problem?

tests/misc_test.go
```diff
@@ -612,3 +612,14 @@ func TestReceiverCapture(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestParenthesis(t *testing.T) {
```
neelance: :+1: for adding a test. Maybe better call it `TestTypeConversion`.

hajimehoshi:
Thank you for reviewing! Fixed.

tests/misc_test.go
hajimehoshi: Done.

neelance:
@hajimehoshi Thanks for contributing!

Issue #136, opened 2014-12-14
Huge file sizes


liamcurry:
Here are a [few simple examples](https://gist.github.com/liamcurry/5175c5b5559fdc126c6f) and their generated file sizes:

Filename | Source | Generated | Minified | Minified + Gzipped (Level 9)
--- | ---:| ---:| ---:| ---:
[simple.go](https://gist.github.com/liamcurry/5175c5b5559fdc126c6f#file-simple-go) | 30b | 61kb | 46kb | 12kb
[console.go](https://gist.github.com/liamcurry/5175c5b5559fdc126c6f#file-console-go) | 86b | 71kb | 53kb | 15kb
[websocket.go](https://gist.github.com/liamcurry/5175c5b5559fdc126c6f#file-websocket-go) | 296b | 1.4mb | 1.4mb | 69kb
[websocket_fork.go](https://gist.github.com/liamcurry/5175c5b5559fdc126c6f#file-websocket_fork-go) | 296b | 78kb | 59kb | 13kb
[websocket_2015-02-03.go](https://gist.github.com/liamcurry/5175c5b5559fdc126c6f#file-websocket_2015-02-03-go) | 399b | 542kb | 390kb | 96kb

(The huge jump in filesize in `websocket.go` is because of [github.com/dominikh/go-js-dom](https://github.com/dominikh/go-js-dom))

There are a lot of implementation details in the generated files that shouldn't be there. For example, I would assume that:
```go
console.Log("hello")
```
would compile to:
```go
console.log("hello")
```
but instead it compiles to:
```go
console.Log(new ($sliceType($emptyInterface))([new $String("hello")]));
```

I'd love to start using gopherjs in production, but this is a show stopper. Any ideas on how we can fix this?

edit: added gzipped sizes
edit: added `websocket_fork.go` to results table
edit: added `websocket_2015-02-03.go` to results table

neelance:
@liamcurry Output size has gotten smaller, especially for `honnef.co/go/js/dom`. Is that better?

liamcurry:
FYI using Google's Closure Compiler with advanced optimizations I was able to get the minified file size down to 218kb (another 179% reduction), but the code no longer works.

neelance:
There are also other changes that might have increased the gzipped output size a bit. In my testing of the version before be37568 and after, the gzipped size reduced.

If working code is not a requirement, then I can reduce the size to zero. ;-)

shurcooL:
[...]

PR #651, opened 2017-06-23
Go 1.9 support (version bump to GopherJS 1.9-1).

shurcooL:
This is a WIP PR to track work to support Go 1.9. `go1.9beta1` has been [released](https://groups.google.com/forum/#!topic/golang-announce/4NhDbvlC4pM).

Each commit is bite-sized, with rationale and description, and fixes a specific issue.

It's at a very early stage, I've only just started making the neccessary updates. It's functional enough to compile a `fmt.Println("hello")` program, but many things remain to be done and fixed. Some packages will fail to compile, many tests will fail. I'll post updates as I make progress.


shurcooL:
A very common test failure right now is:

```
Error: runtime error: native function not implemented: os/signal.signal_enable
```

<details>

```
/home/ubuntu/gopherjs/test.449611240:4
Error.stackTraceLimit=Infinity;var $global,$module;if(typeof window!=="undefined"){$global=window;}else if(typeof self!=="undefined"){$global=self;}else if(typeof global!=="undefined"){$global=global;$global.require=require;}else{$global=this;}if($global===undefined||$global.Array===undefined){throw new Error("no global object found");}if(typeof module!=="undefined"){$module=module;}var $packages={},$idCounter=0;var $keys=function(m){return m?Object.keys(m):[];};var $flushConsole=function(){};var $throwRuntimeError;var $throwNilPointerError=function(){$throwRuntimeError("invalid memory address or nil pointer dereference");};var $call=function(fn,rcvr,args){return fn.apply(rcvr,args);};var $makeFunc=function(fn){return function(){return $externalize(fn(this,new($sliceType($jsObjectPtr))($global.Array.prototype.slice.call(arguments,[]))),$emptyInterface);};};var $unused=function(v){};var $mapArray=function(array,f){var newArray=new array.constructor(array.length);for(var

Error: runtime error: native function not implemented: os/signal.signal_enable
    at $callDeferred (/home/ubuntu/gopherjs/test.449611240:4:28967)
    at $panic (/home/ubuntu/gopherjs/test.449611240:4:29591)
    at $packages.runtime.AK (/home/ubuntu/gopherjs/test.449611240:8:6190)
    at $packages.os/signal.M (/home/ubuntu/gopherjs/test.449611240:31:6122)
    at $packages.os/signal.Q (/home/ubuntu/gopherjs/test.449611240:31:6725)
    at Object.$packages.os/signal.$init (/home/ubuntu/gopherjs/test.449611240:31:7785)
    at Object.$packages.testing.$init (/home/ubuntu/gopherjs/test.449611240:34:113320)
    at Object.$packages.github.com/gopherjs/gopherjs/tests.$init (/home/ubuntu/gopherjs/test.449611240:35:67444)
    at $packages.main.$init (/home/ubuntu/gopherjs/test.449611240:40:1046)
    at $goroutine (/home/ubuntu/gopherjs/test.449611240:4:30168)
    at $runScheduled (/home/ubuntu/gopherjs/test.449611240:4:30879)
    at $schedule (/home/ubuntu/gopherjs/test.449611240:4:31107)
    at $go (/home/ubuntu/gopherjs/test.449611240:4:30756)
    at Object.<anonymous> (/home/ubuntu/gopherjs/test.449611240:44:1)
    at Object.<anonymous> (/home/ubuntu/gopherjs/test.449611240:47:4)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.runMain (module.js:575:10)
    at run (node.js:348:7)
    at startup (node.js:140:9)
    at node.js:463:3
FAIL	github.com/gopherjs/gopherjs/tests	0.366s
```

</details><br>

I've investigated it. The cause is that the `testing` package in Go 1.9 now imports `os/signal` package. That wasn't the case in Go 1.8.

However, GopherJS doesn't support `os/signal` right now. I get the same error when I import that package and try to run it with GopherJS 1.8-2.

shurcooL:
The [4 commits above](https://github.com/gopherjs/gopherjs/compare/31e62b3115f7b67ce0443cb9cd99c761e42bb4ec...71f605b028432679631270607e521234b3c47eca) have fixed the issue in https://github.com/gopherjs/gopherjs/pull/651#issuecomment-310794248. Some small packages, including `errors` and `github.com/gopherjs/gopherjs/tests` have their tests pass successfully now.

Next up, individual package test failures. I'm seeing problems because `sync.Map` isn't supported yet, and some failing test with `%+v` in `fmt` tests.

shurcooL:
The [6 commits above](https://github.com/gopherjs/gopherjs/compare/71f605b028432679631270607e521234b3c47eca...c0827ba1aefbafb15fce9a55e8d6925795be90f8) implement an initial version of `sync.Map` for `GOARCH=js`, fix all known issues in `reflect` package (which in turn fixes the `fmt` tests), fixes a panic in compiler due to type aliases and allows `math/bits`, `math/big` to compile and pass tests.

I saw that `encoding/json` tests pass, but `encoding/gob` fails with:

```
--- FAIL: TestEndToEnd (0.00s)
TypeError: dst.$set is not a function
    at typedmemmove (github.com/gopherjs/gopherjs/reflect.go:487:3)
```

That's up next.

shurcooL:
I've narrowed down the `encoding/gob` test failure with `dst.$set is not a function` to a single new struct field. It's a map with key and value being arrays. For now, I've disabled the failing struct field in 1f5c9b7793fefce1cc23605ee05ec2a642ce4d5b, but a TODO comment was added to figure out a fix.

In the [5 commits above](https://github.com/gopherjs/gopherjs/compare/c0827ba1aefbafb15fce9a55e8d6925795be90f8...a0ce01c9227f04ac82ba974be0043e32af587342), I've implemented missing functions in packages `math` and `net`. The latter allowed `crypto/x509` tests to pass. I've disabled the failing struct field case in `encoding/gob` tests, as mentioned above.

I think `time` package is one of the last remaining things to look at next. It's currently failing some tests due to non-functioning support for monotonic clocks.

shurcooL:
Strange, `crypto/x509` tests pass for me locally:

```
$ gopherjs test --short crypto/x509
PASS
ok  	crypto/x509	46.429s
```

But fail in CI:

```
/home/ubuntu/gopherjs/test.035766589:1480
        throw err;
        ^

Error: runtime error: index out of range
    at $callDeferred (/home/ubuntu/gopherjs/test.035766589:1412:17)
    at $panic (/home/ubuntu/gopherjs/test.035766589:1451:3)
    at $b (/home/ubuntu/gopherjs/test.035766589:43405:5)
    at $callDeferred (/home/ubuntu/gopherjs/test.035766589:1424:23)
    at $panic (/home/ubuntu/gopherjs/test.035766589:1451:3)
    at $packages.runtime.throw$1 (/home/ubuntu/gopherjs/test.035766589:2537:3)
    at $b (/home/ubuntu/gopherjs/test.035766589:50829:62)
    at $packages.testing.tRunner (/home/ubuntu/gopherjs/test.035766589:43442:8)
    at $goroutine (/home/ubuntu/gopherjs/test.035766589:1471:19)
    at $runScheduled (/home/ubuntu/gopherjs/test.035766589:1511:7)
    at $schedule (/home/ubuntu/gopherjs/test.035766589:1527:5)
    at $go (/home/ubuntu/gopherjs/test.035766589:1503:3)
    at Object.<anonymous> (/home/ubuntu/gopherjs/test.035766589:59349:1)
    at Object.<anonymous> (/home/ubuntu/gopherjs/test.035766589:59352:4)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.runMain (module.js:575:10)
    at run (node.js:348:7)
    at startup (node.js:140:9)
    at node.js:463:3
FAIL	crypto/x509	5.196s
```

It might be a macOS vs Linux difference. Needs further investigation.

_([Disabling minification in CI](b827aaee653d5db1b538134a3d02433652a10b42) is already coming in handy!)_

shurcooL:
All tests are now passing after the 4 commits above, except the mysterious index out of range panic in `crypto/x509` that only happens in CI and not for me locally. ¯\\\_(ツ)\_/¯

I guess I'll spin up a Linux VM and see if I can reproduce it there.

shurcooL:
I've investigated the `crypto/x509` failure. I was able to easily reproduce it on a Linux VM. Figuring the cause was easy when I used `-v` flag. There's a new Linux-only `TestEnvVars` test in Go 1.9 that tries to read files and access system roots, which GopherJS doesn't support now. So, I had to `t.Skip` this test, just like similar ones before it.

I think I'll add `-v` flag to CircleCI command that runs `gopherjs test`. It will cause verbose output, but help with debugging issues such as this one in the future.

The panic was happening because there's an off-by-one bug in the error handling of that new test. I've [reported it](https://go-review.googlesource.com/c/36093/#message-1f9e770808913b1259ead63c6cc8515163f9b421) and plan to send a CL upstream to fix it.

shurcooL:
All tests are passing! ✅ 🎉

There are known TODOs in the PR that can be looked at next. There are some things that require investigation. Help with that is always appreciated.

Aside from that, there are no known issues. I plan to start using and testing this new version in order to catch any problems. Help with testing is also welcome! Please report any issues you find.

neelance:
Awesome work! 🚀 

I'll try to find some time for a review soon.

shurcooL:
Friendly ping @neelance, in case it's helpful.

Go 1.9 is expected to be released in August 2017.

compiler/natives/src/sync/sync.go
```diff
@@ -14,11 +14,20 @@ func runtime_Semacquire(s *uint32) {
 }
 
 // SemacquireMutex is like Semacquire, but for profiling contended Mutexes.
-// Mutex profiling is not supported, so just use the same implementation.
+// Mutex profiling is not supported, so just use the same implementation as runtime_Semacquire.
 // TODO: Investigate this. If it's possible to implement, consider doing so, otherwise remove this comment.
-var runtime_SemacquireMutex = runtime_Semacquire
+func runtime_SemacquireMutex(s *uint32, lifo bool) {
+	// TODO: Use lifo if needed/possible.
+	if *s == 0 {
```
neelance: Maybe just call `runtime_Semacquire` here?

compiler/natives/src/time/time.go
```diff
@@ -42,9 +42,10 @@ func runtimeNano() int64 {
 	return js.Global.Get("Date").New().Call("getTime").Int64() * int64(Millisecond)
 }
 
-func now() (sec int64, nsec int32) {
+func now() (sec int64, nsec int32, mono int64) {
+	// TODO: Use mono if needed/possible.
 	n := runtimeNano()
-	return n / int64(Second), int32(n % int64(Second))
+	return n / int64(Second), int32(n % int64(Second)), 0
```
neelance: I'm surprised that returning zero works. Is there a test that actually depends on it?

compiler/natives/src/reflect/reflect.go
```diff
@@ -491,7 +491,8 @@ func makechan(typ *rtype, size uint64) (ch unsafe.Pointer) {
 	return unsafe.Pointer(js.Global.Get("$Chan").New(jsType(ctyp.elem), size).Unsafe())
 }
 
-func makemap(t *rtype) (m unsafe.Pointer) {
+func makemap(t *rtype, cap int) (m unsafe.Pointer) {
+	// TODO: Use cap if needed/possible.
```
neelance: I think there is no way to properly use `cap` here, so the comment can be removed.

compiler/natives/src/os/signal/signal.go
```diff
@@ -0,0 +1,13 @@
+// +build js
+
+package signal
+
+// Package signal is not implemented for GOARCH=js.
+// TODO: Implement if possible.
```
neelance: I don't see any reason why we would need to implement it. We don't have signals in the browser.

compiler/natives/src/testing/testing.go
```diff
@@ -0,0 +1,8 @@
+// +build js
+
+package testing
+
+func callerName(skip int) string {
+	// TODO: Implement if possible.
+	return "caller"
```
neelance: Maybe better return something along `"<unknown>"`.

compiler/natives/src/math/big/big.go
```diff
@@ -2,6 +2,13 @@
 
 package big
 
+// TODO: This is a workaround for https://github.com/gopherjs/gopherjs/issues/652.
+//       Remove after that issue is resolved.
+type Word uintptr
+
+// TODO: Consider using math_big_pure_go build tag for this package, instead of
```
neelance: This is a great idea!

circle.yml
```diff
@@ -18,7 +18,7 @@ test:
     - go tool vet *.go                                                   # Go package in root directory.
     - for d in */; do echo $d; done | grep -v tests/ | grep -v third_party/ | xargs go tool vet # All subdirectories except "tests", "third_party".
     - >
-      gopherjs test --short --minify
+      gopherjs test --short
```
neelance: Minification can sometimes break the code, for example if there are missing semicolons. CI should cover that.

compiler/natives/src/time/time.go
```diff
@@ -43,9 +43,8 @@ func runtimeNano() int64 {
 }
 
 func now() (sec int64, nsec int32, mono int64) {
-	// TODO: Use mono if needed/possible.
 	n := runtimeNano()
-	return n / int64(Second), int32(n % int64(Second)), 0
+	return n / int64(Second), int32(n % int64(Second)), n
```
neelance: Okay, this answers my earlier comment.

compiler/natives/src/math/big/big.go
shurcooL: Done in 8e80e53cc87099c2ae61f3aca9393b96ef955082.

compiler/natives/src/sync/sync.go
shurcooL: I was a little worried about increasing the call stack by 1 more level in something so low-level, which is why I copied the implementation with a comment. But `runtime_SemacquireMutex` should be rarely used, since mutex profiling isn't supported. So it should be fine. Done.

compiler/natives/src/reflect/reflect.go
shurcooL: Done.

compiler/natives/src/os/signal/signal.go
shurcooL: Done.

compiler/natives/src/testing/testing.go
shurcooL: Done.

circle.yml
shurcooL: As discussed, I'll run a short test for minification to cover it. Say, the `fmt` package:

```yml
- gopherjs test -v --minify fmt # Minification should work.
```

Done.

shurcooL:
Go 1.9 final release is out! I've updated this PR to target it. I'll be doing some final checks and merging this to `master` soon.

As usual, a `go1.8` branch will be created with the latest current `master`, for anyone who needs to build `GopherJS 1.8-x` with Go 1.8 support.

circle.yml
```diff
@@ -17,8 +17,10 @@ test:
     - diff -u <(echo -n) <(gofmt -d .)
     - go tool vet *.go                                                   # Go package in root directory.
     - for d in */; do echo $d; done | grep -v tests/ | grep -v third_party/ | xargs go tool vet # All subdirectories except "tests", "third_party".
+    - diff -u <(echo -n) <(go list ./compiler/natives/src/...) # All those packages should have // +build js.
+    - gopherjs install -v net/http # Should build successfully (can't run tests, since only client is supported).
     - >
-      gopherjs test --short --minify
+      gopherjs test -v --short
```
myitcv: Why drop the minify here? 

It actually helps to catch errors in the compiled Javascript (if changes are being made to the GopherJS compiler).

circle.yml
```diff
@@ -107,6 +109,7 @@ test:
       net/rpc/jsonrpc
       net/textproto
       net/url
+      os/user
```
myitcv: Would this list be better maintained as an exclusion list? Not that there are many new packages added to core... but still.

circle.yml
```diff
@@ -127,3 +130,4 @@ test:
       unicode/utf16
       unicode/utf8
     - go test -v -race ./...
+    - gopherjs test -v --minify fmt # Minification should work.
```
myitcv: See comment above.

Running with `--minify` on the standard library increases the test surface area for minified code which is a good thing.

circle.yml
shurcooL: I agree, that'd be better. Not as part of this PR, but let's make a follow up enhancement PR after this one is merged.

circle.yml
shurcooL: Explained in b827aaee653d5db1b538134a3d02433652a10b42 and discussed with @neelance at https://github.com/gopherjs/gopherjs/pull/651#discussion_r128371801, with the follow up done in 2587a1cd5553d8890d4f33cc76fe5eea82ebd7cc.

What are your thoughts after seeing that?

shurcooL:
I'm currently investigating the unsafe package becoming imported in https://github.com/gopherjs/gopherjs.github.io/issues/67#issuecomment-324966268. It's a change since GopherJS 1.8-2. It may be okay, or maybe it's suboptimal.

circle.yml
myitcv: @neelance makes the same point about the `--minify` flag essentially being a means of ensuring the generated Javascript code is valid in https://github.com/gopherjs/gopherjs/pull/651#discussion_r128371801

By only testing `--minify` with `fmt` we are assuming that we get complete coverage of generated output... I'm not sure I could easily reason that that's the case.

I take you point about the stack traces... if there's no particular time constraints we could do two test runs, one without `--minify` followed by one with. That way we get the best of both worlds.

Thoughts?

shurcooL:
> 

Found the root cause. It was suboptimal. I've restored the previous behavior (to avoid an unnecessary and unintended change of behavior) with commit 568295948925518f9dbeddd806482453f555d590.

Also rebased on top of latest master (it was 2 commits behind), in preparations of the final merge.

circle.yml
shurcooL: In b827aaee653d5db1b538134a3d02433652a10b42, I wrote:

> I don't see a strong reason to use minification during tests. Output
size is not a concern.

You said:

> By only testing `--minify` with `fmt` we are assuming that we get complete coverage of generated output... I'm not sure I could easily reason that that's the case.

I agree. I think that's a good reason to keep `--minify` flag.

It is, after all, the de-facto standard of how majority of production GopherJS code is generated, so testing it is more important than non-minified code. It's not worth sacrificing minified code test coverage that for the sake of convenience of having nicer CI error reports in the (rare) case of errors.

I'll revert that change. Thanks.

> I take you point about the stack traces... if there's no particular time constraints we could do two test runs, one without `--minify` followed by one with. That way we get the best of both worlds.

I don't quite like it because it'd mean you'd see CI build results slower. It's already quite slow at x minutes. Also, it'd feels like a waste of resources, given that the expected behavior is that `--minify` does not change behavior, and I would like that to be testable without running _all_ tests.

Let's leave this for later.

flimzy: > I don't quite like it because it'd mean you'd see CI build results slower.

It ought to be easy to run the non-minified tests only in the case of a failure of the minified tests, to pay that penalty only in the rare case of failure.

shurcooL: > It ought to be easy to run the non-minified tests only in the case of a failure of the minified tests, to pay that penalty only in the rare case of failure.

I was thinking that could be close to ideal behavior. But it doesn't seem very easy to me. If you think it's easy, it'd be helpful if you could open an issue and provide details of how to accomplish it.

myitcv: @shurcooL we can easily speed up the tests; they're effectively bound to a single core the way we run them. I've got some rather hacky code locally to runs `gopherjs test X` concurrently up to the number of cores. I'll try and polish that up and submit a PR. It works in much the same way as the runner in `tests/run.go`.

shurcooL:
The release has been delayed a bit because I found issues with type aliases in the last minute.

Using type aliases in a package main caused the compiler to panic, because the exporter it used didn't expect that a type declaration type could be anything other than `*types.Named`. Using exported type aliases in a library wasn't being exported/imported correctly either, so compiling a command that used an exported type alias from a library worked only when everything was compiled from scratch. But once a compiled archive was used on second load, the type alias symbol was missing. It wasn't caught by the type alias in `math/bits` package because it's not exported.

I've fixed the issue in be67827, and further simplified some type alias code in `compiler` package in 
61693e9. In my testing, type aliases work correctly now; at least I could not find any way to break them. Also, all tests are passing and I don't see any other issues after the fix.

I found it hard to write good tests for type aliases because the issues were relevant to the package export data, so it matters whether its a command/library, and whether the package is being compiled or an archive being loaded. Type alias tests will be helpful and needed, but I don't necessarily want to block the 1.9-1 release on them.

I believe everything should be ready for the final release, but because of the unexpected last minute fixes, I want to give some time for further testing. Also, review of those 2 commits would be highly appreciated.

compiler/package.go
```diff
@@ -468,39 +468,38 @@ func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, impor
 				c.Printf(`%s = $newType(%d, %s, "%s.%s", %t, "%s", %t, %s);`, lhs, size, typeKind(o.Type()), o.Pkg().Name(), o.Name(), o.Name() != "", o.Pkg().Path(), o.Exported(), constructor)
 			})
 			d.MethodListCode = c.CatchOutput(0, func() {
-				if _, isInterface := o.Type().Underlying().(*types.Interface); isInterface {
+				if o.IsAlias() {
```
neelance: Shouldn't this filter be already at a higher level, e.g. at line 436? As far as I can see, type aliases only exist at the source level, they should not occur in the compiler output any more.

myitcv: That's my understanding too.

shurcooL: Quite possibly. However, I can't answer with confidence, it would require further investigation.

I am quite confident in its current location because of my understanding of methods (a type alias may not have any additional methods; they're always associated with the aliased type instead), and also because I've inspected the generated code and confirmed it looked good.

One thing to find out, if this is moved higher level, is what happens regarding DCE when a type is only ever referenced via its type alias. E.g.:

```Go
package p

type T int

type A = T
```

```Go
package main

import "p"

func main() {
    var a p.A = 123
    fmt.Printf("%T %v\n", a, a)
}
```

myitcv: I wouldn't anticipate there being a problem in this situation, because everything is done based on the derived type of expressions. Aliases have no impact on that calculation because they've effectively already been erased.

myitcv: e.g.

```go
package main

import (
	"fmt"
	"go/ast"
	"go/importer"
	"go/parser"
	"go/token"
	"go/types"
	"log"
)

const hello = `package main

import "bytes"

type T = bytes.Buffer

var x T

func main() {
}
`

func main() {
	fset := token.NewFileSet()

	f, err := parser.ParseFile(fset, "hello.go", hello, 0)
	if err != nil {
		log.Fatal(err)
	}

	conf := types.Config{Importer: importer.Default()}

	pkg, err := conf.Check("cmd/hello", fset, []*ast.File{f}, nil)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("x: %s\n", pkg.Scope().Lookup("x").Type())
}
```

gives:

```
x: bytes.Buffer
```

hence the dependency on `bytes` is derived.

myitcv: Fixed typo in my closing sentence...

shurcooL: That sounds reasonable. I don't have objections to trying it out. The basic tests I've tried all pass.

neelance: Cool! I prefer this solution because it is less likely to hide bugs (e.g. some bad runtime reference to a type alias).

shurcooL:
GopherJS 1.9-1 is now released and available on `master` branch.

See blog post about it here: https://medium.com/gopherjs/gopherjs-1-9-1-is-released-e4d808e58f19.

See tweet here: https://twitter.com/GopherJS/status/901611150420344832.

Thanks everyone! 🎉

myitcv:
Just done a quick spot check; looks like my minimal [`Hello World` `myitcv.io/react` app](https://github.com/myitcv/react/wiki/Examples) has grown in size by `~20K` with Go 1.9. Very rough numbers, just noting for now; full investigation to follow.

shurcooL:
If you don't tell us its absolute size, it's not possible to know whether 20 KB is a significant change.

Let's continue generated code size discussion in #136. It's normal and unavoidable to see an increase after an upgrade to a new release of Go due to the upstream issue https://github.com/golang/go/issues/6853.

myitcv:
Sorry that's my bad; 20KB on top of ~740KB.

But you're right; this time it's almost certainly a result of upstream changes.

We don't track size at all as part of the CI build; I think it would be worth adding some sort of output so we can trace things over time, if only to spot significant changes as a result of non upstream changes.

Issue #661, opened 2017-07-26
Struct values not being cloned when assigned from/called with when variable type is interface


myitcv:
_(had a look, couldn't recall this having been discussed or raised before)_

Consider the following code:

```go
package main

import (
	"fmt"
)

type S struct {
	Name string
}

func (s S) SetName(n string) {
	s.Name = n
}

type NameSet interface {
	SetName(n string)
}

func main() {
	s1 := S{}
	var s2 NameSet = s1
	s1.Name = "Rob"
	
	fmt.Printf("s1: %v, s2: %v\n", s1, s2)
	
	s2.SetName("Pike")
	fmt.Printf("s1: %v, s2: %v\n", s1, s2)
}
```

Per [the Go Playground](https://play.golang.org/p/LNldeawtHJ), this should output:

```
s1: {Rob}, s2: {}
s1: {Rob}, s2: {}
```

But per [the GopherJS Playground](https://gopherjs.github.io/playground/#/qWZCc7PNyD) we get:

```
s1: {Rob}, s2: {Rob} 
s1: {Pike}, s2: {Pike} 
```

Reason being the assignment to `s2` is not [`$clone`](https://github.com/gopherjs/gopherjs/blob/96e2b0e61b742f88faa02813d1b1c657d9a7f798/compiler/prelude/prelude.go#L290)-ing the value from `s1`. 

Similarly, in the call `s2.SetName` we are not `$clone`-ing.

If the type of `s2` is changed to `S` then everything works as expected.

Will look into this a bit more later, just logging here for now.

shurcooL:
Thanks for reporting, that looks like a legitimate bug related to JavaScript references vs Go values/pointers.

myitcv:
I've pushed up https://github.com/gopherjs/gopherjs/pull/669 as an attempt to fix this bug.

@neelance @shurcooL - would very much appreciate feedback and comments.

PR #669, opened 2017-07-31
Investigate/fix struct val bug

myitcv:
**!! DO NOT REVIEW !!**

Investigate/fix struct val bug reported in #661


compiler/expressions.go
```diff
@@ -1134,7 +1139,7 @@ func (c *funcContext) translateImplicitConversion(expr ast.Expr, desiredType typ
 			return c.formatExpr("new %s(%e)", c.typeName(exprType), expr)
 		}
 		if _, isStruct := exprType.Underlying().(*types.Struct); isStruct {
-			return c.formatExpr("new %1e.constructor.elem(%1e)", expr)
+			return c.formatExpr("new %1e.constructor.elem($clone(%1e, %s))", expr, c.typeName(exprType))
```
myitcv: Address case 1

compiler/expressions.go
```diff
@@ -1113,7 +1113,12 @@ func (c *funcContext) translateImplicitConversion(expr ast.Expr, desiredType typ
 
 	exprType := c.p.TypeOf(expr)
 	if types.Identical(exprType, desiredType) {
-		return c.translateExpr(expr)
+		switch exprType.Underlying().(type) {
```
myitcv: Address case 2

compiler/prelude/prelude.go
```diff
@@ -293,6 +293,14 @@ var $clone = function(src, type) {
   return clone;
 };
 
+var $copyIntf = function(src) {
```
myitcv: Part of addressing case 2

myitcv:
Hmm seems I've broken the tests... will investigate tomorrow.

myitcv:
Added support for array values.

myitcv:
Ok, I think everything is fixed up now. CircleCI shows the test for `text/template` failed, but it passes locally. So not entirely clear what's going on there...

compiler/prelude/prelude.go
```diff
@@ -293,6 +293,14 @@ var $clone = function(src, type) {
...
+	if (src.constructor.copy) {
+		return new src.constructor($clone(src.$val, src.constructor));
+	} else {
+		return src;
+	}
```
flimzy: No need for this `else`, since it's preceded by a `return`.

myitcv: 👍 - fixed.

circle.yml
```diff
@@ -18,7 +18,7 @@ test:
     - go tool vet *.go                                                   # Go package in root directory.
     - for d in */; do echo $d; done | grep -v tests/ | grep -v third_party/ | xargs go tool vet # All subdirectories except "tests", "third_party".
     - >
-      gopherjs test --short --minify
+      gopherjs test --minify
```
myitcv: Temp change to debug failure on CI server...

myitcv:
Ok I've narrowed down the test failures in `text/template` to `TestMaxExecDepth`.

```
$ rm -rf /home/myitcv/gostuff/pkg/linux_js* && gopherjs test -v --short --minify text/template
...
=== RUN   TestMissingFieldOnNil
--- PASS: TestMissingFieldOnNil (0.00s)
=== RUN   TestMaxExecDepth
FAIL    text/template   1.528s
```

The test sometimes passes locally for me, sometimes not (`node v8.2.1` or indeed `node v6.2.2`, the version running on CI). It seems to consistently fail on the CI server.

When it fails it always fails after the line:

```
=== RUN   TestMaxExecDepth
```

I also get a core dump when it fails.

@shurcooL has this been an issue before?

myitcv:
Having just said it fails consistently on the CI server of course it just passed:

https://circleci.com/gh/gopherjs/gopherjs/1189

I then pushed up a commit to try with a later version of node:

https://circleci.com/gh/gopherjs/gopherjs/1190

... and that passed too.

So I'm at something of a loss 😢

myitcv:
Ok, well on the assumption that it's some form of memory issue (unclear why it's so random) I've switched to using the latest Node version, as well as increasing the amount of memory available to the test run by passing in a V8 parameter via `NODE_OPTIONS="--max-old-space-size=4096"`

circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 6.2.2
+    version: 8.2.1
```
myitcv: Upgrade Node version and pass in V8 option to increase available memory

compiler/expressions.go
```diff
@@ -1112,6 +1112,12 @@ func (c *funcContext) translateImplicitConversion(expr ast.Expr, desiredType typ
 	}
 
 	exprType := c.p.TypeOf(expr)
+
+	// whenever expr has an underlying *types.Interface type we need to $copyIntf
+	if _, isIntf := exprType.Underlying().(*types.Interface); isIntf {
```
myitcv: Address case 2.

compiler/expressions.go
```diff
@@ -1130,12 +1136,17 @@ func (c *funcContext) translateImplicitConversion(expr ast.Expr, desiredType typ
 			// wrap JS object into js.Object struct when converting to interface
 			return c.formatExpr("new $jsObjectPtr(%e)", expr)
 		}
+
+		switch exprType.Underlying().(type) {
```
myitcv: Address case 1.

myitcv:
Raised https://github.com/nodejs/node/issues/14228 to see if there is anything from the Node side they can help with.

circle.yml
```diff
@@ -18,7 +19,7 @@ test:
     - go tool vet *.go                                                   # Go package in root directory.
     - for d in */; do echo $d; done | grep -v tests/ | grep -v third_party/ | xargs go tool vet # All subdirectories except "tests", "third_party".
     - >
-      gopherjs test --short --minify
+      ulimit -s $GOPHERJS_STACK_SIZE && ulimit -s && gopherjs test --short --minify
```
myitcv: Per https://github.com/nodejs/node/issues/14567#issuecomment-319367412, I've not unified the value of `ulimit -s` and the `--stack_size=` flag passed to node.

myitcv:
Per the update to the linked node issue, I think the random failures have been resolved by the `ulimit -s` and `--stack_size` synchronisation changes I've also made part of this PR.

So we're back to a passing build and I now expect this built to pass reliably.

circle.yml
shurcooL: Does this change need to be a part of this PR?

If not, can you factor it out? It's easier to deal with 2 smaller independent PRs.

myitcv:
@shurcooL per your question about splitting the Node upgrade into a separate PR (which I can't find for some reason). 

Having just completed a test run using `v6.2.2` (current `master` version), i.e. relying only on the harmonisation of `ulimit -s` and `--stack_size`, it seems we _don't_ need to upgrade Node to `v8.2.1`. So I've reverted that change from this PR.

shurcooL:
(The comment is likely gone because I left it on a line that has since been removed from the PR; GitHub doesn't seem to handle that well.)

I meant that change and all the `GOPHERJS_STACK_SIZE` stuff. Is it needed to fix #661?

myitcv:
> I meant that change to Node version and all the GOPHERJS_STACK_SIZE and ulimit stuff. Is it needed to fix #661?

The changes to harmonize `ulimit -s` and `--stack_size` are required for the changes (at least as currently implemented in this PR) to fix #661. For some reason the changes to fix #661 in this PR cause the `TestMaxExecDepth` test in `text/template` to randomly break the stack limit as discussed in https://github.com/nodejs/node/issues/14567 which then fails the tests.

My guess is that @neelance introduced `--stack_size` in the first place precisely for this reason, i.e. to handle situations where we intentionally recurse (as the test does).  However, as advised in https://github.com/nodejs/node/issues/14567#issuecomment-319367412, we need to harmonise `ulimit -s` and `--stack_size`. 

We can do that in a separate PR, just so long as that PR goes in before this one (assuming we're otherwise happy with the changes in this PR). 

Would you like me to create a separate PR for the `ulimit -s` and `--stack_size` harmonisation changes?

shurcooL:
No need, if it's needed to resolve the issue, then it belongs here. I wasn't sure that was the case.

shurcooL:
I've read https://github.com/nodejs/node/issues/14567. My question/suggestion, can we isolate that issue to CI only, with a comment to that issue?

I'm not happy with the complexity it adds, so the smaller surface area it has, the better IMO. I don't want all GopherJS users to have to understand the new `GOPHERJS_STACK_SIZE` variable, I don't want us to have to document it, I don't want us to maintain it, etc., if it's at all avoidable.

If I understand correctly, it's something we need to fix intermittent CI failures related to `node`?

If so, can we simplify it to just doing this hard-coded hack somewhere in the Circle CI script:

```
test:
   override:
    - ulimit -s 10000 # Fix for https://github.com/nodejs/node/issues/14567.
    ... the rest of commands, unmodified
```

Would that work, or is it not possible?

shurcooL:
I see you tried something like that in f24a681bb3a8029d078355754b7a880443f732f6, why didn't it work?

shurcooL:
Can you please link me to some failures that result from not setting `ulimit -s` to 10000?

shurcooL:
Also, do you have any theories or ideas why this stack size comes up only on this PR, but not on master? Or does it also happen on master, just rarely?

I'm really concerned about it and want to have better understanding and find the best and simplest solution to the problem, so thanks for your patience.

myitcv:
> Can you please link me to some failures that result from not setting ulimit -s to 10000?

https://circleci.com/gh/gopherjs/gopherjs/1188

is one such example.

But I can reliably cause local failures if `ulimit -s < 10000` (for me the default `ulimit -s` value is `8192`):

```bash
ulimit -s 8192
for i in $(seq 1 100)
do
  echo $i
  gopherjs test --minify --short text/template || break
done
```

Just ran that locally at it failed after 4 iterations.

> Would that work, or is it not possible?

If we look at the default value of `--stack_size` (via `node --v8-options`):

```
  --stack_size (default size of stack region v8 is allowed to use (in kBytes))
        type: int  default: 984
```

https://github.com/gopherjs/gopherjs/commit/1f895455a6b93229254433b53142be00b649fa0c introduced the change in the first place, to handle situations where deeply recursive Go code resulted in deeply recursive Javascript code to the point that this stack space was exceeded. `text/template`'s `TestMaxExecDepth` is one such test that intentionally recurses.

Indeed if you remove `--stack_size` from `tool.go` you quickly get error messages of the sort:

```
--- FAIL: TestMaxExecDepth (0.01s)

/home/myitcv/gostuff/src/github.com/gopherjs/gopherjs/test.716549077:37009
        state.ptr.prototype.evalPipeline = function(dot, pipe) {
                                            ^
RangeError: Maximum call stack size exceeded
    at $packages.text/template.state.ptr.evalPipeline (/home/myitcv/gostuff/src/github.com/gopherjs/gopherjs/test.716549077:37009:45)
    at $packages.text/template.state.ptr.walkTemplate (/text/template/exec.go:383:3)
    at $packages.text/template.state.ptr.walk (/text/template/exec.go:245:4)
    at $packages.text/template.state.ptr.walk (/text/template/exec.go:240:5)
...
```

So we need to increase `--stack_size` for the standard library tests, at least.

But as pointed out in https://github.com/nodejs/node/issues/14567#issuecomment-319363833, the value of `--stack_size` simply tells V8 to assume there is that much stack space, as well as acting as a limit itself (as demonstrated above). `ulimit -s` is the ultimate OS limit on the V8 process. 

The segfaults come about because the OS limit is breached by a process that assumes it can allocate more space (because `--stack_size > ulimit -s` on the CI server and my local machine). 

> I'm really concerned about it and want to have better understanding and find the best and simplest solution to the problem, so thanks for your patience.

I'm not too concerned. This deeply recursive code is not the norm. This would simply explode in the browser.

With a default `ulimit -s` value of `8192` my local machine fails after (roughly) on average 20 runs of `gopherjs test text/template`. On the CI server it fails more reliably, but it did pass once: 

https://circleci.com/gh/gopherjs/gopherjs/1189

A rough guess at what's going on here is that as a result of the increased "cloning" of struct/array values, in a deeply recursive situation we run out of stack space more quickly. 

Furthermore, I suspect it was more by luck than judgement that we didn't run into this before now. i.e. the `text/template` tests have probably taken us close to the `8192` limit. 

Trying to cut down on the amount of stack space consumed is a sensible but significantly bigger goal, well beyond the scope of this PR.

> I'm really concerned about it and want to have better understanding and find the best and simplest solution to the problem, so thanks for your patience.

Couldn't agree more. Having written this reply, I actually think the approach I've taken as of right now is wrong. Specifically the "interface" via the `GOPHERJS_STACK_SIZE` environment variable is wrong.

With `gopherjs test` and `gopherjs run` we are effectively wrapping a call to Node. 

As pointed out above we need to set `--stack_size` at least when running the standard library tests (because the default value is too small) but I don't think we want to start exposing flags like this.

So what if instead we set the value of `--stack_size` to be the value of `ulimit -s`. That way the "interface" to controlling the value we pass to Node is defined by existing OS commands - no flag required.

_In this context I note: http://www.cs.nyu.edu/exact/core/doc/stackOverflow.txt (not sure how we handle windows properly)_

So the relevant part of `tool.go` would look something like:

```go
if runtime.GOOS != "windows" {
	r := syscall.Rlimit{}

	err := syscall.Getrlimit(syscall.RLIMIT_STACK, &r)
	if err != nil {
		panic(err)
	}

	stackSize := fmt.Sprintf("--stack_size=%v", r.Cur>>10)

	allArgs = append(allArgs, stackSize, script)
}
```

The onus is then on the developer (who uses `gopherjs test` or `gopherjs run`) to have an appropriate value for `ulimit -s`; we no longer need to worry about exposing Node/V8 flags.

Thoughts?

shurcooL:
Thanks for the detailed analysis.

About the proposed change, I'll take more time to think about it, but my initial reaction is that it sounds much better.

myitcv:
> About the proposed change, I'll take more time to think about it, but my initial reaction is that it sounds much better.

👍  - if anything I think we should put the addition of a `--stack_size` flag in a function behind a built constraint. Unix systems can follow the above approach - we can then choose how to handle other systems if/when we run into the issue.

myitcv:
@shurcooL - I've now pushed up the limit-based setting of `--stack_size` as a separate commit.

Let me know what you think.

Once we're happy with the result we can squash down the commits.

tool.go
```diff
@@ -750,7 +750,17 @@ func runNode(script string, args []string, dir string, quiet bool) error {
 	}
 
 	if runtime.GOOS != "windows" {
-		allArgs = append(allArgs, "--stack_size=10000", script)
+		r := syscall.Rlimit{}
```
shurcooL: Since you don't care about initial value (i.e., you want zero, it will be populated in next line), use `var r syscall.Rlimit` form. It's slightly more common in this type of situation.

tool.go
```diff
@@ -750,7 +750,17 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+
```
shurcooL: Remove blank line.

tool.go
```diff
@@ -750,7 +750,17 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		err := syscall.Getrlimit(syscall.RLIMIT_STACK, &r)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "gopherjs: failed to get stack size limit: %v", err)
```
shurcooL: You're missing a blank line at the end of the string.

tool.go
```diff
@@ -750,7 +750,17 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+			os.Exit(1)
```
shurcooL: What would happen if you did `return err` instead of handling the error right here?

tool.go
```diff
@@ -750,7 +750,17 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		}
+
```
shurcooL: Remove blank line. It's not helpful.

tool.go
```diff
@@ -750,7 +750,17 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		stackSize := fmt.Sprintf("--stack_size=%v", r.Cur>>10)
+
```
shurcooL: Remove blank line.

tool.go
```diff
@@ -750,7 +750,17 @@ func runNode(script string, args []string, dir string, quiet bool) error {
 	}
 
 	if runtime.GOOS != "windows" {
-		allArgs = append(allArgs, "--stack_size=10000", script)
```
shurcooL: Add a comment explaining this section. Doesn't have to be long.

tool.go
myitcv: Returning an error in this case would appear to enter the flow of `handleError`, `printError` (which also handles the newline where required) - I'm not so familiar with how you look to handle this sort of "system" error vs. compile errors, etc. Any thoughts on what you'd prefer?

shurcooL: I don't see evidence that we must treat this error differently than another error in this scope. So, doing `return fmt.Errorf(...)` seems better to me, if it works.

Basically, I see doing this `syscall.Getrlimit` call as just another of the steps of the `runNode` process. If executing `node` fails, we return error. Why should this be different?

tool.go
myitcv: Done.

myitcv:
@shurcooL - addressed the points you raised.

tool.go
```diff
@@ -750,7 +750,28 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		// for non-windows OS environments, we've seen issues with stack space
```
shurcooL: Capitalize first letter of first word of a sentence.

tool.go
```diff
@@ -750,7 +750,28 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		// limits causeing Go std library tests that are recursion-heavy to fail
+		// (see https://github.com/gopherjs/gopherjs/issues/661 for more detail).
+		//
+		// There are two limits that come into play here, listed in order:
+		//
+		// 1. V8 limit (NodeJS effectively wraps V8) 2. OS process limit
```
shurcooL: I think it'd be more readable to have them on separate lines:

```
// 1. V8 limit (NodeJS effectively wraps V8)
// 2. OS process limit
```

tool.go
```diff
@@ -750,7 +750,28 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		// limit 2. That is to say, whatever value is returned by ulimit -s is
+		// essentially the value that we pass on to NodeJS via the appropriate V8
+		// flag.
+		var r syscall.Rlimit
+		err := syscall.Getrlimit(syscall.RLIMIT_STACK, &r)
+		if err != nil {
+			return fmt.Errorf("failed to get stack size limit: %v", err)
+		}
+		// rlimit value is in bytes, we need rounded kBytes value per node
+		// --v8-options
```
shurcooL: I think you can move "--v8-options" back on previous line, to keep the comment 1 line. It's not worth making it 2 lines just because of 1 word.

Also, this sentence is missing a period at the end.

tool.go
```diff
@@ -750,7 +750,28 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		stackSize := fmt.Sprintf("--stack_size=%v", r.Cur/1000)
+		allArgs = append(allArgs, stackSize, script)
```
shurcooL: Now that you have the `// rlimit value is in bytes, we need rounded kBytes value per node --v8-options.` comment above, it would help to have a blank line separating these. Otherwise, it reads as if that comment applies to the append line, but it doesn't.

Also, I guess we should fix the issue mentioned at https://github.com/gopherjs/gopherjs/commit/40d8f0402d9c5ed59bc00f14c7b686cab8cc8414#commitcomment-23559628.

So:

```Go
		...
	}
	// rlimit value is in bytes, we need rounded kBytes value per node --v8-options.
	stackSize := fmt.Sprintf("--stack_size=%v", r.Cur/1000)

	allArgs = append(allArgs, stackSize)
}

allArgs = append(allArgs, script)
...
```

compiler/prelude/prelude.go
shurcooL: I've never seen interfaces shortened to "intf". Instead of creating a novel way, let's reuse the same abbreviation this codebase (and Go codebase) already contains: "iface".

(https://github.com/golang/go/search?q=iface vs https://github.com/golang/go/search?q=intf)

Please update all places in this PR, not just this line.

tool.go
myitcv: Sorry, that's my bad for not having caught this. It's auto-formatting in vim gone wrong.

tool.go
myitcv: Again, this my auto-formatting of comments kicking in.

compiler/prelude/prelude.go
myitcv: Fair enough - I'd be using [`msgp`](http://godoc.org/github.com/tinylib/msgp/msgp#AppendIntf) as my pattern here. 

Is there actually any reason to abbreviate here?

We could simply name this function the more precise `$copyInterfaceVal`

tests/interface_test.go
```diff
@@ -0,0 +1,88 @@
...
+	if i1s != iexp {
+		t.Fatalf("is should have been %q; got %q", iexp, i1s)
+	}
+
+	if i2s != iexp {
+		t.Fatalf("is should have been %q; got %q", iexp, i2s)
+	}
+}
+
+func TestStructValIntfMethodCall(t *testing.T) {
```
shurcooL: /s/Intf/Iface/

compiler/expressions.go
```diff
@@ -1112,6 +1112,12 @@ func (c *funcContext) translateImplicitConversion(expr ast.Expr, desiredType typ
 	}
 
 	exprType := c.p.TypeOf(expr)
+
+	// whenever expr has an underlying *types.Interface type we need to $copyInterfaceVal
+	if _, isIntf := exprType.Underlying().(*types.Interface); isIntf {
```
shurcooL: /s/isIntf/isIface/

myitcv:
@shurcooL - apologies. Should be good now

compiler/expressions.go
```diff
@@ -1112,6 +1112,12 @@ func (c *funcContext) translateImplicitConversion(expr ast.Expr, desiredType typ
 	}
 
 	exprType := c.p.TypeOf(expr)
+
+	// whenever expr has an underlying *types.Interface type we need to $copyInterfaceVal
+	if _, isiface := exprType.Underlying().(*types.Interface); isiface {
```
myitcv: I'm clearly having a bad day :(

shurcooL:
Thanks. This addresses all of my initial review feedback, which covered `tool.go` and style comments.

Next up the actual meat of the PR, the changes to compiler to fix #661, need to be reviewed. I'm not sure I'm qualified to review that because I'm not expert at compiler internals, so ideally I'd want @neelance to review that. (Otherwise, I'll need to spend more time getting to understand the change. I won't have time for that today)

I can help out by running benchmarks to try to catch it if the performance regressed significantly as a result of this change.

myitcv:
@shurcooL - thanks, appreciate your patience with the couple of bad fixes that crept in 👍 

@neelance - happy to chat through my thinking/reasoning if that helps to get you into context on this.

compiler/expressions.go
```diff
@@ -1112,6 +1112,12 @@ func (c *funcContext) translateImplicitConversion(expr ast.Expr, desiredType typ
 	}
 
 	exprType := c.p.TypeOf(expr)
+
+	// whenever expr has an underlying *types.Interface type we need to $copyInterfaceVal
+	if _, isIface := exprType.Underlying().(*types.Interface); isIface {
```
neelance: I don't think this is the right place for this code. As far as I can see, the only place it is needed is for the receiver when calling a method on an interface (your test case). Thus, this code should be at a place where it only handles the receiver. I even think that when moving the code there, you can get rid of the `$copyInterfaceVal` helper.

compiler/expressions.go
neelance: Can this maybe be solved instead by making `translateImplicitConversionWithCloning` look at `c.p.TypeOf(expr)` instead of `desiredType`? This would also catch the case of "array/struct to interface".

compiler/expressions.go
myitcv: > As far as I can see, the only place it is needed is for the receiver when calling a method on an interface (your test case)

This was my initial understanding as well. But I put the code here because I wasn't sure whether I'd considered all cases. And furthermore I reasoned to myself that whilst there may be unnecessary copying going on, everything would still be "correct" (I actually now think that conclusion on correctness is, ironically, incorrect, but it's now something of a moot point).

Having just given it some more thought I think we can, as you suggest, handle this via the receiver, save ourselves lots of unnecessary copying.

I'll do so in my next commit.

compiler/expressions.go
myitcv: As I understand it, we effectively need some combination of the logic from `translateImplicitConversionWithCloning` and `translateImplicitConversion` when it comes to array/struct values.

As this PR stands:

* `translateImplicitConversionWithCloning` handles the case where the target is an array/struct type which by definition means the source must be exactly the same type, hence a `$clone` is required
* `translateImplicitConversion` handles the conversion from an array/struct type to an interface type, hence the `new X` is required to wrap the `$clone`-ed value (part of #661 concerned the fact that we weren't cloning before wrapping)

I'm relatively relaxed on where the code lives because once your first comment is addressed I think this PR looks simpler as far as `translateImplicitConversion` is concerned in any case.

Let me know what you think

myitcv:
@neelance - thanks for your comments. I've added a couple of responses. Suspect I won't have a chance to do anything further on the PR this week. So in case you had any thoughts on my responses that would be great else I'll ping you when I've addressed the first of your comments.

compiler/expressions.go
neelance: Cool. Just a small remark:

> I wasn't sure whether I'd considered all cases

Imho don't let that bother you. In my experience it is better to aim for the simplest code that handles the cases you know.

shurcooL: (Highly agreed with @neelance's remark. Adding unused code IMO is much worse than not handling some cases. Unhandled cases are obvious to spot when it happens, but unused code is not obvious and will confuse readers, adds complexity, needs to be maintained, etc.)

myitcv:
Thanks @neelance @shurcooL for the feedback.

I've now pushed up a new version that makes specific the use of `$copyInterfaceVal` per your comments @neelance. 

To pick up your point that you thought we might be able to do away with the definition of `$copyInterfaceVal` altogether, I can't see how this would be possible (but I could quite easily be missing something obvious). By definition we don't know until runtime the type of the value pointed to by the interface, so only at runtime will we know whether that value can (and should therefore) be `$clone`-d or not. That logic seemed best extracted into a function. Unless, as I say, I'm missing something...

shurcooL:
I've seen this failure for the first time recently when working on `go1.9` branch:

```
=== RUN   TestMaxExecDepth
FAIL	text/template	2.315s
```

It has happened 2 times out of 10~20 CI runs so far.

So that is validation for the stack change in this PR.

At this point, I think it'd be better to factor it out into a separate PR (after GopherJS 1.9-1 release). That way, it can be merged (and reverted, if neccessary) independently. Only if it's not too much trouble.

myitcv:
> At this point, I think it'd be better to factor it out into a separate PR (after GopherJS 1.9-1 release). 

Happy to. I'll do exactly that once 1.9 lands.

myitcv:
@shurcooL #687 raised for the separate `ulimit -s` and `--stack_size` changes.

Hence the diff as part of this PR (which has also been rebased following the merge of the Go1.9 changes) should be viewed as if #687 has already been merged.

@neelance any other changes you want to see as part of this PR?

Thanks

myitcv:
@neelance - can you take a final look at this please?

This issue came up again in the context of #783

neelance:
I see that you addressed my comments where possible and I've looked at the current changes and they seem okay to me, since they only modify very specific cases now. However, since I haven't worked on GopherJS for a while, I don't have all the context in my head any more and it would take quite a while to properly assess the PR again. I've removed my review since I have no reason to block the merge any more, but it would be good if @shurcooL would give the final approval.

myitcv:
Thanks @neelance 

@shurcooL - please let me know if I can provide any background etc on this to help with the review.

myitcv:
Gentle ping @shurcooL

shurcooL:
Thanks for ping. I'll try to find time to look at this and the other PR this weekend.

Sorry, I am unfortunately still in a more-overloaded-than-usual state and that's why I can't provide timely reviews of complex changes. :( I should be out of this state sometime within the next few weeks.

shurcooL:
Thanks for working on this bug, and sorry it's taking so long. I've found some time and started reviewing this now. I will need more time, but I found something that I haven't seen mentioned so far. The fix so far works for a value assigned to an interface variable:

```Go
s1 := S{}
var s2 NameSet = s1
s1.Name = "Rob"
fmt.Printf("s1: %v, s2: %v\n", s1, s2)
s2.SetName("Pike")
fmt.Printf("s1: %v, s2: %v\n", s1, s2)
```

But it has no effect when a pointer is assigned to the interface variable:

```Go
s1 := S{}
var s2 NameSet = &s1
s1.Name = "Rob"
fmt.Printf("s1: %v, s2: %v\n", s1, s2)
s2.SetName("Pike")
fmt.Printf("s1: %v, s2: %v\n", s1, s2)
```

That code continues to produce incorrect output with GopherJS (unchanged on master vs this PR):

```
s1: {Rob}, s2: &{Rob}
s1: {Pike}, s2: &{Pike}
```

Instead of the expected:

```
s1: {Rob}, s2: &{Rob}
s1: {Rob}, s2: &{Rob}
```

First, I want to find out if you consider fixing that case in scope of this issue and/or PR, or if you'd rather leave the scope for fixing the existing interface value cases covered by the tests?

It might be that the fix for both may end up simpler, but maybe it's not as closely related as it might seem. Let me know how you'd prefer to proceed. Thanks again.

myitcv:
@shurcooL - very good catch, thank you!

Digging in to this today, it's opened up a whole can of worms. I'm going to have to park looking at this further for now, but this example shows some of the worms involved ([playground](https://gopherjs.github.io/playground/#/G1yiHLilNt) - open developer console): 😄 

```go
package main

import (
)

type S struct {}

func main() {
	s := S{}
	js.Global.Get("console").Call("log", js.InternalObject(s).Get("constructor").Get("string"))
}
```

gives the output:

```
*main.S
```

That said, I think I see a path to properly solving #661 that also untangles some of those worms.

PR #681, opened 2017-08-19
Ensure that sorting of variables is stable

### LINKED ISSUE #679: Generated Javascript is non-deterministic for functions that contain multiple variables with the same name
I created the following test file:

```go
package main

import (
  "reflect"

)

func main() {
  js.Global.Set("tester", map[string]interface{}{
    "Tester": tester,
  })
}

func tester(input string) string {
  reflect.DeepEqual(nil, nil)
  return input
}
```

I would like the Javascript that is generated to be constant across platforms. As such, I defined the following `go generate`:

```bash
GOOS=darwin gopherjs build main.go -o go.js
```

However, the output of the generated Javascript for the `reflect` package is subtly different. When run on a Darwin system versus a Linux system (tested in a Docker image).

Specifically, the code generated by Linux has extra `$1` characters. Examples:

```
-  typ[0] = tt.elem;
+  typ$1[0] = tt.elem;
```

```
-  return wrapJsObject(typ[0], a[0][i[0]]);
+  return wrapJsObject(typ$1[0], a$1[0][i[0]]);
```

For now I can work around this by requiring the generation to be done in a Docker container so that the results are always consistent, but it would be much nicer if the output could be guaranteed to be stable/consistent for a given `GOOS` regardless of the host environment.

nmiyake:
If names are equal, sort ties by position.

Fixes #679

### DIFF
--- a/compiler/utils.go
+++ b/compiler/utils.go
@@ -369,7 +369,12 @@ func (c *funcContext) handleEscapingVars(n ast.Node) {
 
 	var names []string
 	objs := analysis.EscapingObjects(n, c.p.Info.Info)
-	sort.Sort(varsByName(objs))
+	sort.Slice(objs, func(i, j int) bool {
+		if objs[i].Name() == objs[j].Name() {
+			return objs[i].Pos() < objs[j].Pos()
+		}
+		return objs[i].Name() < objs[j].Name()
+	})
 	for _, obj := range objs {
 		names = append(names, c.objectName(obj))
 		c.p.escapingVars[obj] = true
@@ -637,18 +642,4 @@ func endsWithReturn(stmts []ast.Stmt) bool {
 
 func encodeIdent(name string) string {
 	return strings.Replace(url.QueryEscape(name), "%", "$", -1)
-}
-
-type varsByName []*types.Var
-
-func (s varsByName) Len() int {
-	return len(s)
-}
-
-func (s varsByName) Swap(i, j int) {
-	s[i], s[j] = s[j], s[i]
-}
-
-func (s varsByName) Less(i, j int) bool {
-	return s[i].Name() < s[j].Name()
 }


compiler/utils.go
```diff
@@ -650,5 +650,8 @@ func (s varsByName) Swap(i, j int) {
 }
 
 func (s varsByName) Less(i, j int) bool {
-	return s[i].Name() < s[j].Name()
+	if s[i].Name() != s[j].Name() {
+		return s[i].Name() < s[j].Name()
+	}
+	return s[i].Pos() < s[j].Pos()
```
shurcooL: More often than not, I think names will differ. Only when they don't, do you want to sort by pos. Therefore, I think it'd be more readable to make the less frequent case indented, by flipping them:

```Go
if s[i].Name() == s[j].Name() {
	return s[i].Pos() < s[j].Pos()
}
return s[i].Name() < s[j].Name()
```

This is equivalent. Just a style suggestion.

compiler/utils.go
```diff
@@ -650,5 +650,8 @@ func (s varsByName) Swap(i, j int) {
 }
 
 func (s varsByName) Less(i, j int) bool {
```
shurcooL: `varsByName` is being used in exactly one place in this package. We're requiring Go 1.8, that means we have [`sort.Slice`](https://godoc.org/sort#Slice) available to us. Let's use it.

It will bring this code closer to where it's being used, making it more readable in the relevant context.

nmiyake:
Thanks -- updated PR based on suggestions.

shurcooL:
LGTM.

I tested it, it helps improve reproducibility of generated code. It hasn't resolved the issue fully for me, but I can see improvement.

Compare [before](https://user-images.githubusercontent.com/1924134/29483647-c044f88a-8478-11e7-848b-b2c71fff6f6f.png) vs [after](https://user-images.githubusercontent.com/1924134/29483655-f4a253d4-8478-11e7-8028-43bf7d8457f8.png). It seems the only remaining issue for me after this PR is the order of packages declared in `$packages["main"]`.

Thanks a lot for the fix!

shurcooL:
I've tested some local changes, and I found that if I make the following changes to `compiler.WritePkgCode`:

```diff
diff --git a/compiler/compiler.go b/compiler/compiler.go
index 460293d..d5e21e5 100644
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -9,6 +9,7 @@ import (
 	"go/token"
 	"go/types"
 	"io"
+	"sort"
 	"strings"
 
 	"github.com/gopherjs/gopherjs/compiler/prelude"
@@ -190,8 +191,12 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, minify bool, w
 	}
 	vars := []string{"$pkg = {}", "$init"}
 	var filteredDecls []*Decl
+	sort.Slice(pkg.Declarations, func(i, j int) bool {
+		return pkg.Declarations[i].FullName < pkg.Declarations[j].FullName
+	})
 	for _, d := range pkg.Declarations {
 		if _, ok := dceSelection[d]; ok {
+			sort.Strings(d.Vars)
 			vars = append(vars, d.Vars...)
 			filteredDecls = append(filteredDecls, d)
 		}
```

Then the generated output is completely deterministic for me! Very exciting!

nmiyake:
Awesome, that's great to hear! If the determinism is resolved and the vendor issue is resolved as well, then I think that will be sufficient to guarantee completely self-contained and consistent generated output for a given revision of a product, which is great!

shurcooL:
@neelance Do you have a chance to take a look at this? It's a small change, but I wanted to make sure you see it before we merge in case you have some thoughts. You've worked on related changes in the past, so it's FYI as well.

markcampanelli-wf:
@shurcooL Any followup implementation of your suggestion above for further improving determinism?

shurcooL:
I tried to look into it briefly, and had the following changes locally for investigation:

```diff
diff --git a/compiler/compiler.go b/compiler/compiler.go
index ad3af54..203bb96 100644
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -13,6 +13,8 @@ import (
 
 	"github.com/gopherjs/gopherjs/compiler/prelude"
 	"golang.org/x/tools/go/gcimporter15"
+
+	"github.com/shurcooL/go-goon"
 )
 
 var sizes32 = &types.StdSizes{WordSize: 4, MaxAlign: 8}
@@ -190,12 +192,18 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, minify bool, w
 	}
 	vars := []string{"$pkg = {}", "$init"}
 	var filteredDecls []*Decl
+	//sort.Slice(pkg.Declarations, func(i, j int) bool {
+	//	return pkg.Declarations[i].FullName < pkg.Declarations[j].FullName
+	//})
 	for _, d := range pkg.Declarations {
 		if _, ok := dceSelection[d]; ok {
 			vars = append(vars, d.Vars...)
 			filteredDecls = append(filteredDecls, d)
 		}
 	}
+	goon.DumpExpr(vars)
+	//sort.Strings(vars)
+	//goon.DumpExpr(vars)
 	if _, err := w.Write(removeWhitespace([]byte(fmt.Sprintf("\tvar %s;\n", strings.Join(vars, ", "))), minify)); err != nil {
 		return err
 	}
```

However, I didn't arrive to definitely results and knew it needed further investigation. I didn't have time, so I stashed the changes. I haven't had a change to revisit this issue since.

PR #683, opened 2017-08-22
compiler: fix incorrect emitted Javascript for type switch

### LINKED ISSUE #682: Inconsistent type conversion for *js.Object
I've discovered that doing a type switch on a *js.Object behaves differently than an explicit type assert.  [Playground link](https://gopherjs.github.io/playground/#/a3Z1zkvL5I)

```Go
package main

import (
	"fmt"

)

func main() {
	obj := js.Global.Get("Object").New()
	obj.Set("foo","bar")
	
	fmt.Printf("obj.foo = %s\n", obj.Get("foo").String())
	
	var x interface{} = obj
	switch t := x.(type) {
	case *js.Object:
	    fmt.Printf("t.foo = %s\n", t.Get("foo").String())
	}
	
	if y, ok := x.(*js.Object); ok {
	    fmt.Printf("y.foo = %s\n", y.Get("foo").String())
	}
}
```

Produces the following output:

    obj.foo = bar 
    t.foo = undefined 
    y.foo = bar 

where I would have expected:

    obj.foo = bar 
    t.foo = bar
    y.foo = bar

myitcv:
This brings the code emitted for a type switch with a `*js.Object`
case in line with the runtime `$assertType` function. In the case of
a `*js.Object` value, we have to dereference via `.$val.object`.

Fixes #682

### DIFF
--- a/compiler/statements.go
+++ b/compiler/statements.go
@@ -123,7 +123,9 @@ func (c *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 			var bodyPrefix []ast.Stmt
 			if implicit := c.p.Implicits[clause]; implicit != nil {
 				value := refVar
-				if _, isInterface := implicit.Type().Underlying().(*types.Interface); !isInterface {
+				if typesutil.IsJsObject(implicit.Type().Underlying()) {
+					value += ".$val.object"
+				} else if _, ok := implicit.Type().Underlying().(*types.Interface); !ok {
 					value += ".$val"
 				}
 				bodyPrefix = []ast.Stmt{&ast.AssignStmt{
--- a/js/js_test.go
+++ b/js/js_test.go
@@ -571,3 +571,28 @@ func TestUint8Array(t *testing.T) {
 		t.Errorf("Non-empty byte array is not externalized as a Uint8Array")
 	}
 }
+
+func TestTypeSwitchJsObject(t *testing.T) {
+	obj := js.Global.Get("Object").New()
+	obj.Set("foo", "bar")
+
+	exp := "bar"
+
+	if act := obj.Get("foo").String(); act != exp {
+		t.Fatalf("Direct access to *js.Object field gave %q; expected %q", act, exp)
+	}
+
+	var x interface{} = obj
+	switch x := x.(type) {
+	case *js.Object:
+		if act := x.Get("foo").String(); act != exp {
+			t.Fatalf("Value passed through interface and type switch gave %q; expected %q", act, exp)
+		}
+	}
+
+	if y, ok := x.(*js.Object); ok {
+		if act := y.Get("foo").String(); act != exp {
+			t.Fatalf("Value passed through interface and type assert gave %q; expected %q", act, exp)
+		}
+	}
+}
--- a/js/js_test.go
+++ b/js/js_test.go
@@ -572,27 +572,27 @@ func TestUint8Array(t *testing.T) {
 	}
 }
 
-func TestTypeSwitchJsObject(t *testing.T) {
+func TestTypeSwitchJSObject(t *testing.T) {
 	obj := js.Global.Get("Object").New()
 	obj.Set("foo", "bar")
 
-	exp := "bar"
+	want := "bar"
 
-	if act := obj.Get("foo").String(); act != exp {
-		t.Fatalf("Direct access to *js.Object field gave %q; expected %q", act, exp)
+	if got := obj.Get("foo").String(); got != want {
+		t.Fatalf("Direct access to *js.Object field gave %q, want %q", got, want)
 	}
 
 	var x interface{} = obj
 	switch x := x.(type) {
 	case *js.Object:
-		if act := x.Get("foo").String(); act != exp {
-			t.Fatalf("Value passed through interface and type switch gave %q; expected %q", act, exp)
+		if got := x.Get("foo").String(); got != want {
+			t.Fatalf("Value passed through interface and type switch gave %q, want %q", got, want)
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


compiler/statements.go
```diff
@@ -126,6 +126,9 @@ func (c *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 				if _, isInterface := implicit.Type().Underlying().(*types.Interface); !isInterface {
 					value += ".$val"
 				}
+				if typesutil.IsJsObject(implicit.Type().Underlying()) {
+					value += ".object"
+				}
```
shurcooL: Would everything still work if the if statement is moved into the one above?

In other words, can ".object" be added without ".$val" in front of it?

Asking to learn.

shurcooL:
The test you added isn't passing in CI. Was it passing for you locally?

myitcv:
Hmm yes things were passing locally. Will take a look tomorrow and reply
properly then. Thanks for looking at this so promptly.

compiler/statements.go
myitcv: > Would everything still work if the if statement is moved into the one above?

If you mean:

```go
if _, isInterface := implicit.Type().Underlying().(*types.Interface); !isInterface {
  value += ".$val"

  if typesutil.IsJsObject(implicit.Type().Underlying()) {
    value += ".object"
  }
}
```

then yes, that's logically equivalent (because something cannot have an underlying type that is both an interface and `*js.Object`).

> In other words, can ".object" be added without ".$val" in front of it?

This is more to do with the structure of the value being type asserted (in Javascript world). Where a Javascript variable, say `x`, holds a value of an interface type, `x.$val` will give us the underlying value, regardless of type. Where the underlying value is a struct val or a pointer to a struct val, then it so happens that `x.$val === x`, hence `x.$val.object === x.object`.

So yes, we could elide the `.$val` in the case where the type in the switch case is `*js.Object` (because we know the structure of `js.Object`) but to my mind it's clearer and more consistent (with the rest of the code base) to keep things accessing via `.$val`.

> Asking to learn.

I've still got lots to learn about this code base: I dip in and out of it as required 😄 

... and it's quite possible some of my understanding is wrong let alone incomplete!

myitcv:
> The test you added isn't passing in CI. Was it passing for you locally?

Worked out what I did wrong; added the test to the wrong package. Tests that import the `js` package can only be run via `gopherjs`.

flimzy:
Thanks for tackling this one!

shurcooL:
> Worked out what I did wrong; added the test to the wrong package. Tests that import the `js` package can only be run via `gopherjs`.

Got it. What was the package you added it to originally? (A question I wouldn't have to ask if GitHub preserved history of a PR when its branch is force-pushed to.)

myitcv:
> What was the package you added it to originally?

`github.com/gopherjs/gopherjs/tests`

myitcv:
Rebased against `master` now the Go1.9 changes have been merged.

@shurcooL - anything else you want to look at in this? @neelance has given it 👍

compiler/statements.go
shurcooL: > If you mean:
> 
> ```go
> if _, isInterface := implicit.Type().Underlying().(*types.Interface); !isInterface {
>   value += ".$val"
> 
>   if typesutil.IsJsObject(implicit.Type().Underlying()) {
>     value += ".object"
>   }
> }
> ```

Yes, that's what I meant. I wanted to know if something would break/change if the code were written that way.

> then yes, that's logically equivalent (because something cannot have an underlying type that is both an interface and `*js.Object`).

Okay, I see.

To confirm my understanding is correct, let me answer my question:

> In other words, can ".object" be added without ".$val" in front of it?

From my understanding of what you said, no, ".object" cannot be added without ".$val" also being added.

Only these scenarios are possible:

- `implicit.Type().Underlying()` type is an interface - then `value` stays as is
- `implicit.Type().Underlying()` type is not an interface, but it's not `*js.Object` either - then `value` gets ".$val" appended
- `implicit.Type().Underlying()` type is not an interface, it is `*js.Object` - then `value` gets ".$val.object" appended

If my understanding is indeed correct, then wouldn't the code become simpler (to understand and reason about) if you do move the if statement inside?

shurcooL:
Sorry for the delay. I wanted to follow up to your comment, see https://github.com/gopherjs/gopherjs/pull/683#discussion_r136222495.

compiler/statements.go
shurcooL: Actually, I think the readability/clarity of can happen would be improved even more if you made it an if/else if statement with these 2 cases:

```Go
if typesutil.IsJsObject(implicit.Type().Underlying()) {
	value += ".$val.object"
} else if _, ok := implicit.Type().Underlying().(*types.Interface); !ok {
	value += ".$val"
}
```

(I also renamed `isInterface` to just `ok` because I think the long variable name doesn't actually improve readability here; but this is orthogonal and optional.)

What do you think? Is that better, worse, or the same in your opinion?

shurcooL:
@myitcv If you're in agreement with my suggestion at https://github.com/gopherjs/gopherjs/pull/683#discussion_r136223381, I can apply and merge this PR (unless you prefer to do it yourself). Let me know when you have a chance to look it over.

compiler/statements.go
myitcv: I easy either way. I don't think it's particularly unclear one way or the other. I've made the change, rebased this branch and pushed.

myitcv:
@shurcooL see https://github.com/gopherjs/gopherjs/pull/683#discussion_r137175588 - pushed up a new rebased commit. Over to you!

PR #687, opened 2017-08-26
build: unify setting of ulimit and stack_size

myitcv:
~Waiting for the Go 1.9 branch to land first - DONE in https://github.com/gopherjs/gopherjs/pull/651~

Picking up on @shurcooL's comment in https://github.com/gopherjs/gopherjs/pull/669#issuecomment-325138883.

We've started randomly seeing std library tests tip over the V8 stack size limit, causing text/template TestMaxExecDepth to fail randomly locally and on CI. This had previously been addressed by @neelance in 1f89545; the `--stack_size` flag was passed to NodeJS which in turn passed the value onto V8. But per https://github.com/nodejs/node/issues/14567#issuecomment-319367412 it was pointed out that the value of `ulimit -s` must be `>=` the value of `--stack_size` for the `--stack_size` to make any sort of sense. Hence this commit also harmonises the setting of `ulimit -s` during the CI test run with the value of `--stack_size` that is passed to NodeJS (which in turn passes that value onto V8) when running either gopherjs test or gopherjs run.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -20,7 +20,7 @@ test:
     - diff -u <(echo -n) <(go list ./compiler/natives/src/...) # All those packages should have // +build js.
     - gopherjs install -v net/http # Should build successfully (can't run tests, since only client is supported).
     - >
-      gopherjs test --minify -v --short
+      ulimit -s 10000 && ulimit -s && gopherjs test --short -v --minify
       github.com/gopherjs/gopherjs/tests
       github.com/gopherjs/gopherjs/tests/main
       github.com/gopherjs/gopherjs/js
--- a/circle.yml
+++ b/circle.yml
@@ -20,7 +20,7 @@ test:
     - diff -u <(echo -n) <(go list ./compiler/natives/src/...) # All those packages should have // +build js.
     - gopherjs install -v net/http # Should build successfully (can't run tests, since only client is supported).
     - >
-      ulimit -s 10000 && ulimit -s && gopherjs test --short -v --minify
+      ulimit -s 10000 && gopherjs test --minify -v --short
       github.com/gopherjs/gopherjs/tests
       github.com/gopherjs/gopherjs/tests/main
       github.com/gopherjs/gopherjs/js
--- a/tool.go
+++ b/tool.go
@@ -750,27 +750,27 @@ func runNode(script string, args []string, dir string, quiet bool) error {
 	}
 
 	if runtime.GOOS != "windows" {
-		// For non-Windows environments, we've seen issues with stack space
-		// limits causing Go standard library tests that are recursion-heavy to
-		// fail (see https://github.com/gopherjs/gopherjs/issues/661 for more details).
+		// We've seen issues with stack space limits causing
+		// recursion-heavy standard library tests to fail (e.g., see
+		// https://github.com/gopherjs/gopherjs/pull/669#issuecomment-319319483).
 		//
-		// There are two limits that come into play here:
+		// There are two separate limits in non-Windows environments:
 		//
-		// 1. V8 limit (NodeJS effectively wraps V8)
-		// 2. OS process limit
+		// -	OS process limit
+		// -	Node.js (V8) limit
 		//
-		// In order to limit the surface area of the gopherjs command and not
-		// expose V8 flags/options etc. to the caller, we control limit 1 via
-		// limit 2. Whatever value is returned by ulimit -s is the value that we
-		// pass on to NodeJS via the appropriate V8 flag.
+		// GopherJS fetches the current OS process limit, and sets the
+		// Node.js limit to the same value. So both limits are kept in sync
+		// and can be controlled by setting OS process limit. E.g.:
+		//
+		// 	ulimit -s 10000 && gopherjs test
+		//
 		var r syscall.Rlimit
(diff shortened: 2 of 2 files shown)


myitcv:
@shurcooL @neelance this is ready for review.

tool.go
```diff
@@ -750,9 +750,31 @@ func runNode(script string, args []string, dir string, quiet bool) error {
 	}
 
 	if runtime.GOOS != "windows" {
-		allArgs = append(allArgs, "--stack_size=10000", script)
+		// For non-windows OS environments, we've seen issues with stack space
+		// limits causeing Go std library tests that are recursion-heavy to fail
```
shurcooL: s/causeing/causing/

shurcooL: "std library" sounds weird. Why shorten one word, but not the other? Let's not shorten any, it's not helpful here.

s/std/standard/

tool.go
```diff
@@ -750,9 +750,31 @@ func runNode(script string, args []string, dir string, quiet bool) error {
 	}
 
 	if runtime.GOOS != "windows" {
-		allArgs = append(allArgs, "--stack_size=10000", script)
+		// For non-windows OS environments, we've seen issues with stack space
```
shurcooL: /s/windows/Windows/ since it's a brand name.

Also get rid of "OS", "For non-Windows environments" sounds better I think.

tool.go
```diff
@@ -750,9 +750,31 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		// limits causeing Go std library tests that are recursion-heavy to fail
+		// (see https://github.com/gopherjs/gopherjs/issues/661 for more detail).
```
shurcooL: /s/detail/details/ maybe? "more detail" sounds odd.

tool.go
```diff
@@ -750,9 +750,31 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		//
+		// There are two limits that come into play here, listed in order:
```
shurcooL: What does "listed in order" signify? If their order is important, how?

If not, just remove the phrase.

tool.go
```diff
@@ -750,9 +750,31 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		//
+		// 1. V8 limit (NodeJS effectively wraps V8)
+		// 2. OS process limit
+		//
+		// In order to limit the surface area of the gopherjs command and not
+		// expose V8 flags/options etc to the caller, we control limit 1 via
```
shurcooL: s/etc/, etc.,/

tool.go
```diff
@@ -750,9 +750,31 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		// limit 2. That is to say, whatever value is returned by ulimit -s is
```
shurcooL: We can probably get rid of "That is to say," it doesn't seem to add value to the sentence.

tool.go
```diff
@@ -750,9 +750,31 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		// essentially the value that we pass on to NodeJS via the appropriate V8
```
shurcooL: Remove "essentially". It's exactly equal to the value, isn't it?

tool.go
```diff
@@ -750,9 +750,31 @@ func runNode(script string, args []string, dir string, quiet bool) error {
...
+		// flag.
+		var r syscall.Rlimit
+		err := syscall.Getrlimit(syscall.RLIMIT_STACK, &r)
+		if err != nil {
+			return fmt.Errorf("failed to get stack size limit: %v", err)
+		}
+		// rlimit value is in bytes, we need rounded kBytes value per node --v8-options.
```
shurcooL: /s/kBytes/kilobytes/

Shortening just looks out of place and not saving much here.

circle.yml
```diff
@@ -20,7 +20,7 @@ test:
     - diff -u <(echo -n) <(go list ./compiler/natives/src/...) # All those packages should have // +build js.
     - gopherjs install -v net/http # Should build successfully (can't run tests, since only client is supported).
     - >
-      gopherjs test --minify -v --short
+      ulimit -s 10000 && ulimit -s && gopherjs test --short -v --minify
```
shurcooL: I understand `ulimit -s 10000` sets maximum stack size. Wha does `ulimit -s` (the second command) do?

shurcooL: Also, please don't change order of flags (unless you have a reason to be doing it). I placed `--minify` first on purpose in the previous commits to signify it's the most important flag.

My understanding this flag switch is because you rebased the entire line without changing it, is that right?

myitcv: `ulimit -s` is a means of echo-ing out the value of the OS-level stack size limit; I included it as a means of confirming the previous command was having the intended effect: not strictly necessary so I've removed it.

> Also, please don't change order of flags (unless you have a reason to be doing it). I placed --minify first on purpose in the previous commits to signify it's the most important flag.

Ok. 

> My understanding this flag switch is because you rebased the entire line without changing it, is that right?

Yes.

myitcv:
@shurcooL addressed those points. Let me know if there's anything else.

myitcv:
Thanks @shurcooL

PR #690, opened 2017-09-02
Remove unused variable in output

flimzy:
**TL;DR** The `$ptr` variable in every output function appears never to be used, so this eliminates it. I expect it's an obsolete hold-over from some previous change.

The story:

I noticed while playing with `uglifyjs`, a ton of warnings such as:

    WARN: Dropping unused variable $ptr [main.js:95,666198]
    WARN: Dropping unused variable $ptr [main.js:95,667947]

Which piqued my curiosity, to see if GopherJS could possibly be made intelligent enough to eliminate this `$ptr` value from the output when it's not used.

Looking at the output of one of my compiled projects, I saw many examples such as:

```javascript
        Object.ptr.prototype.Get = function(key) {
                var $ptr, key, o;
                o = this;
                return o.object[$externalize(key, $String)];
        };
```

Some grepping further revealed that the only time we _ever_ set the variable is in lines such as:

```javascript
                /* */ } return; } if ($f === undefined) { $f = { $blk: Cond.ptr.prototype.Wait }; } $f.$ptr = $ptr; $f._r = _r; $f.c = c; $f.$s = $s; $f.$r = $r; return $f;
```

which is just restoring the function state after returning from a possibly-blocking function call.

I found no other use of this variable, so I removed it, and re-compiled me project, both with and without the `-m` flag, and got the following size reductions:

| Build command     | Original bytes  | sans `$ptr` bytes  | Savings (%)     |
|-------------------|----------------:|--------------:|----------------:|
| gopherjs build    | 11,743,976      | 11,608,077    | 135,899 (1.16%)  |
| gopherjs build -m | 7,634,995       | 7,523,944     | 111,051 (1.45%)  |

My project appears to run perfectly with the modification; I look forward to the CI test results as well.


flimzy:
Hrm, one failure:

    === RUN   TestMaxExecDepth
    FAIL	text/template	2.272s

This passes fine on my local machine. Is this possibly related to #687?  (Apologies if the answer should be obvious--I've not been following that issue at all, except reading headlines)

shurcooL:
That failure does thanks for this PR, this is very interesting!

> Is this possibly related to #687? 

Yes, definitely looks related.

> Some grepping further revealed that the only time we _ever_ set the variable is in lines such as:
> 
> ```javascript
> /* */ } return; } if ($f === undefined) { $f = { $blk: Cond.ptr.prototype.Wait }; } $f.$ptr = $ptr; $f._r = _r; $f.c = c; $f.$s = $s; $f.$r = $r; return $f;
> ```

> which is just restoring the function state after returning from a possibly-blocking function call.

Can you elaborate on what will happen after the variable is no longer generated?

> I expect it's an obsolete hold-over from some previous change.

It would be nice to find the source commit where it's introduced, perhaps it'll help explain the original motivation.

flimzy:
> > which is just restoring the function state after returning from a possibly-blocking function call.

> Can you elaborate on what will happen after the variable is no longer generated?

The line in question is constructed [here](https://github.com/gopherjs/gopherjs/blob/master/compiler/package.go#L758-L762).  So by not adding `$ptr` to the `localVars` list, it is neither set nor unset.  Does this answer the question?

flimzy:
> It would be nice to find the source commit where it's introduced, perhaps it'll help explain the original motivation.

It was added with df432fe6, and since modified with df432fe6.

From reading these two commits, it looks as though it was never used. It looks like it was intended for future use, and that never came to fruition?  I'm sure @neelance will have to answer that.

flimzy:
> It would be nice to find the source commit where it's introduced, perhaps it'll help explain the original motivation.

It was added in df432fe6, where it was referenced in compiler/expressions.go, but then a few days later with 923aee9f21965, that reference was removed.

shurcooL:
Now that #687 is merged, I'm curious to rebase this PR and try CI again.

PR #696, opened 2017-09-06
Add `strings.Join` function that uses native code

microo8:
Added `strings.Join` function in `natives/src/strings` that uses the
native js function `array.join(separator)`

### DIFF
--- a/compiler/natives/src/strings/strings.go
+++ b/compiler/natives/src/strings/strings.go
@@ -20,6 +20,10 @@ func LastIndex(s, sep string) int {
 	return js.InternalObject(s).Call("lastIndexOf", js.InternalObject(sep)).Int()
 }
 
+func Join(a []string, sep string) string {
+	return js.InternalObject(a).Get("$array").Call("join", js.InternalObject(sep)).String()
+}
+
 func Count(s, sep string) int {
 	n := 0
 	// special cases


shurcooL:
How does it perform? Have you tried some benchmarks?

microo8:
I've tried this benchmark:

```go
package jointest

import (
	"strings"
	"testing"

)

func Join(a []string, sep string) string {
	return js.InternalObject(a).Get("$array").Call("join", js.InternalObject(sep)).String()
}

var testArray = []string{"a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "x", "y", "z"}

func TestEqual(t *testing.T) {
	want := strings.Join(testArray, ",")
	got := Join(testArray, ",")
	if len(want) != len(got) {
		t.Fatalf("not equal")
	}
	for i := range want {
		if want[i] != got[i] {
			t.Errorf("not equal")
		}
	}
}

func BenchmarkNative(b *testing.B) {
	for n := 0; n < b.N; n++ {
		Join(testArray, ",")
	}
}

func BenchmarkStrings(b *testing.B) {
	for n := 0; n < b.N; n++ {
		strings.Join(testArray, ",")
	}
}
```

The native implementation is much faster:

```
$ gopherjs test --bench=.               ~/Workspace/src/github.com/microo8/meh/gopherjs_tests/strings
goos: linux
goarch: js
BenchmarkNative          1000000              1622 ns/op
BenchmarkStrings           50000             30560 ns/op
PASS
ok      github.com/microo8/meh/gopherjs_tests/strings   6.562s
```

myitcv:
Just to flag one thing; not quite sure how the UTF16/UTF8 conversion affects things here. That's a bit of a drive by comment because I haven't given the point any more thought than simply noting it here.

flimzy:
> Just to flag one thing; not quite sure how the UTF16/UTF8 conversion affects things here. That's a bit of a drive by comment because I haven't given the point any more thought than simply noting it here.

This was a concern of mine, too, but I also hadn't invested the thought into how it might break things.

microo8:
So I've made some tests with these inputs:

```go
var faces = "☺☻☹"

var tests = [][]string{
	[]string{"a", "bcd"},
	[]string{"abcd"},
	[]string{"a", "b", "c", "d"},
	[]string{"1,", "2,", "3,", "4"},
	[]string{"1...", ".2...", ".3...", ".4"},
	[]string{"☺☻☹", ""},
	[]string{faces},
	[]string{"☺", "☻", "☹"},
}
```

It doesn't work for the faces. So i'm closing it.

flimzy:
> It doesn't work for the faces. So i'm closing it.

Sad. This is quite similar to the problems I ran into with trying to use `btoa` and `atob` for Base64 en/decoding.

shurcooL:
Yeah, that's unfortunate. The edge cases are often the things that make it hard to directly use a native function to implement something from Go standard library.

One possibility sometimes is to write a hybrid function that checks whether the input is one of the edge cases, then uses usual Go code for it; otherwise it uses the native function. Of course, doing the input check may be expensive and cause the optimization not to be worth it. But it's something to keep in mind.

In any case, thanks for trying and sending this PR and following up on it!

microo8:
I don't mind, I've learned something and for next time I'll think about the unicode edge cases. Also make more testing and benchmarks.

Issue #699, opened 2017-09-08
Build Issue


4ydx:
GopherJS 1.9-1
Golang 1.9

When building on one of my machines (Ubuntu 16.04.3 LTS), the resulting file is named "..js".  Any thoughts?  Running go build results in a binary that is properly named the parent folder.

I am running build outside the GOPATH.  This appears to be the reason for this so I am now using the "-o" flag in order to specify the resulting filename.  I don't know if this should be considered a bug or not.

shurcooL:
If I understand the issue correctly, the problem is that running `gopherjs build` outside GOPATH produces a suboptimally named "..js" file. Using `go build` produces a better name, using the parent folder name.

I suspect we can fix this, make `gopherjs build` behave as `go build` does.

It would be helpful if you could provide reproducible steps to trigger this.

4ydx:
go version go1.9 darwin/amd64
GopherJS 1.9-1

In a folder outside GOPATH build the following using gopherjs.

```
package main

import (
	"fmt"

	"github.com/gopherjs/gopherjs/js"
)

func main() {
	fmt.Println("Hello, playground")
	js.Global.Call("alert", "Hello, JavaScript")
	println("Hello, JS console")
}
```

4ydx:
I have a fix written.  I am a little unclear about the process of getting it approved.

PR #701, opened 2017-09-14
Issue (#699)

4ydx:
Running gopherjs build outside of GOPATH created files named "..js".

### DIFF
--- a/compiler/statements.go
+++ b/compiler/statements.go
@@ -123,7 +123,9 @@ func (c *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 			var bodyPrefix []ast.Stmt
 			if implicit := c.p.Implicits[clause]; implicit != nil {
 				value := refVar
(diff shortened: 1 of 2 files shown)


tool.go
```diff
@@ -143,7 +143,15 @@ func main() {
 					}
 					if len(pkgs) == 1 { // Only consider writing output if single package specified.
 						if pkgObj == "" {
-							pkgObj = filepath.Base(pkg.ImportPath) + ".js"
+							if pkg.ImportPath == "." {
+								dirPath, err := os.Getwd()
```
shurcooL: No need to do `os.Getwd` here, you can use [`currentDirectory`](https://github.com/gopherjs/gopherjs/blob/b40cd48c38f9a18eb3db20d163bad78de12cf0b7/tool.go#L42) variable:

https://github.com/gopherjs/gopherjs/blob/b40cd48c38f9a18eb3db20d163bad78de12cf0b7/tool.go#L42-L50

PR #702, opened 2017-09-14
Build issue outside of GOPATH resulted in file named '..js'. (#699)

4ydx:


4ydx:
Apologies for creating a new pull request.

tool.go
```diff
@@ -143,7 +143,11 @@ func main() {
 					}
 					if len(pkgs) == 1 { // Only consider writing output if single package specified.
 						if pkgObj == "" {
-							pkgObj = filepath.Base(pkg.ImportPath) + ".js"
+							if pkg.ImportPath == "." {
```
flimzy: I wonder if this would be a problem for the (much less common) case of `..` as well?

Why someone would `go build ..` I have no idea... :)  So whether to consider this is entirely up to you.

shurcooL: Good catch. It's possible for this to happen, I verified. It'd be nice to fix it too.

tool.go
```diff
@@ -143,7 +143,8 @@ func main() {
 					}
 					if len(pkgs) == 1 { // Only consider writing output if single package specified.
 						if pkgObj == "" {
-							if pkg.ImportPath == "." {
+							importPath := filepath.Base(pkg.ImportPath)
+							if importPath == "." || importPath == ".." {
 								pkgObj = filepath.Base(currentDirectory) + ".js"
```
flimzy: I don't think this is right. I think for the `..` case, we want to use the name of the parent directory.

4ydx: Thanks for pointing that out.

tool.go
```diff
@@ -143,7 +143,20 @@ func main() {
...
 					if len(pkgs) == 1 { // Only consider writing output if single package specified.
 						if pkgObj == "" {
-							pkgObj = filepath.Base(pkg.ImportPath) + ".js"
+							importPath := filepath.Base(pkg.ImportPath)
+							if importPath == "." {
+								pkgObj = filepath.Base(currentDirectory) + ".js"
+							} else if importPath == ".." {
+								// allow for build requests like "build .. or build ../.."
+								path := currentDirectory
+								count := strings.Count(pkg.ImportPath, "..")
```
flimzy: I hate to keep commenting on this... It's such a corner case. But I don't think this is right. It would incorrectly count for a path like `foo..bar/..`.

4ydx: You are tenacious.

4ydx: I was a bit confused by what you might mean, thinking that perhaps there are other ways of building that I am unfamiliar with (not to say there are not), but if you mean to say "gopherjs build over..here/.." then I just tried that and it works for me with the patch as-is.

flimzy: I think that means your new codepath is never being executed.  Because "over..here/.." clearly contains two instances of "..".

When I execute similar tests, I also get the right output, but I also see that `pkg.ImportPath` has already had the trailing "/.." removed.  In fact, I can't even get `gopherjs build ..` to break.  I expect this is related to the operating system and/or shell I'm using (Linux/bash).  Since @shurcooL  was able to produce the problem with "..", maybe he can provide further details (although I'm pretty sure he's using OSX) and/or test some other ".." combinations.

shurcooL:
The implementation has grown quite complex, trying to deal with the edge cases one by one.

I suspect this can be resolved more simply. Would the following strategy work?

1. Check if import path is local using [`build.IsLocalImport`](https://godoc.org/go/build#IsLocalImport).

2. If not, use `path.Base` as before.

3. If local, use `filepath.Base` of `pkg.Dir`.

I.e., something like this:

```Go
if !build.IsLocalImport(pkg.ImportPath) {
	pkgObj = path.Base(pkg.ImportPath) + ".js"
} else {
	pkgObj = filepath.Base(pkg.Dir) + ".js"
}
```

Any problems with that algorithm?

tool.go
```diff
@@ -143,7 +143,11 @@ func main() {
 					}
 					if len(pkgs) == 1 { // Only consider writing output if single package specified.
 						if pkgObj == "" {
-							pkgObj = filepath.Base(pkg.ImportPath) + ".js"
+							if !build.IsLocalImport(pkg.ImportPath) {
+								pkgObj = filepath.Base(pkg.ImportPath) + ".js"
```
shurcooL: Import paths are always separated by a forward slash, never backslash, so we should use `path.Base` rather than `filepath.Base` on it. Mind making that change?

shurcooL:
Okay, this is a simple solution that should work.

If we discover concrete cases where it doesn't work (e.g., due to use of symlinks), then we can reintroduce the code that uses base of import path when it's not local, along with a comment explaining why it's needed. So please report any issues as soon as you spot them.

Until then, let's enjoy the simpler solution.

LGTM. Thanks!

PR #706, opened 2017-10-07
Remove an unused local variable canBlock

hajimehoshi:
This PR removes `canBlock` property from `$curGoroutine`, that is not used anywhere.

### DIFF
--- a/compiler/prelude/jsmapping.go
+++ b/compiler/prelude/jsmapping.go
@@ -158,13 +158,7 @@ var $externalizeFunction = function(v, t, passThis) {
         }
         args.push($internalize(arguments[i], t.params[i]));
       }
-      var canBlock = $curGoroutine.canBlock;
-      $curGoroutine.canBlock = false;
-      try {
-        var result = v.apply(passThis ? this : undefined, args);
-      } finally {
-        $curGoroutine.canBlock = canBlock;
-      }
+      var result = v.apply(passThis ? this : undefined, args);
       switch (t.results.length) {
       case 0:
         return;


hajimehoshi:
ping @neelance ?

shurcooL:
For those how know JavaScript better than I do, is there any difference in behavior between these two snippets of code?

```JS
try {
    var result = v.apply(passThis ? this : undefined, args);
} finally {}
```

vs

```JS
var result = v.apply(passThis ? this : undefined, args);
```

That is the only part of this change that I'm not fully confident about.

slimsag:
@shurcooL From my knowledge:

1. If the snippets are exactly as you wrote, then *no*, there is not a difference between their behavior.
2. If the `finally {}` block actually has some code inside of it, then yes. Statements inside of `finally` execute regardless of whether or not the `try` code succeeded (like how a function in Go can return an error but `defer` code still runs).

You can read more here if you are interested: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

shurcooL:
Thanks @slimsag, that helps me understand this change better.

The finally block being removed here has code `$curGoroutine.canBlock = canBlock;`, but removing that should be safe since `canBlock` is not used anywhere else (it was removed in 2d0acd02933cf349b065398c73490a14f7d9bada), so changing its value has no effect.

I just wanted to make sure there's no other effect to having code be in a `try`/`finally` block (e.g., in case of exceptions or something).

shurcooL:
The `var result = v.apply(...);` line was not inside a `try` block originally. It got placed in one only when the local `canBlock` variable was introduced in 6e21f510064ca112881fe5ff3542a082a4f9da0f.

So, given `canBlock` is no longer used anywhere else, it should be safe to remove the `try` block and go back to a simple `v.apply` call.

There have been no objections from @neelance after some time, and I'm quite confident of this being correct now, so merging. Thanks again @hajimehoshi and @slimsag!

Issue #624, opened 2017-04-16
'gopherjs test' always fails tests


slimsag:
With a simple test like:

```go
package example

import "testing"

func TestExample(t *testing.T) {

}
```

It will always fail when ran via `gopherjs test`:
```
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
PASS
fatal error: all goroutines are asleep - deadlock!
FAIL	github.com/gopherjs/vecty/example	0.344s
```

I'm wondering if this is a known issue / people are not using `gopherjs test`, or if it's potentially something configured incorrectly on my system.

- Using `gopherjs` version 3496c6f94e1b945f3fc9580f6982675c0a74cd6c
- `node --version` reports `v7.9.0` (the latest available via `brew`)

shurcooL:
Is that the only test you have, or are there more?

It should definitely work.

I'm guessing you don't have system call support installed, do you? https://github.com/gopherjs/gopherjs/blob/master/doc/syscalls.md It's possible something we changed more recently has made it mandatory.

shurcooL:
For reference, this is what my output is:

```
$ gopherjs test
PASS
ok  	_/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/shurcooL/play/31	0.458s
$ gopherjs test -v
=== RUN   TestExample
--- PASS: TestExample (0.00s)
PASS
ok  	_/Users/Dmitri/Dropbox/Work/2013/GoLand/src/github.com/shurcooL/play/31	0.469s
$ 
```

(I'm not sure why it's printing the full path instead of import paths though...)

slimsag:
> Is that the only test you have, or are there more?

Correct, just that test. No more.

> I'm guessing you don't have system call module installed, do you?

Ok, I gave that a shot and it fixed it. My output is now:

```bash
$ gopherjs test
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
PASS
ok  	github.com/gopherjs/vecty/example	0.326s
$ gopherjs test -v
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
=== RUN   TestExample
--- PASS: TestExample (0.00s)
PASS
ok  	github.com/gopherjs/vecty/example	0.334s
```

I don't know why I still get the Node.js 4.x warning when my `node --version` reports `v7.9.0` though.

[...]

Issue #531, opened 2016-10-01
calling os.Exit(0) from a `gopherjs run` invocation fails with "fatal error: all goroutines are asleep - deadlock!"


nullstyle:
Here's the repro script:

``` golang
package main

import (
    "fmt"
    "os"
)

func main() {
    fmt.Println("hello world")
    os.Exit(0)
}
```

sample results:

``` shell
➜  testj git:(mithril) ✗ gopherjs run ./main.go
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
hello world
fatal error: all goroutines are asleep - deadlock!
```

This appears to be the root cause of #395 as well.

shurcooL:
Just checking, are you familiar with https://github.com/gopherjs/gopherjs#application-lifecycle?

nullstyle:
My understanding was that only applies to the browser... `gopherjs test` runs using node, AFAIK.

nullstyle:
You can see the workaround I'm using presently at https://github.com/nullstyle/gopherjs/commit/8d4d6d07a8d9acca07efebeba727363324c33004

shurcooL:
I cannot reproduce with latest version of GopherJS.

I have Node version v6.7.0 with both node-syscall and source-map-support modules installed.

Running the program above gives me:

```bash
$ node --version
v6.7.0

$ go run main.go
hello world
$ gopherjs run main.go 
hello world
```

Can you try again with both node-syscall and source-map-support modules installed? I suspect that they might make a difference.

If so, this is definitely related to https://github.com/gopherjs/gopherjs/pull/449 (see discussion about source-map-support there).

nullstyle:
after installing node-syscall, the issue no longer occurs.   I didn't even know node-syscall was a thing, seems like it should be in the README.  Are you open to me making a PR to add that?

Is there anyway to get node-syscall into npm?

shurcooL:
Okay, thanks for confirming.

> I didn't even know node-syscall was a thing, seems like it should be in the README.

It is in the README, in the Installation and Usage section.

https://github.com/gopherjs/gopherjs#installation-and-usage

nullstyle:
> It is in the README, in the Installation and Usage section.

IMO It's not clear that it's required to get basic behavior like `gopher test` working.

hajimehoshi:
I'm not familiar with node, and I don't know how to install node-syscall...

hajimehoshi:
[...]

Issue #417, opened 2016-03-06
`gopherjs test github.com/eaburns/flac` fails


hajimehoshi:
```
$ go test github.com/eaburns/flac
ok      github.com/eaburns/flac 0.006s

$ gopherjs test github.com/eaburns/flac
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
PASS
fatal error: all goroutines are asleep - deadlock!
FAIL    github.com/eaburns/flac 0.338s
```

IIUC that flac package and its dependencies are in pure Go so the test should be passed in theory.

shurcooL:
Cannot reproduce with latest version of GopherJS and that package.

```
$ go get -u github.com/eaburns/flac
$ go test github.com/eaburns/flac
ok  	github.com/eaburns/flac	0.023s
$ gopherjs test github.com/eaburns/flac
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
PASS
ok  	github.com/eaburns/flac	0.548s
$ 
```

Can you try updating GopherJS to latest version (`go get -u github.com/gopherjs/gopherjs`), updating that package to latest version (`go get -u github.com/eaburns/flac`) and redoing the test?

There have been some bug fixes with goroutine deadlock detection in GopherJS in the recent past.

hajimehoshi:
Yeah, I've updated GopherJS but the still deadlock happens.

hajimehoshi:
Ah, it's succeeded after installing node-gyp. This is still mysterious but it'd not be a practical problem. Thanks.

Issue #395, opened 2016-02-10
gopherjs test fails


davelondon:
GopherJS: 0faf2ef97c0ada75462340a914f01fec80521216
Node: node/5.6.0
Go: go version go1.5.1 darwin/amd64

Given the following project:

gotest.go:
````go
package jstest
````
gotest_test.go
```go
package jstest

import "testing"

func TestNil(t *testing.T) {

}
```

... running gopherjs test produces an error:

```
$ gopherjs test
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
PASS
fatal error: all goroutines are asleep - deadlock!
FAIL	github.com/davelondon/jstest	0.359s
```

Of course go test passes:

```
$ go test
PASS
ok  	github.com/davelondon/jstest	0.008s
```

shurcooL:
Hi, thanks for reporting this.

However, I'm unable to reproduce. `gopherjs test` works as expected for that input:

```
$ go test
PASS
ok  	github.com/shurcooL/play/31	0.018s
$ gopherjs test
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
PASS
ok  	_/Users/Dmitri/Go/src/github.com/shurcooL/play/310.504s
$ gopherjs test -v
gopherjs: Source maps disabled. Use Node.js 4.x with source-map-support module for nice stack traces.
=== RUN   TestNil
--- PASS: TestNil (0.00s)
PASS
ok  	_/Users/Dmitri/Go/src/github.com/shurcooL/play/310.503s
$ 
```

Using:

```
go version go1.5.3 darwin/amd64
GopherJS version 0faf2ef97c0ada75462340a914f01fec80521216
```

davelondon:
Yes I can't reproduce it in a docker container. I'll see if there's something wrong with the setup of my Mac.

davelondon:
OK, this is what I've done:

Rebooted Mac
Deleted $GOPATH/src, $GOPATH/bin, $GOPATH/pkg
Upgraded Go to 1.5.3
Downgraded Node to 4.3.0
go get github.com/gopherjs/gopherjs
go get github.com/davelondon/jstest

Still has the same error. Can you think of anything else to try?

davelondon:
I'll close this issue and continue trying to resolve this myself. I'll post again here when I find the solution.

nullstyle:
[...]

PR #710, opened 2017-10-12
syscall: fallback to process.exit when available

### LINKED ISSUE #626: Document that gopherjs test requires syscall module (or make it not require it, if possible).
See https://github.com/gopherjs/gopherjs/issues/624#issuecomment-294386098 for details.

Lekensteyn:
"go test" currently always fails with "fatal error: all goroutines are
asleep - deadlock!". In fact, all programs that use "os.Exit" will fail
with this.

It turns out that the current "runtime.Goexit" function throws an
exception which propagates up which violates the expected behavior of
the exit syscall (which should not return). Therefore call
"process.exit" when available.

Fixes #626
___
Test with `go test` on any package or just:
```go
package main

import "os"

func main() {
	os.Exit(0)
}
```

Also note that any functions registered with `process.on("exit", ...)` will still be executed. Not ideal, but this is better than nothing when syscalls are not available.

### DIFF
--- a/compiler/natives/src/syscall/syscall_unix.go
+++ b/compiler/natives/src/syscall/syscall_unix.go
@@ -67,6 +67,9 @@ func Syscall(trap, a1, a2, a3 uintptr) (r1, r2 uintptr, err Errno) {
 		return uintptr(array.Length()), 0, 0
 	}
 	if trap == SYS_EXIT {
+		if process := js.Global.Get("process"); process != js.Undefined {
+			process.Call("exit", a1)
+		}
 		runtime.Goexit()
 	}
 	printWarning()
--- a/compiler/natives/src/syscall/syscall_unix.go
+++ b/compiler/natives/src/syscall/syscall_unix.go
@@ -61,9 +61,24 @@ func Syscall(trap, a1, a2, a3 uintptr) (r1, r2 uintptr, err Errno) {
 	}
 	if trap == SYS_WRITE && (a1 == 1 || a1 == 2) {
 		array := js.InternalObject(a2)
-		slice := make([]byte, array.Length())
-		js.InternalObject(slice).Set("$array", array)
-		printToConsole(slice)
+		if process := js.Global.Get("process"); process != js.Undefined {
+			var buffer interface{}
+			if bufferFrom := js.Global.Get("Buffer").Get("from"); bufferFrom != js.Undefined {
+				// since node 5.10.0
+				buffer = bufferFrom.Invoke(array)
+			} else {
+				buffer = js.Global.Get("Buffer").New(array)
+			}
+			if a1 == 1 {
+				process.Get("stdout").Call("write", buffer)
+			} else {
+				process.Get("stderr").Call("write", buffer)
+			}
+		} else {
+			slice := make([]byte, array.Length())
+			js.InternalObject(slice).Set("$array", array)
+			printToConsole(slice)
+		}
 		return uintptr(array.Length()), 0, 0
 	}
 	if trap == SYS_EXIT {


theclapp:
Do we care about the issue mentioned [here](https://stackoverflow.com/questions/5266152/how-to-exit-in-node-js/37592669#37592669), which concludes

> If you can you should set the exit code and allow Node to exit gracefully:
>
>     process.exitCode = 1;

Lekensteyn:
Nope we do not, we want to exit immediately and not just set the status code. See https://nodejs.org/api/process.html#process_process_exit_code

Btw, this issue was also known before as #395, #417, #531, #624 and also as (still open) #342.

flimzy:
> Nope we do not,

Can you elaborate?  Why do we not care?  This seems to have the potential to break expected behavior, if stdout isn't flushed before exit.

shurcooL:
> This seems to have the potential to break expected behavior, if stdout isn't flushed before exit.

@flimzy, `os.Exit` exits the process immediately as well, it doesn't wait for buffers to flush or anything.

shurcooL:
Thanks for this @Lekensteyn. At a high level, this makes sense.

compiler/natives/src/syscall/syscall_unix.go
```diff
@@ -67,6 +67,10 @@ func Syscall(trap, a1, a2, a3 uintptr) (r1, r2 uintptr, err Errno) {
 		return uintptr(array.Length()), 0, 0
 	}
 	if trap == SYS_EXIT {
+		process := js.Global.Get("process")
+		if process != js.Undefined {
```
shurcooL: Style-wise, this is a good time to inline `process := js.Global.Get("process")` into the if statement:

```Go
if process := js.Global.Get("process"); process != js.Undefined {
	process.Call("exit", a1)
}
```

`process` is only needed inside the if statement, so it's good to give it a smaller scope.

(/cc @gmlewis FYI.)

flimzy:
Does output to stdout not normally block in Go?

shurcooL:
I think it does.

Are you referring to how writing to `os.Stdout` in GopherJS includes some internal buffering for browsers, so that a single line of output maps to a single line in browser  console, and we might need to take that into account before exiting?

flimzy:
My concern, without any attempt to test this, is that something like this might lead to lost output, particularly in tests running in Node.js (which I think is only place this patch matters, right?

    fmt.Println("something")
    t.FailNow()

If my understanding of the SO link above, plus Go/GopherJS is correct, the Println would be translated to an async nidejs syscall, but GopherJS would treat it as if it were a completed blocking operation. So it the Node.js call may not actually complete before the program exists.

It's certainly quite possible that I'm wrong about one or more of the pieces to this puzzle, though. In that case, please accept my apologies for wasting time. :)

Lekensteyn:
@flimzy Your test (or a main function which calls `fmt.Println` before `os.Exit`) does seem to print stuff correctly here. However `fmt.Print("x")` does not. Tested with node 8.6.0. By the time `syscall(SYS_EXIT)` gets executed, it is probably already too late to flush buffers (if any).

`fmt.Print("x")` did not work because `printToConsole` is doing (unconditional) line buffering. Fine for a browser, not so nice though for node. Two options: use `process.stdout.write` directly or invoke the internal `$flushConsole` call before exit.

I have just pushed one that implements the former, WDYT? The generated code is a bit ugly though, 
```js
buffer = $ifaceNil;
bufferFrom = $global.Buffer.from;
if (!(bufferFrom === undefined)) {
	buffer = new $jsObjectPtr(bufferFrom(array));
} else {
	buffer = new $jsObjectPtr(new ($global.Buffer)(array));
}
if (a1 === 1) {
	process.stdout.write($externalize(buffer, $emptyInterface));
} else {
	process.stderr.write($externalize(buffer, $emptyInterface));
}
```
it could also have been written as:
```js
if (Buffer.from) {
    buffer = Buffer.from(array);
} else {
    buffer = new Buffer(array);
}
if (a1 === 1) {
    process.stdout.write(buffer);
} else {
    process.stderr.write(buffer);
}
```

Related, but probably not affecting this: https://github.com/nodejs/node/issues/6456

flimzy:
@Lekensteyn Thanks for testing that.

Your approach of using `process.stdout.write` makes sense to me.

shurcooL:
We should take this documented behavior into account:

> In the browser, calling `os.Exit` (e.g. indirectly by `log.Fatal`) also does not terminate the execution of the program. For convenience, it calls `runtime.Goexit` to immediately terminate the calling goroutine.

_(Source: https://github.com/gopherjs/gopherjs#application-lifecycle.)_

Do we want to change that? It might break existing programs. There might've been a good reason for making that design decision, we should remember what it was.

Lekensteyn:
The rationale for the documented behavior for browsers is noted in https://github.com/gopherjs/gopherjs/issues/229#issuecomment-105253457. It does not apply to node. At least, not when you run the script directly, I don't think that importing a gopherjs program as package is supported, is it?

Also note that with the syscalls package installed, `os.Exit` also does not return.

Lekensteyn:
Ping

shurcooL:
Thanks for the ping, and sorry about the delay. I will try to find some time to work on this.

I still need to get a better understanding of the change and gain confidence that it doesn't have any unexpected unintended side-effects.

Lekensteyn:
Poke :-)

Lekensteyn:
Any updates on this one?

flimzy:
Is any of this still relevant after the syscall rewwork 

cc/ @nevkontakte

nevkontakte:
I think this is obsolete as of https://github.com/gopherjs/gopherjs/pull/1111.

PR #711, opened 2017-10-13
Remove unused argument 'direct'

hajimehoshi:
@shurcooL PTAL

### DIFF
--- a/compiler/prelude/goroutines.go
+++ b/compiler/prelude/goroutines.go
@@ -114,7 +114,7 @@ var $throw = function(err) { throw err; };
 var $noGoroutine = { asleep: false, exit: false, deferStack: [], panicStack: [] };
 var $curGoroutine = $noGoroutine, $totalGoroutines = 0, $awakeGoroutines = 0, $checkForDeadlock = true;
 var $mainFinished = false;
-var $go = function(fun, args, direct) {
+var $go = function(fun, args) {
   $totalGoroutines++;
   $awakeGoroutines++;
   var $goroutine = function() {


theclapp:
FWIW, LGTM.  I see no references to `direct` in the function itself or use of it when called, either, in [compiler/statements.go#L444](https://github.com/gopherjs/gopherjs/blob/master/compiler/statements.go#L444).

shurcooL:
Where are the places that call function `$go`? Do they pass 2 or 3 parameters? If 3, we should remove the 3rd parameter from the call sites as well.

shurcooL:
Looking at some generated .js code, I see that all places that call `$go` provide 2 parameters.

I'll try to find in history when `direct` was added and removed so we have more context on this.

shurcooL:
There are only 2 places that generate $go calls:

https://github.com/gopherjs/gopherjs/blob/415225646bb92c4449bb484646f2c95a98402f6f/compiler/statements.go#L444

https://github.com/gopherjs/gopherjs/blob/415225646bb92c4449bb484646f2c95a98402f6f/compiler/compiler.go#L171

Both provide only 2 arguments.

It has been that way ever since go statement support was added in https://github.com/gopherjs/gopherjs/commit/cf716756eec6bc4a64d99c1227aa5178c622df37.

Going to look where `direct` parameter was added, because it wasn't there originally.

shurcooL:
`direct` was added in 3afebab303ea8bdfb55ca4d71d28375dd0ddefd3 and _was used_ then. Let's see when it stopped being used in `$go`...

shurcooL:
2d0acd02933cf349b065398c73490a14f7d9bada is the key commit. It removed uses of `direct` from `$go` function and callers, but forgot to remove it as a parameter.

theclapp:
Nice detective work, @shurcooL.

shurcooL:
Merged. Thanks again!

I didn't merge sooner because I wanted to give a few days for anyone to leave any comments, but then forgot about it. If there's been no activity for more than 7 days, a friendly ping is welcome and appreciated.

PR #713, opened 2017-10-17
Minor readabilty improvements

leitzler:


tool.go
```diff
@@ -115,10 +115,8 @@ func main() {
 							s.Watcher.Add(name)
 						}
 					}
-					if err := s.BuildFiles(args, pkgObj, currentDirectory); err != nil {
-						return err
-					}
-					return nil
+					err := s.BuildFiles(args, pkgObj, currentDirectory)
+					return err
```
flimzy: If you're going to make the change, why not go all the way to:

    return s.BuildFiles(args, pkgObj, currentDirectory)

leitzler: Imo it's more readable to see that it returns an error, but it isn't that strong opinion :)

shurcooL: That would work too, but personally, I don't mind (in fact, I slightly prefer) it as written. It has to do with where I perceive the emphasis to be when reading the code.

When I see:

```Go
err := s.BuildFiles(args, pkgObj, currentDirectory)
return err
```

I read that as "do `s.BuildFiles`, then return the error result."

When I see:

```Go
return s.BuildFiles(args, pkgObj, currentDirectory)
```

I read that as "return the result of building files", which seems to put more emphasis on the return value, but in fact the return value is pretty insignificant (it doesn't return a package it built, just a nil error most of the time).

Anyway, this is a personal style and both ways are fine.

shurcooL:
Thanks for the PR. I agree these changes are an improvement.

However, I remember seeing an exact same (or very similar) PR in the past (can't easily find it now), and @neelance rejected it then, because he preferred the current style at the time.

However, it was a few years ago, and I wouldn't be surprised if he's okay with this code simplification now. I think we should merge it, because the code is simpler, and it allows other people who read it not get distracted and want to send a PR just like this one.

shurcooL:
There've been no objections, and I think this is valid, idiomatic Go code that is slightly simpler. It'll avoid future people sending the same simplification. Merging. Thanks again!

PR #728, opened 2017-12-13
net/http: Implement cancellation via context for XHRTransport

FlorianUekermann:
Cancellation support for fetch would be nice too, but doesn't seem to be widely supported. Maybe we should default to XHR again.


shurcooL:
(It would probably be better to have this discussion outside of this PR, since it's not related to XHR...)

> Cancellation support for fetch would be nice too, but doesn't seem to be widely supported.

The spec for abortable fetch has been [finalized](https://github.com/whatwg/fetch/issues/447#event-1257380033) in September. It has already landed in some browsers and will be lading in more soon, see https://developers.google.com/web/updates/2017/09/abortable-fetch.

> fetch doesn't seem to have any benefits over XHR until streaming support is implemented.

Streaming support is implemented in GopherJS's fetch transport though.

FlorianUekermann:
> Streaming support is implemented in GopherJS's fetch transport though.
True. I remembered that wrong. It is implemented for incoming data, but not for the outgoing body.
So I guess, we should only use XHR if fetch doesn't support abort. I agree, that is definitely beyond this pull request.

Let me know if you have any concerns.

compiler/natives/src/net/http/http.go
```diff
@@ -99,6 +99,9 @@ func (t *XHRTransport) RoundTrip(req *Request) (*Response, error) {
 	}
 
 	select {
+	case <-req.Context().Done():
+		xhr.Call("abort")
+		return nil, <-errCh
```
shurcooL: Not that I want you to change anything, just looking to learn. How did you decide between doing `return nil, <-errCh` and something like `return nil, ctx.Err()`? Did you consider the tradeoffs?

FlorianUekermann: Since abort may cause onabort to execute, the errCh channel needs to be drained.

Actually... The code also relies on abort not triggering onabort if the request already failed or succeeded.
But that assumption was already present, since it was already possible to call abort. I haven't checked. I think https://xhr.spec.whatwg.org/#the-abort()-method implies that this assumption is correct, but I have misunderstood WHATWG standards before, since it is an unfamiliar style of writing specs for me.

FlorianUekermann:
Can we merge this? It is 3 lines that make sense and have been working fine for me for a while now. Just ping me so I can resend so it fits in with the new fx_vfsdata.go from the commits in between.

FlorianUekermann:
I dropped XHR from my code today and am using fetch instead. We should revisit this issue once abortable fetch is implemented works in browsers. Closing.

Issue #736, opened 2018-01-03
sync: inconsistent mutex state


canning-duck:
Hello,

We have some piece of code with a few goroutines and mutexes. Given a "certain" number of concurrent operations, the runtime panics with "sync: inconsistent mutex state" in GopherJS 1.9 (this one: https://golang.org/src/sync/mutex.go#L143). The issue does not occur with GopherJS 1.8, or on any other platform with Go 1.9 (android/ios/macos).

Regarding the Go and GopherJS source code, and as far as I understand it (ie. not very much), it seems that:
* Go 1.9 implements a new "mutex fairness" behavior (https://golang.org/src/sync/mutex.go#L42) that relies on the new parameters `lifo` and `handoff` of `runtime_SemacquireMutex` and `runtime_Semrelease` respectively (https://golang.org/src/sync/runtime.go).
* GopherJS 1.9 does not take theses parameters into account (https://github.com/gopherjs/gopherjs/blob/master/compiler/natives/src/sync/sync.go).

I'm not yet at ease isolating the issue in a unit-test but I'm ready to try patching our local copy of GopherJS and see how it goes. If I get it right, `lifo` is about adding `ch` at the head of `semWaiters` instead of the tail. But what about `handoff`? Has anyone any clue?

canning-duck:
I've found a scenario:

```
package main

import (
	"fmt"
	"sync"
	"time"
)

func main() {
	var mx sync.Mutex
	var wg sync.WaitGroup
	work := func(i int) {
		defer func() {
			err := recover()
			if err != nil {
				fmt.Println(i, "recover", err)
			}
		}()

		mx.Lock()
		time.Sleep(2 * time.Millisecond)
		mx.Unlock()

		mx.Lock()
		time.Sleep(2 * time.Millisecond)
		mx.Unlock()

		fmt.Println(i, "work done")
	}
	for i := 0; i < 10; i += 1 {
		wg.Add(1)
		go func(i int) {
			defer wg.Done()
			work(i)
		}(1 + i)
	}
	wg.Wait()
	fmt.Println("all work done")
}
```

https://gopherjs.github.io/playground/#/uWcv2N7D2L

[...]

PR #737, opened 2018-01-05
Fix "sync: inconsistent mutex state" panic

canning-duck:
* Fix the panic by preventing undue acquisition of the semaphore.
* Implement the `lifo` parameter in `runtime_SemacquireMutex` (not strictly required but improve the mutex fairness).

TestMutexFairness is now passing. Tested with Go 1.9 (`gopherjs test -v sync`).

https://github.com/gopherjs/gopherjs/issues/736


compiler/natives/src/sync/sync.go
```diff
@@ -5,22 +5,32 @@ package sync
 import "github.com/gopherjs/gopherjs/js"
 
 var semWaiters = make(map[*uint32][]chan bool)
+var semAwoken = make(map[*uint32]uint32)
```
shurcooL: It would be nice to document what this map tracks.

shurcooL:
Thank you again @canning-duck for investigating and fixing this!

PR #744, opened 2018-01-27
Replaces sync import with nosync before build

dave:
The `parseAndAugment` function replaces the `sync` import with `github.com/gopherjs/gopherjs/nosync`, however [this](https://github.com/gopherjs/gopherjs/blob/00a3767085dffc9baa5e1c320b28e3629a56d323/build/build.go#L568) builds all imports before the replacement. 

This results in the `sync` package being built when compiling `math/rand` amongst others. Copying the sync/nosync replacement logic from `parseAndAugment` fixes the problem.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -565,6 +565,14 @@ func (s *Session) BuildPackage(pkg *PackageData) (*compiler.Archive, error) {
 			if importedPkgPath == "unsafe" || ignored {
 				continue
 			}
+
+			switch pkg.ImportPath {
+			case "crypto/rand", "encoding/gob", "encoding/json", "expvar", "go/token", "log", "math/big", "math/rand", "regexp", "testing", "time":
+				if importedPkgPath == "sync" {
+					importedPkgPath = "github.com/gopherjs/gopherjs/nosync"
+				}
+			}
+
 			importedPkg, _, err := s.buildImportPathWithSrcDir(importedPkgPath, pkg.Dir)
 			if err != nil {
 				return nil, err


shurcooL:
Thanks for reporting this and sending a PR.

Do you have any suggestions for how I can verify that this is a problem, and that the PR fixes it? I want to confirm your findings.

If I understand correctly, the way to verify this would be to try to import a standard library package that uses nosync (e.g., `math/rand`) and avoid having `sync` imported via other packages. Due to the bug #745, `sync` package will be included in the build anyway. With this fix, it wouldn't be, making the output size smaller. Is that right?

dave:
So, it doesn't add anything extra to the JS output, but it does a whole lot of compiling that's not needed. If you add some logging in [here](https://github.com/gopherjs/gopherjs/blob/00a3767085dffc9baa5e1c320b28e3629a56d323/build/build.go#L529-L532) like this:

```go
func (s *Session) BuildPackage(pkg *PackageData) (*compiler.Archive, error) {
	if archive, ok := s.Archives[pkg.ImportPath]; ok {
		fmt.Println("found", pkg.ImportPath, "in cache")
		return archive, nil
	}
	fmt.Println("building", pkg.ImportPath)
```

... and create a little file like this `main.go`:

```go
package main

import "math/rand"

func main() {
	rand.Int()
}
```

... and run `gopherjs build main.go`... With the master branch you get this output:

```
$ gopherjs build rand.go
building main
building math/rand
building math
building github.com/gopherjs/gopherjs/js
building sync
building internal/race
building runtime
building runtime/internal/sys
found github.com/gopherjs/gopherjs/js in cache
building sync/atomic
found github.com/gopherjs/gopherjs/js in cache
found internal/race in cache
found sync/atomic in cache
found runtime in cache
found github.com/gopherjs/gopherjs/js in cache
found math in cache
building github.com/gopherjs/gopherjs/nosync
```

... but with the PR branch you get this:

```
$ gopherjs build rand.go
building main
building math/rand
building math
building github.com/gopherjs/gopherjs/js
building github.com/gopherjs/gopherjs/nosync
found math in cache
found github.com/gopherjs/gopherjs/nosync in cache
building runtime
building runtime/internal/sys
found github.com/gopherjs/gopherjs/js in cache
```

shurcooL:
Perfect, that's very helpful, thanks!

flimzy:
@nevkontakte and I have decided to close this issue, pending a deeper investigation and some improvements to the logic this touches, which we believe will make this fix obsolete.

PR #747, opened 2018-02-02
build: Add "purego" build tag to default context.

### LINKED ISSUE #746: Use purego build tag
The proposal in golang/go#23172 was accepted. The `purego` build tag is intended to be a community agreed upon soft-signal to indicate the forbidden use of unsafe, assembly, or cgo. Can gopherjs start respecting the tag when filtering Go source files when building?

shurcooL:
The proposal in golang/go#23172 was accepted. The "purego" build tag
is intended to be a community agreed upon soft-signal to indicate the
forbidden use of unsafe, assembly, or cgo.

Since GopherJS doesn't support those, it should always use "purego"
versions of Go packages when they're available. This is done by always
setting the "purego" build constraint.

Fixes #746.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -41,7 +41,7 @@ func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
 		GOARCH:        "js",
 		InstallSuffix: installSuffix,
 		Compiler:      "gc",
-		BuildTags:     append(buildTags, "netgo"),
+		BuildTags:     append(buildTags, "netgo", "purego"),
 		ReleaseTags:   build.Default.ReleaseTags,
 		CgoEnabled:    true, // detect `import "C"` to throw proper error
 	}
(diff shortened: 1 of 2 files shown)


build/build.go
```diff
@@ -41,7 +41,7 @@ func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
 		GOARCH:        "js",
 		InstallSuffix: installSuffix,
 		Compiler:      "gc",
-		BuildTags:     append(buildTags, "netgo"),
+		BuildTags:     append(buildTags, "netgo", "purego"),
```
dsnet: Just saw this and I realized I have no idea what `netgo` is. Perhaps separate these out onto each of their own line and comment with a reference to what they are and/or why they are set?

shurcooL: Sure, I can document them.

"netgo" is a build tag used in `net` package to get a Go (rather than cgo) version of the `net` package, using Go's DNS resolver rather than a cgo DNS resolver. It's documented at https://godoc.org/net#hdr-Name_Resolution, and you can also grep https://gotools.org/net for "netgo" to see it in action.

It was added to GopherJS in commit 273a70496de06c38fddef4a8b2dd31d0446dcb51.

shurcooL: Done in e9ae517; PTAL.

dsnet: LGTM

PR #752, opened 2018-02-11
internal/sysutil: Fix build error on FreeBSD due to Rlimit.Cur type mismatch.

juergenhoetzel:
Trying to build [Perkeep](https://github.com/perkeep/perkeep) on FreeBSD:

```
➜ perkeep git:(master) go run make.go 
Now rebuilding gopherjs at /usr/home/juergen/perkeep/tmp/build-gopath-nosqlite/bin/gopherjs
error building gopherjs: error while building gopherjs: exit status 2, # perkeep.org/vendor/github.com/gopherjs/gopherjs/inter\
nal/sysutil
internal/sysutil/sysutil.go:12:10: cannot use r.Cur (type int64) as type uint64 in return argument
exit status 1

```

### DIFF
--- a/internal/sysutil/sysutil.go
+++ b/internal/sysutil/sysutil.go
@@ -9,5 +9,5 @@ import "golang.org/x/sys/unix"
 func RlimitStack() (cur uint64, err error) {
 	var r unix.Rlimit
 	err = unix.Getrlimit(unix.RLIMIT_STACK, &r)
-	return uint64(r.Cur), err
+	return uint64(r.Cur), err // Type conversion because Cur is one of uint64, int64 depending on unix flavor.
 }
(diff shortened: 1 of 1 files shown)


internal/sysutil/sysutil.go
```diff
@@ -1,4 +1,4 @@
-// +build !windows
+// +build !windows, !freebsd
```
shurcooL: As a heads up, this wouldn't be right. Both commas and spaces are significant, so you shouldn't use both. See https://godoc.org/go/build#hdr-Build_Constraints:

> A build constraint is evaluated as the OR of space-separated options; each option evaluates as the AND of its comma-separated terms

juergenhoetzel:
> I think a simpler fix is to just always do a type conversion in syscall.go. If you don't have objections, let's go with that for now, until further deviations between flavors of unix are needed.

Indeed a much simpler solution. I amended to commit. Thanks for your feedback!

Issue #759, opened 2018-02-24
Reproducible output: image/draw


dave:
While developing jsgo, I noticed that the entire stdlib builds reproducibly apart from `image/draw`. Only the un-minified code shows the problem.

This switch statement: https://github.com/golang/go/blob/master/src/image/draw/draw.go#L608-L615

... the variables `src0$1`, `src0$2` and `src0` are re-arranged.

<img width="1142" alt="screen shot 2018-02-24 at 08 43 40" src="https://user-images.githubusercontent.com/925351/36627161-1486867e-193f-11e8-84c6-d6d4c2cea3ac.png">

I'll dig into the compiler code and try to work out what's wrong.

dave:
So the problem seems to be that [this](https://github.com/gopherjs/gopherjs/blob/95deb33d587c9f6e24b494ea9bdf9648c48f9a60/compiler/utils.go#L371) is returning several copies of the `src0` object... These have identical `Name` and `Pos()`, so can't be sorted properly [here](https://github.com/gopherjs/gopherjs/blob/95deb33d587c9f6e24b494ea9bdf9648c48f9a60/compiler/utils.go#L373-L376).

dave:
https://github.com/gopherjs/gopherjs/issues/759 fixes this bug. I have a suspicion that `EscapingObjects` shouldn't be returning these objects at all, but this is way out of my sphere of knowledge, so this is a light-touch fix that won't change the behaviour.

dave:
I'm running a few final checks to see if it resolves the reproducible code bug and doesn't introduce any more...

dave:
Yes, this after this PR, I get reproducible JS across the whole standard library. @shurcooL

shurcooL:
> #759 fixes this bug.

That's this issue. Did you mean PR #760?

Also, how does this intersect with #742? I haven't gotten around to reviewing it properly, but I want to soon.

dave:
Yes sorry #760 is the right PR. I believe this is unconnected to #742.

PR #760, opened 2018-02-24
Improve ordering of EscapingObjects

dave:
EscapingObjects [returns a slice](https://github.com/gopherjs/gopherjs/blob/95deb33d587c9f6e24b494ea9bdf9648c48f9a60/compiler/analysis/escape.go#L9), which is created [from a map](https://github.com/gopherjs/gopherjs/blob/95deb33d587c9f6e24b494ea9bdf9648c48f9a60/compiler/analysis/escape.go#L18-L20) so the order is undefined. It is then [sorted here](https://github.com/gopherjs/gopherjs/blob/95deb33d587c9f6e24b494ea9bdf9648c48f9a60/compiler/utils.go#L372-L377). 

This PR improves the situation by maintaining the order that the objects were found in the source. This means that the sorting in `utils.go` is not needed. 

This fixes https://github.com/gopherjs/gopherjs/issues/759

### DIFF
--- a/compiler/analysis/escape.go
+++ b/compiler/analysis/escape.go
@@ -14,16 +14,13 @@ func EscapingObjects(n ast.Node, info *types.Info) []*types.Var {
 		bottomScopes: make(map[*types.Scope]bool),
 	}
 	ast.Walk(&v, n)
-	var list []*types.Var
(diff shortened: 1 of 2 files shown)


flimzy:
Merged as #1205

PR #761, opened 2018-02-25
Minify prelude

dave:
This PR adds minified versions of the prelude JS.  This saves 8910 bytes on minified output. We can compare the size of the outputted JS for a trivial hello world:

```go
package main

func main() {
  println("Hello, World!")
}
```

`gopherjs build -m main.go`

Master: 51170 bytes
This PR: 42260 bytes


dave:
I did the minifying with https://skalman.github.io/UglifyJS-online/ - we need to remember to update the minified versions whenever updating the un-minified prelude! @shurcooL

myitcv:
@dave 💯 👍 to what you're proposing here. Would love to benefit from this too.

> we need to remember to update the minified versions whenever updating the un-minified prelude

I think we can automate this fairly easily and add a check to `circle.yml` to ensure the committed minified code is consistent with the unminified code. 

Something along the lines of:

* add a `package.json` to the repo root, have `uglify-es` as a dev dependency
* add a `// +build ignore`-d `main`-package `.go` file to the `compiler/prelude` directory that outputs `github.com/gopherjs/gopherjs/compiler/prelude.Prelude` which can then be piped to `uglifyjs` and then written to a generated file that defines `github.com/gopherjs/gopherjs/compiler/prelude.PredudeMinified` in another file in `compiler/prelude`
* add a step to `circle.yml` to re-run this generation step via `go run`, and then check that `git status --porcelain` is the empty string (else fail the build)

Makes things a little less fragile.

Happy to help flesh this out further if it sounds reasonable?

compiler/compiler.go
```diff
@@ -154,7 +154,13 @@ func WriteProgramCode(pkgs []*Archive, w *SourceMapFilter) error {
...
 		return err
 	}
-	if _, err := w.Write(removeWhitespace([]byte(prelude.Prelude), minify)); err != nil {
+	var preludeString string
+	if minify {
+		preludeString = prelude.PreludeMinified
+	} else {
+		preludeString = prelude.Prelude
+	}
+	if _, err := w.Write([]byte(preludeString)); err != nil {
```
shurcooL: You can use `io.WriteString` here.

dave:
Done!

myitcv:
> We can automate later if needed—it's not worth all the extra machinery overhead now.

I don't think the extra machinery creates that much overhead; I just fleshed out what I outlined above:

https://github.com/gopherjs/gopherjs/pull/765

And the benefit is that it saves having to manually re-run minification at all.

PR #765, opened 2018-02-27
Example of regenerated minified files

myitcv:
**DO NOT MERGE**

Example of auto-regenerated prelude files; potential follow-up to #761


shurcooL:
Thanks for prototyping this @myitcv. This approach looks pretty much as I expected. 👍 I like the idea of using `npm install` to get the `uglify-es` dependency.

I still prefer not to automate this just yet, because the prelude doesn't change often and it's not much harder to do it as done in #761.

Should that change, we can reopen this PR and resume here. Thanks.

myitcv:
@shurcooL

> I still prefer not to automate this just yet, because the prelude doesn't change often and it's not much harder to do it as done in #761.

As I understand things going the extra step to automate this (especially given the prototype confirms it's straightforward) is worthwhile because:

* the alternative requires multiple manual steps (for each part of the prelude that changes) 
* it's impossible to tell if things are out of sync without running all of the manual steps
* requires someone to remember that the manual step needs to be re-run... which requires someone to remember to check this during review time
* the linked tool may change (_far_ less likely with pinned `npm` deps)

Unless I'm missing something in the way the alternative works?

shurcooL:
I think it's worthwhile if the prelude changes often. [Historically](https://github.com/gopherjs/gopherjs/commits/master/compiler/prelude), it has changed very little (a few times in the last few years), although the frequency might be increasing recently.

I think it's fair to wait until it changes 2-3 more times before spending time on automating it. To me, that's easier.

myitcv:
> I think it's fair to wait until it changes 2-3 more times before going through with automating it. To me, that's easier.

I'm not entirely sure I follow, particularly given the experience of #775 (specifically https://github.com/gopherjs/gopherjs/pull/775#discussion_r174611168)

Because the work to automate it is already done; that's what this PR represents.

Or am I missing something?

Can I suggest I rebase this PR, ensure CI works and then we see what the diff looks like?

PR #768, opened 2018-03-04
Use high resolution time API

hajimehoshi:
High resolution time API is available in most browsers: https://caniuse.com/#feat=high-resolution-time

As the Go time is expected to have nano precision (e.g. `Nano()` function), I think it is natural to return more precise time than `Date.getTime`.

### DIFF
--- a/compiler/natives/src/time/time.go
+++ b/compiler/natives/src/time/time.go
@@ -39,7 +39,7 @@ func initLocal() {
 }
 
 func runtimeNano() int64 {
-	return js.Global.Get("Date").New().Call("getTime").Int64() * int64(Millisecond)
+	return int64(js.Global.Get("performance").Call("now").Float() * float64(time.Millisecond))
 }
 
 func now() (sec int64, nsec int32, mono int64) {


flimzy:
Sounds good on the surface. I wonder if it would be worth supporting a fall-back to the current method for the few old browsers that don't support it?  (I don't know of any official browser-compatibility target for GopherJS that would make answering this question easy).

flimzy:
Looking at it more closely, though, `performance.now` isn't at all the same as `Date.getTime`. 
 In particular, `performance.now` returns the age of the current document, whereas `getTime` returns the system time. Further, as [documented](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now):

> unlike `Date.now()`, the values returned by `Performance.now()` always increase at a constant rate, independent of the system clock (which might be adjusted manually or skewed by software like NTP). Otherwise, `performance.timing.navigationStart + performance.now()` will be approximately equal to `Date.now()`.

So from reading this, I think it's pretty clear that the current PR is not correct, and even an adjusted PR which accounts for `performance.timing.navigationStart` will be a behavioral change. Whether that change is appropriate is another question.

I'm not sure of the details on how Go's clock is expected to behave wrt system time vs elapsed time, except that I know Go 1.9 added a monotonic feature to the clock, which probably relates to `performance.now()`'s functionality.

hajimehoshi:
> Looking at it more closely, though, performance.now isn't at all the same as Date.getTime.
In particular, performance.now returns the age of the current document, whereas getTime returns the system time.

Ah yes, I missed that. Thank you for pointing it out.

> even an adjusted PR which accounts for performance.timing.navigationStart will be a behavioral change.

Isn't this a precision matter? Or would this be semantically different?

flimzy:
> Isn't this a precision matter? Or would this be semantically different?

It's semantically different. In the case where the system clock changes, subsequent calls to `Date.getTime()` will advance at a different rate, possibly even going backwards, whereas subsequent calls to `performance.now()` are guaranteed always to advance at the same rate.

hajimehoshi:
Thank you for elaborating. It'd not be so easy to make timer precise without breaking the current behavior. Let me take down this PR.

shurcooL:
> I know Go 1.9 added a monotonic feature to the clock, which probably relates to performance.now()'s functionality.

Indeed. I think it might be possible to use `performance.now` for just the monotonic component of time. I mentioned that in the commit message of https://github.com/gopherjs/gopherjs/pull/651/commits/63f374fec74816f70e158c128baab044a7445310.

antong:
I tested using `performance.now()` for the monotonic clock and did get better resolution for stuff like `t.Since(t0)`.

```
diff --git a/compiler/natives/src/time/time.go b/compiler/natives/src/time/time.go
index e898897..5220f01 100644
--- a/compiler/natives/src/time/time.go
+++ b/compiler/natives/src/time/time.go
@@ -42,9 +42,13 @@ func runtimeNano() int64 {
        return js.Global.Get("Date").New().Call("getTime").Int64() * int64(Millisecond)
 }

+func nanotime() int64 {
+       return int64(js.Global.Get("performance").Call("now").Float() * float64(Millisecond))
+}
+
 func now() (sec int64, nsec int32, mono int64) {
        n := runtimeNano()
-       return n / int64(Second), int32(n % int64(Second)), n
+       return n / int64(Second), int32(n % int64(Second)), nanotime()
 }

 func Sleep(d Duration) {

```

flimzy:
Using `performance.now()` for the monotonic portion of the time struct might make sense.

PR #775, opened 2018-03-13
compiler/prelude: Bug fix: specifying out of range index must panic

### LINKED ISSUE #732: bug: len(byteArray) can be minus without panic
```go
package main

import (
	"fmt"
)

func main() {
	a := make([]byte, 4096)
	b := a[8192:]
	fmt.Println(len(b))
}
```

https://gopherjs.github.io/playground/#/Sxg5Yfe4Y0

Actual result:
-4096

Expected result:
panic: runtime error: slice bounds out of range

hajimehoshi:
Fixes #732


shurcooL:
CI is failing. You've modified prelude, which means you'll need to update the minified prelude (manually for now, see discussion in #765). I suspect that's why.

Try to follow the steps outlined in the commit message of 9c698606f4000dee7999cacc64ebaccf1e79e288 and let me know if you run into any issues.

hajimehoshi:
Hmm, so this is not automated yet?

hajimehoshi:
Done

compiler/prelude/prelude.go
```diff
@@ -4,7 +4,7 @@ const Prelude = prelude + numeric + types + goroutines + jsmapping
...
...if(closedDuringSend){$throwRuntimeError("send on closed channel")}}}};var $recv=function(chan){var queuedSend=chan.$sendQueue.shift();if(queuedSend!==undefined){chan.$buffer.push(queuedSend(false))}var bufferedValue=chan.$buffer.shift();if(bufferedValue!==undefined){return[bufferedValue,true]}if(chan.$closed){return[chan.$elem.zero(),false]}var thisGoroutine=$curGoroutine;var f={$blk:function(){return this.value}};var queueEntry=function(v){f.value=v;$schedule(thisGoroutine)};chan.$recvQueue.push(queueEntry);$block();return f};var $close=function(chan){if(chan.$closed){$throwRuntimeError("close of closed channel")}chan.$closed=true;while(true){var queuedSend=chan.$sendQueue.shift();if(queuedSend===undefined){break}queuedSend(true)}while(true){var queuedRecv=chan.$recvQueue.shift();if(queuedRecv===undefined){break}queuedRecv([chan.$elem.zero(),false])}};var $select=function(comms){var ready=[];var selection=-1;for(var i=0;i<comms.length;i++){var comm=comms[i];var chan=comm[0];switch(comm.length){case 0:selection=i;break;case 1:if(chan.$sendQueue.length!==0||chan.$buffer.length!==0||chan.$closed){ready.push(i)}break;case 2:if(chan.$closed){$throwRuntimeError("send on closed channel")
```
shurcooL: I'm not getting the same minified output as you @hajimehoshi. My output is exactly 32740 bytes. Yours is 41045, which is significantly larger than the previous size of 32762.

How did you create this? Did you minify the value of the `Prelude` constant or something else?

The way I do it is with the help of [`goexec`](https://github.com/shurcooL/goexec#readme):

```
$ goexec -quiet 'io.WriteString(os.Stdout, prelude.Prelude)' | pbcopy
```

That prints the value of the `Prelude` constant to stdout, and copies it to my clipboard (on macOS). Then I paste that at https://skalman.github.io/UglifyJS-online/ and get the output.

hajimehoshi: I used command-line uglifyjs installed by node, and I think that's why the difference caused.

hajimehoshi:
Updated

compiler/prelude/prelude.go
```diff
@@ -99,19 +99,19 @@ var $ifaceMethodExpr = function(name) {
 };
 
 var $subslice = function(slice, low, high, max) {
+  if (high === undefined) {
+    high = slice.$length;
+  }
+  if (max === undefined) {
+    max = slice.$capacity
```
shurcooL: This line is missing a semicolon at the end.

tests/misc_test.go
```diff
@@ -623,3 +623,15 @@ func TestTypeConversion(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestSliceOutOfRange(t *testing.T) {
```
shurcooL: What do you think about moving this test and placing it next to `TestSliceOfString`, since it's quite similar and related?

tests/misc_test.go
```diff
@@ -623,3 +623,15 @@ func TestTypeConversion(t *testing.T) {
...
+	defer func() {
+		if err := recover(); err == nil || !strings.Contains(err.(error).Error(), "slice bounds out of range") {
+			t.Fail()
+		}
+	}()
+
+	a := make([]byte, 4096)
+	b := a[8192:]
```
shurcooL: There's no need to use numbers so large, the same issue would happen with `make([]byte, 4096)` and `a[8:]`, right? Maybe it's better to use those smaller numbers to make the test focus more on the underlying issue and less resource intensive.

compiler/prelude/prelude.go
```diff
@@ -4,7 +4,7 @@ const Prelude = prelude + numeric + types + goroutines + jsmapping
...
...Function:var i=$funcType([$sliceType($emptyInterface)],[$jsObjectPtr],!0);return new i($internalize(e,i));case Number:return new $Float64(parseFloat(e));case String:return new $String($internalize(e,$String));default:if($global.Node&&e instanceof $global.Node)return new $jsObjectPtr(e);var a=$mapType($String,$emptyInterface);return new a($internalize(e,a))}case $kindMap:for(var o={},$=$keys(e),c=0;c<$.length;c++){var u=$internalize($[c],n.key);o[n.key.keyFor(u)]={k:u,v:$internalize(e[$[c]],n.elem)}}return o;case $kindPtr:if(n.elem.kind===$kindStruct)return $internalize(e,n.elem);case $kindSlice:return new n($mapArray(e,function(e){return $internalize(e,n.elem)}));case $kindString:if(e=String(e),$isASCII(e))return e;var l="";for(c=0;c<e.length;){var s=e.charCodeAt(c);if(55296<=s&&s<=56319){var f=e.charCodeAt(c+1);l+=$encodeRune(1024*(s-55296)+f-56320+65536),c+=2}else l+=$encodeRune(s),c++}return l;case $kindStruct:var d={},p=function(n){if(n===$jsObjectPtr)return e;switch(n===$jsObjectPtr.elem&&$throwRuntimeError("cannot internalize js.Object, use *js.Object instead"),n.kind){case $kindPtr:return p(n.elem);case $kindStruct:var r=n.fields[0],t=p(r.typ);if(t!==d){var i=new n.ptr;retur
```
shurcooL: That's more like it. 👍

compiler/prelude/prelude.go
hajimehoshi: Done.

tests/misc_test.go
hajimehoshi: Done.

tests/misc_test.go
hajimehoshi: I don't have a strong opinion on it :-) Done.

PR #779, opened 2018-03-24
compiler: support arbitrary value js struct tag values

### LINKED ISSUE #778: compiler fails to handle arbitrary string js tags
Consider the following test:

```go
type StructWithNonIdentifierJsTags struct {
        object *js.Object
        Name   string `js:"my name"`
}

func TestStructWithNonIdentifierJsTags(t *testing.T) {
        s := StructWithNonIdentifierJsTags{
                object: js.Global.Get("Object").New(),
        }

        const want = "Paul"

        // externalise a value
        s.Name = want

        // internalise again
        got := s.Name

        if want != got {
                t.Errorf("Value via non-identifier js tag field gave %q, want %q", got, want)
        }
}
```

This currently fails at runtime (`gopherjs test`) with:

```
/home/myitcv/.mountpoints/myitcv_io_react/src/myitcv.io/react/_vendor/src/github.com/gopherjs/gopherjs/test.554077738:28596
                s.object.my name = $externalize("Paul", $String);
                            ^^^^

SyntaxError: Unexpected identifier
    at new Script (vm.js:51:7)
    at createScript (vm.js:138:10)
    at Object.runInThisContext (vm.js:199:10)
    at Module._compile (module.js:624:28)
    at Object.Module._extensions..js (module.js:671:10)
    at Module.load (module.js:573:32)
    at tryModuleLoad (module.js:513:12)
    at Function.Module._load (module.js:505:3)
    at Function.Module.runMain (module.js:701:10)
    at startup (bootstrap_node.js:193:16)
FAIL    github.com/gopherjs/gopherjs/js 0.185s
```

This is because the compiler currently translates field access (getting or setting) via the dot notation:

```
s.object.my name
```

Where a string contains a non-identifier value (e.g. space above) this is destined to fail. 

Some external libraries (e.g. React) require properties to be set that are composed of non-identifier strings, e.g. https://reactjs.org/docs/dom-elements.html, where `"data-*"` and `"aria-*"` are used for attribute names. 

An equivalent way to set properties is via the bracket notation:

```
s.object.["my name"]
```

Hence it probably makes more sense in general to use this for js tag compilation. 

For more details see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors

PR that addresses this will follow shortly.

myitcv:
Javascript allows for arbitrary strings to be used as index values on an
Object. A good example of where this comes up is React where `"data-*"`
and `"aria-*"` indexes are used on objects
(https://reactjs.org/docs/dom-elements.html). 

By switching from property selectors to the index operator (where required) we can support arbitrary string values in js-tagged fields for `*js.Object` special structs.

Fixes #778

### DIFF
--- a/compiler/utils.go
+++ b/compiler/utils.go
@@ -646,22 +646,28 @@ func encodeIdent(name string) string {
 	return strings.Replace(url.QueryEscape(name), "%", "$", -1)
 }
 
-// formatJSStructTagVal returns a string of JavaScript code appropriate for
-// accessing the property identified by jsTag. If the jsTag value can be safely
-// encoded in JavaScript using the dot notation this is used, else we fallback
-// to the bracket notation. For more details see:
+// formatJSStructTagVal returns JavaScript code for accessing an object's property
+// identified by jsTag. It prefers the dot notation over the bracket notation when
+// possible, since the dot notation produces slightly smaller output.
 //
-// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors
+// For example:
 //
-// Uses definition of an identifier from
-// https://developer.mozilla.org/en-US/docs/Glossary/Identifier
+// 	"my_name" -> ".my_name"
+// 	"my name" -> `["my name"]`
+//
+// For more information about JavaScript property accessors and identifiers, see
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors and
+// https://developer.mozilla.org/en-US/docs/Glossary/Identifier.
+//
 func formatJSStructTagVal(jsTag string) string {
 	for i, r := range jsTag {
 		ok := unicode.IsLetter(r) || (i != 0 && unicode.IsNumber(r)) || r == '$' || r == '_'
 		if !ok {
-			return "[\"" + template.JSEscapeString(jsTag) + "\"]"
+			// Saw an invalid JavaScript identifier character,
+			// so use bracket notation.
+			return `["` + template.JSEscapeString(jsTag) + `"]`
 		}
 	}
-
+	// Safe to use dot notation without any escaping.
 	return "." + jsTag
 }
--- a/compiler/utils.go
+++ b/compiler/utils.go
@@ -656,19 +656,12 @@ func encodeIdent(name string) string {
 // Uses definition of an identifier from
 // https://developer.mozilla.org/en-US/docs/Glossary/Identifier
 func formatJSStructTagVal(jsTag string) string {
-	useDot := true
-
 	for i, r := range jsTag {
 		ok := unicode.IsLetter(r) || (i != 0 && unicode.IsNumber(r)) || r == '$' || r == '_'
 		if !ok {
-			useDot = false
-			break
+			return "[\"" + template.JSEscapeString(jsTag) + "\"]"
 		}
 	}
 
-	if useDot {
-		return "." + jsTag
-	}
-
-	return "[\"" + template.JSEscapeString(jsTag) + "\"]"
+	return "." + jsTag
 }
--- a/compiler/utils.go
+++ b/compiler/utils.go
@@ -12,6 +12,8 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"text/template"
+	"unicode"
 
 	"github.com/gopherjs/gopherjs/compiler/analysis"
 	"github.com/gopherjs/gopherjs/compiler/typesutil"
@@ -642,4 +644,31 @@ func endsWithReturn(stmts []ast.Stmt) bool {
 
 func encodeIdent(name string) string {
 	return strings.Replace(url.QueryEscape(name), "%", "$", -1)
+}
+
+// formatJSStructTagVal returns a string of JavaScript code appropriate for
+// accessing the property identified by jsTag. If the jsTag value can be safely
+// encoded in JavaScript using the dot notation this is used, else we fallback
+// to the bracket notation. For more details see:
+//
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors
+//
+// Uses definition of an identifier from
+// https://developer.mozilla.org/en-US/docs/Glossary/Identifier
+func formatJSStructTagVal(jsTag string) string {
+	useDot := true
+
+	for i, r := range jsTag {
+		ok := unicode.IsLetter(r) || (i != 0 && unicode.IsNumber(r)) || r == '$' || r == '_'
+		if !ok {
+			useDot = false
+			break
+		}
+	}
+
+	if useDot {
+		return "." + jsTag
+	}
+
+	return "[\"" + template.JSEscapeString(jsTag) + "\"]"
 }
--- a/js/js_test.go
+++ b/js/js_test.go
@@ -598,33 +598,25 @@ func TestTypeSwitchJSObject(t *testing.T) {
 	}
 }
 
-type StructWithNonIdentifierJsTags struct {
-	object *js.Object
-	Name   string `js:"@&\"'<>//my name"`
-}
-
-func TestStructWithNonIdentifierJsTags(t *testing.T) {
-	s := StructWithNonIdentifierJsTags{
-		object: js.Global.Get("Object").New(),
+func TestStructWithNonIdentifierJSTag(t *testing.T) {
+	type S struct {
+		*js.Object
+		Name string `js:"@&\"'<>//my name"`
 	}
+	s := S{Object: js.Global.Get("Object").New()}
 
-	const want = "Paul"
+	// externalise a value via field
+	s.Name = "Paul"
 
-	// externalise a value
-	s.Name = want
-
-	// internalise again
+	// internalise via field
 	got := s.Name
-
-	if want != got {
-		t.Errorf("Value via non-identifier js tag field gave %q, want %q", got, want)
+	if want := "Paul"; got != want {
+		t.Errorf("value via field with non-identifier js tag gave %q, want %q", got, want)
 	}
 
 	// verify we can do a Get with the struct tag
-	got = s.object.Get("@&\"'<>//my name").String()
-
-	if want != got {
-		t.Errorf("Value via .Get gave %q, want %q", got, want)
+	got = s.Get("@&\"'<>//my name").String()
+	if want := "Paul"; got != want {
+		t.Errorf("value via js.Object.Get gave %q, want %q", got, want)
 	}
-
 }
--- a/js/js_test.go
+++ b/js/js_test.go
@@ -597,3 +597,34 @@ func TestTypeSwitchJSObject(t *testing.T) {
 		}
 	}
 }
+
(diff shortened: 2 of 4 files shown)


js/js_test.go
```diff
@@ -597,3 +597,26 @@ func TestTypeSwitchJSObject(t *testing.T) {
...
+
+	// externalise a value
+	s.Name = want
+
+	// internalise again
+	got := s.Name
+
+	if want != got {
+		t.Errorf("Value via non-identifier js tag field gave %q, want %q", got, want)
+	}
```
shurcooL: Should we also add coverage for `js.Object.Get` and `js.Object.Set` methods? Are they equally affected?

myitcv: No, they are not affected: https://gopherjs.github.io/playground/#/osNIk-Yqbn

[Here's the relevant line in the compiler](https://github.com/gopherjs/gopherjs/blob/82b322028c96512b15077093b16a5f1c7ea897ac/compiler/expressions.go#L601-L610)

myitcv:
> Can you check how much of an impact this change makes on the generated file size of a moderately sized program compiled with GopherJS? With minification, before and after gzip compression.

Good point. I think it's probably harder to come up with a definition for a representative "moderately sized program" that to make the change, so I made the change to implement this fallback logic. Please take a look.

myitcv:
Gentle ping @shurcooL - thanks.

compiler/utils.go
```diff
@@ -643,3 +644,38 @@ func endsWithReturn(stmts []ast.Stmt) bool {
...
+// jsTagFormat returns a string of Javascript code appropriate for accessing
+// the property identified by jsTag. If the jsTag value can be safely encoded
+// in Javascript using the dot notation this is used, else we fallback to the
+// bracket notation. For more details see:
+//
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors
+//
+// Uses definition of an identifier from
+// https://developer.mozilla.org/en-US/docs/Glossary/Identifier
+func jsTagFormat(jsTag string) string {
```
flimzy: This is entirely a nitpick, so ignore it if you wish... But I wonder if `formatJSTag` would be a better name, and more consistent with the other func names here, which use the `<verb><object>` convention (i.e. `formatExpr` and `encodeIdent` etc)

myitcv: Done.

myitcv:
@hajimehoshi - any thoughts?

Gentle ping @shurcooL

hajimehoshi:
I'll look at this.

>  js tags

struct tags for js?

compiler/utils.go
```diff
@@ -643,3 +644,38 @@ func endsWithReturn(stmts []ast.Stmt) bool {
...
+// formatJSTag returns a string of Javascript code appropriate for accessing
+// the property identified by jsTag. If the jsTag value can be safely encoded
+// in Javascript using the dot notation this is used, else we fallback to the
+// bracket notation. For more details see:
+//
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors
+//
+// Uses definition of an identifier from
+// https://developer.mozilla.org/en-US/docs/Glossary/Identifier
+func formatJSTag(jsTag string) string {
```
hajimehoshi: I feel like the word "js tag" is confusing. Isn't this actually a struct tag for js? How about `formatStructTagForJS`? (I'm not an English native speaker so my suggestion might not sound natural)

js/js_test.go
```diff
@@ -597,3 +597,26 @@ func TestTypeSwitchJSObject(t *testing.T) {
 		}
 	}
 }
+
+type StructWithNonIdentifierJsTags struct {
+	object *js.Object
+	Name   string `js:"my name"`
```
hajimehoshi: How about adding more test cases like starting with a char that is not for IDs?

compiler/utils.go
```diff
@@ -643,3 +644,38 @@ func endsWithReturn(stmts []ast.Stmt) bool {
...
+	useDot := true
+
+RuneRange:
```
shurcooL: Trivial style nitpick, maybe use `Outer` label. It'd be more consistent with:

https://github.com/gopherjs/gopherjs/blob/558a9132744c22476178edf3126fd35a9754f565/build/build.go#L162

(In general, `RuneRange` is not a label I see often, but `Outer` is pretty common.)

js/js_test.go
hajimehoshi: As there are such tests, the results should be changed?

myitcv:
@hajimehoshi I've tweaked the wording of the description (and commit) to be more precise and inline with https://github.com/gopherjs/gopherjs/wiki/JavaScript-Tips-and-Gotchas

Revised commit now pushed up. PTAL

compiler/utils.go
myitcv: Done

js/js_test.go
myitcv: I've made the test more "exciting" - I don't think any more are required because if we were doing the wrong thing then the generated Javascript would not parse.

compiler/utils.go
myitcv: Addressed in the latest PR - I chose a slightly different version to keep more consistency with https://github.com/gopherjs/gopherjs/wiki/JavaScript-Tips-and-Gotchas

compiler/utils.go
```diff
@@ -643,3 +644,38 @@ func endsWithReturn(stmts []ast.Stmt) bool {
...
+				break Outer
+			}
+		}
+	}
+
+	if useDot {
+		return "." + jsTag
+	}
+
+	return "[\"" + jsTag + "\"]"
```
hajimehoshi: Don't we need escaping some chars like backslash? `text/template.JSEscapeString` is needed here.

hajimehoshi: Hmm, there are very subtle differences between `strconv.Quote` and `template.JSEscapeString`...

1. `<` and `>` is treated as special chars in `JSEscapeString`
2. `strconv.Quote` can produce \U000XXXXX for non-BMP and non-printable char, which is not valid escaping in ES5

myitcv: You're absolutely right - thank you for correcting me!

I've just added a test in the latest commit that does a `.Get` in order to ensure the property that has been set via the js struct tag is using the expected key.

PTAL

compiler/utils.go
```diff
@@ -12,7 +12,9 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"unicode"
 
+	"github.com/alecthomas/template"
```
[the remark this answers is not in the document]
hajimehoshi: Oh I missed that... Thanks. I think `text/template` is enough (`html/template` is also fine and does the same thing).

compiler/utils.go
```diff
@@ -643,3 +645,38 @@ func endsWithReturn(stmts []ast.Stmt) bool {
...
+// accessing the property identified by jsTag. If the jsTag value can be safely
+// encoded in Javascript using the dot notation this is used, else we fallback
+// to the bracket notation. For more details see:
+//
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors
+//
+// Uses definition of an identifier from
+// https://developer.mozilla.org/en-US/docs/Glossary/Identifier
+func formatJSStructTagVal(jsTag string) string {
+	useDot := true
```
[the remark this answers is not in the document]
myitcv: I'm not sure I agree. We want the default behaviour to be to use the dot notation because, as noted before, this is more efficient in output terms. "Dot notation" is taken directly from the MDN. To my mind it's better to have the variable name reflect the intended semantics, hence `useDot`.

myitcv:
@shurcooL - thanks for reviewing. I've just pushed up some changes, comments against one of your bits of feedback left above.

compiler/utils.go
```diff
@@ -643,3 +645,38 @@ func endsWithReturn(stmts []ast.Stmt) bool {
...
+				break Outer
+			}
+		}
+	}
+
+	if useDot {
+		return "." + jsTag
+	}
+
+	return "[\"" + template.JSEscapeString(jsTag) + "\"]"
```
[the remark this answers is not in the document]
flimzy: Why not just eliminate the variable entirely, and further shorten the function?

    func formatJSStructTagVal(jsTag string) string {
        for i, r := range jsTag {
            if unicode.IsLetter(r) || (i != 0 && unicode.IsNumber(r)) || r == '$' || r == '_' {
                return "[\"" + template.JSEscapeString(jsTag) + "\"]"
            }
        }

        return "." + jsTag
    }

compiler/utils.go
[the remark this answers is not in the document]
shurcooL: > We want the default behaviour to be to use the dot notation because, as noted before, this is more efficient in output terms.

My suggestion doesn't contradict that. `needsEscaping` controls whether non-default behavior of using escaping is used. If its value remains `false`, the default behavior (using dot notation) is used.

It's similar to when you handle error conditions (non-default behaviors), you typically write:

```Go
f, err := os.Open("file.txt")
if err != nil {
    // less common (non-default) path
    return fmt.Errorf("couldn't open file.txt: %v", err)
}
// normal (default) path
```

Instead of:

```Go
successfullyOpened := true
f, err := os.Open("file.txt")
if err != nil {
    successfullyOpened = false
}
if successfullyOpened {
    // normal (default) path
}
// less common (non-default) path
return fmt.Errorf("couldn't open file.txt: %v", err)
```

shurcooL:
> I've just pushed up some changes

I saw the comments, but not seeing changes. This PR currently has 1 commit, so it's hard to tell what has changed (if anything) from last time.

If you don't mind, please avoid squashing and force pushing to the PR branch, but rather feel free to push additional commits. We can easily squash them into one commit when merging the PR. It makes reviewing easier, because the diffs show only what has changed since last time. (Unfortunately, GitHub isn't as good as Gerrit in this sense.)

compiler/utils.go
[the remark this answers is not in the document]
myitcv: I'll take a look, thanks for the tip.

compiler/utils.go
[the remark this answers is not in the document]
myitcv: > My suggestion doesn't contradict that. needsEscaping controls whether non-default behavior of using escaping is used. If its value remains false, the default behavior (using dot notation) is used.

No I understand what you were getting at, it's just the name `needsEscaping` that doesn't make sense to me. I'm quite happy having the inverse logic, but if we were to do so then `useBracket` would probably be a better name (it's the [alternative form](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors)).

myitcv:
> I saw the comments, but not seeing changes. This PR currently has 1 commit, so it's hard to tell what has changed (if anything) from last time.

> Unfortunately, GitHub isn't as good as Gerrit in this sense; Gerrit makes it possible to see the entire history of a change at all times.)

Sorry, yes. I keep forgetting that. Old habits die hard. Will try to change that.

myitcv:
Thanks @shurcooL - looks fine. Let's get this merged!

shurcooL:
Thanks again @myitcv!

PR #784, opened 2018-04-01
compiler: automate regeneration of prelude

myitcv:
This is a rebase of https://github.com/gopherjs/gopherjs/pull/765

Whilst rebasing https://github.com/gopherjs/gopherjs/pull/669 I fell into a similar trap to https://github.com/gopherjs/gopherjs/pull/775#discussion_r174611168, by forgetting to commit the manually minified prelude. 

Then going through the manual process of manually minifying the prelude I wasted a whole load of time installing `goexec`, missing something the first time etc, issues that this PR addresses by automating the process, as [previously discussed](https://github.com/gopherjs/gopherjs/pull/765#issuecomment-372007301)


myitcv:
@hajimehoshi any thoughts?

Gentle ping @shurcooL

compiler/prelude/prelude_min.go
```diff
@@ -0,0 +1,6 @@
...
...};var keys=$keys(v);for(var i=0;i<keys.length;i++){var k=$internalize(keys[i],t.key);m[t.key.keyFor(k)]={k:k,v:$internalize(v[keys[i]],t.elem)}}return m;case $kindPtr:if(t.elem.kind===$kindStruct){return $internalize(v,t.elem)}case $kindSlice:return new t($mapArray(v,function(e){return $internalize(e,t.elem)}));case $kindString:v=String(v);if($isASCII(v)){return v}var s="";var i=0;while(i<v.length){var h=v.charCodeAt(i);if(55296<=h&&h<=56319){var l=v.charCodeAt(i+1);var c=(h-55296)*1024+l-56320+65536;s+=$encodeRune(c);i+=2;continue}s+=$encodeRune(h);i++}return s;case $kindStruct:var noJsObject={};var searchJsObject=function(t){if(t===$jsObjectPtr){return v}if(t===$jsObjectPtr.elem){$throwRuntimeError("cannot internalize js.Object, use *js.Object instead")}switch(t.kind){case $kindPtr:return searchJsObject(t.elem);case $kindStruct:var f=t.fields[0];var o=searchJsObject(f.typ);if(o!==noJsObject){var n=new t.ptr;n[f.prop]=o;return n}return noJsObject;default:return noJsObject}};var o=searchJsObject(t);if(o!==noJsObject){return o}}$throwRuntimeError("cannot internalize "+t.string)};var $isASCII=function(s){for(var i=0;i<s.length;i++){if(s.charCodeAt(i)>=128){return false}}return true};
```
shurcooL: `Minified` was 32740 bytes before.

Now it's 41046 bytes. How come? Can we make it match previous output (from https://skalman.github.io/UglifyJS-online/ with its default options) or closer to it in size?

compiler/prelude/regen.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		log.Fatalf("uglifyjs: %v\n%s", err, out)
+		os.Exit(1)
+	}
+
+	err = ioutil.WriteFile("prelude_min.go", []byte(`// Code generated by regen. DO NOT EDIT.
+
+package prelude
+
+const Minified = `+"`"+string(out)+"`\n"), 0644)
```
shurcooL: Is `out` guaranteed not to contain <code>`</code> characters?

If so, please document the reason. Otherwise, consider using `strconv.Quote` or something equivalent that is safe.

compiler/prelude/regen.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func main() {
+	cmd := exec.Command("uglifyjs")
+	cmd.Stdin = strings.NewReader(prelude.Prelude)
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Fatalf("uglifyjs: %v\n%s", err, out)
+		os.Exit(1)
```
shurcooL: Calling `os.Exit(1)` after `log.Fatalf` is a no-op, it will never run.

Can you refactor the code to follow this pattern?

```Go
package main

import (
	"log"
)

func main() {
	err := run()
	if err != nil {
		log.Fatalln(err)
	}
}

func run() error {
	// move the code that can fail here
	if err != nil {
		return nil
	}
	// ...
	return nil
}

File: compiler/prelude/regen.go (hunk 2 above)
Comment (shurcooL) [reply compiler/prelude/regen.go]: Also, `Minified` should have a comment. It's an exported symbol.

File: compiler/prelude/prelude.go [hunk 4]
```diff
@@ -1,10 +1,8 @@
 package prelude
 
-const Prelude = prelude + numeric + types + goroutines + jsmapping
+//go:generate go run regen.go
```
shurcooL: Also, I would suggest renaming `regen.go` to `gen.go` to be more consistent with other Go projects (`gen.go` generates `prelude_min.go`, "regenerates" is not a better description). Or `genmin.go` if you want it to be more specific.

Examples:

- https://github.com/golang/go/blob/master/src/crypto/md5/gen.go
- https://github.com/golang/go/blob/master/src/sort/genzfunc.go
- https://github.com/golang/go/blob/master/src/time/genzabbrs.go

compiler/prelude/prelude_min.go
myitcv: Thanks, good catch. So it appears Actually it appears https://skalman.github.io/UglifyJS-online/ is using some non-default options. I've added `uglify_options.json` in a [new commit](https://github.com/gopherjs/gopherjs/pull/784/commits/5633de29373c2e6456ca8c74313b09785d39524f) and now things look better:

```
$ git rev-parse HEAD
5633de29373c2e6456ca8c74313b09785d39524f
$ cat compiler/prelude/prelude_min.go | wc -c
33300
```

compiler/prelude/regen.go
myitcv: > Is `out` guaranteed not to contain ``` characters?

Good catch. I don't know, but I've now made it safe in the latest commit (we need to `strings.Replace`, not quote). In any case the worst that would have happened here is that we get a compile error. But that's a moot point because we now output the safe version.

> Also, `Minified` should have a comment. It's an exported symbol.

Done. And I've also commented `Prelude` because that wasn't commented.

compiler/prelude/regen.go
myitcv: Done

compiler/prelude/prelude.go
myitcv: Done.

myitcv:
Thanks @shurcooL - pushed up a new commit addressing the feedback you gave.

myitcv:
Just to note, instead of merging this PR we could merge https://github.com/gopherjs/gopherjs/pull/791 instead. For discussion.

circle.yml
```diff
@@ -10,10 +10,13 @@ dependencies:
   post:
     - mv ./gopherjs $HOME/bin
     - npm install --global node-gyp
+    - npm install
```
hajimehoshi: I'm not familiar with npm, but why `npm install` is required here? (comments would be helpful)

compiler/prelude/prelude_min.go
```diff
@@ -0,0 +1,27 @@
+// Code generated by genmin. DO NOT EDIT.
```
hajimehoshi: IIUC, the last period is not needed:

https://github.com/golang/go/issues/13560#issuecomment-276866852

compiler/prelude/prelude.go
```diff
@@ -1,10 +1,10 @@
...
+// Prelude is the GopherJS JavaScript interop layer that is behind
+// the github.com/gopherjs/gopherjs/js package
+const Prelude = prelude + numeric + types + goroutines + jsmapping
```
hajimehoshi: I think this time is ok (I think @shurcooL has already confirmed), but I'd separate this PR into moving this part and generating this part so that we could confirm the generated code was same.

compiler/prelude/genmin.go
```diff
@@ -0,0 +1,60 @@
...
+
+	args := []string{
+		filepath.Join(bpkg.Dir, "node_modules", ".bin", "uglifyjs"),
+		"--config-file",
+		filepath.Join(preludeDir, "uglifyjs_options.json"),
+	}
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Stdin = strings.NewReader(prelude.Prelude)
+
+	out, err := cmd.CombinedOutput()
```
hajimehoshi: I think it is fragile to assume there is no stderr when err is not found. Isn't just `Output` enough?

compiler/prelude/genmin.go
```diff
@@ -0,0 +1,60 @@
...
+	fn := "prelude_min.go"
+
+	safeOut := strings.Replace(string(out), "`", "\\0060", -1)
+
+	err = ioutil.WriteFile(fn, []byte(`// Code generated by genmin. DO NOT EDIT.
+
+package prelude
+
+// Minified is an uglifyjs-minified version of Prelude
+const Minified = `+"`"+safeOut+"`\n"), 0644)
```
hajimehoshi: `strconv.Quote(string(out))` or `fmt.Sprintf("%q", string(out))` would be better instead of `safeOut` (I'm not sure `safeOut` is really enough escaping)

compiler/prelude/prelude.go
hajimehoshi: Oh, https://github.com/gopherjs/gopherjs/pull/791 is the moving part? I missed that, sorry.

compiler/prelude/genmin.go
myitcv: `safeOut` is I think enough escaping because the only thing that can't [appear between back quotes is a backquote](https://golang.org/ref/spec#String_literals). But I agree using `strconv.Quote` makes the code cleaner (although I note that the `string()` conversion of the `[]byte` variable is unnecessary when using `%q`)

compiler/prelude/prelude.go
myitcv: Yes #791 is separate. Unless we want to collapse this and #791 into one PR. Which makes sense from my perspective.

compiler/prelude/prelude_min.go
myitcv: Fixed.

myitcv:
Apologies, I managed to mess up a commit there somehow @hajimehoshi - your feedback is addressed as of 75a3424.

compiler/prelude/prelude.go
```diff
@@ -1,10 +1,10 @@
...
```
hajimehoshi: (I thought I've already mentioned but just in case)

Could we have another PR to move this part so that we could confirm that the generated min codes are same?

hajimehoshi: I'm confused: #794 moves the minimized js?

myitcv: @hajimehoshi I've updated the issue description in https://github.com/gopherjs/gopherjs/issues/793 to explain how the PRs are linked.

hajimehoshi: OK so does #794 do moving the minimized code and reformatting the original js at the same time?

myitcv: Not sure what you mean by "moving"? There are no changes in terms of target files in #794, per https://github.com/gopherjs/gopherjs/issues/793

myitcv: That move happens in #784, per the summary at the bottom of #793

hajimehoshi: Hm, #784 also changes the way of generating the minimized code, right?

myitcv: Correct

hajimehoshi: Yeah, so my point is:

> Could we have another PR to move this part so that we could confirm that the generated min codes are same?

In #784, it is hard to say the generated codes are same, right?

myitcv: We're going to get a diff introduced by switching to use `uglify-es` at some point, simply because it's a different program/environment minifying the code with nothing that guarantees the same behaviour. So I'm not entirely convinced of the benefit of having a separate PR that moves the line of minified code from `prelude.go` to `prelude_min.go`. 

To my mind the things that matter here are:

1. the tests all pass (they intentionally run with minified code)
2. the size of the minified prelude is roughly the same:

```
$ git rev-parse HEAD
e14987c0ef06db387b90fec85e8d06dc05598e24
$ goexec -quiet 'io.WriteString(os.Stdout, prelude.Minified)' | wc -c
32740
$ git checkout regen_min
$ goexec -quiet 'io.WriteString(os.Stdout, prelude.Minified)' | wc -c
32767
```

Does that sound a reasonable compromise to you?

I've just pushed up another commit that ensures we follow the exact same convention on comments for what it's worth.

compiler/prelude/genmin.go
```diff
@@ -0,0 +1,63 @@
...
+		"--config-file",
+		filepath.Join(preludeDir, "uglifyjs_options.json"),
+	}
+
+	stderr := new(bytes.Buffer)
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Stdin = strings.NewReader(prelude.Prelude)
+	cmd.Stderr = stderr
+
+	byts, err := cmd.Output()
```
shurcooL: This isn't a big deal, but it's pretty unusual to see a `[]byte` variable named `byts`. As far as I understand, it's a shortened version of "bytes"? It comes up a total of 5 times in the Go project.

In this context, just `b` or `out` would be more idiomatic and clear I think.

I see that `out` is already taken below, in that case, I would make this variable name more specific. For example, `minified`.

The reason I point this out is because I think it's a good idea to strive to write Go code that looks like Go code other Go programmers are most likely to write.

compiler/prelude/genmin.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to run %v: %v\n%s", strings.Join(args, " "), err, stderr.String())
+	}
+
+	fn := "prelude_min.go"
+
+	out := fmt.Sprintf(`// Code generated by genmin. DO NOT EDIT
```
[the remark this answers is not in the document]
myitcv: Thanks - we'd been back and forward on this already, but both @hajimehoshi had a link to the wrong version of Rob's proposal. Will fix.

compiler/prelude/genmin.go
```diff
@@ -0,0 +1,63 @@
...
+func run() error {
+	bpkg, err := build.Import("github.com/gopherjs/gopherjs", "", build.FindOnly)
+	if err != nil {
+		return fmt.Errorf("failed to locate path for github.com/gopherjs/gopherjs: %v", err)
+	}
+
+	preludeDir := filepath.Join(bpkg.Dir, "compiler", "prelude")
+
+	args := []string{
+		filepath.Join(bpkg.Dir, "node_modules", ".bin", "uglifyjs"),
```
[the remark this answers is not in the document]
myitcv: I'd tend to disagree; using `go/build` a lot, it feels very idiomatic to do exactly this sort of resolution. Wrapping exactly the same behaviour in a function just adds to the non-exported API surface and therefore cognitive overhead to my mind (i.e. I have to remember that `importPathToDir` exists and to then use it). So I'll leave as is if that's ok with you.

compiler/prelude/prelude.go
```diff
@@ -1,10 +1,10 @@
...
+// Prelude is the GopherJS JavaScript interop layer that is behind
+// the github.com/gopherjs/gopherjs/js package
```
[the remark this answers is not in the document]
myitcv: Will be fixed in my next commit.

compiler/prelude/genmin.go
```diff
@@ -0,0 +1,63 @@
...
+		return fmt.Errorf("failed to run %v: %v\n%s", strings.Join(args, " "), err, stderr.String())
+	}
+
+	fn := "prelude_min.go"
+
+	out := fmt.Sprintf(`// Code generated by genmin. DO NOT EDIT
+
+package prelude
+
+// Minified is an uglifyjs-minified version of Prelude
```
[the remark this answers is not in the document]
myitcv: Ditto.

compiler/prelude/genmin.go
myitcv: Fixed.

myitcv:
Thanks for the reviews @shurcooL @hajimehoshi - once the build passes this is ready to be merged from my perspective.

compiler/prelude/genmin.go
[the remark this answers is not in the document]
hajimehoshi: Oh I was misunderstanding that https://github.com/golang/go/issues/13560#issuecomment-276866852 was the final suggestion, without period...

compiler/prelude/prelude.go
hajimehoshi: Yeah, fair enough. Thank you!

PR #787, opened 2018-04-08
Embed core GopherJS packages into build system; enable vendoring of GopherJS.

### LINKED ISSUE #415: Vendoring gopherjs code causes generated js to fail

### LINKED ISSUE #462: "undeclared name: efaceOf" build error when using different GOPATHs
This reproduction case evolved from issue #434. Related issue #459.

Go version go1.6.2 linux/amd64
GopherJS master https://github.com/gopherjs/gopherjs/commit/ce68d341900a377d121b925cf3454b5af31fa79d

```
$ pwd
/tmp
$ mkdir go go2
$ export GOPATH=/tmp/go
$ go get github.com/gopherjs/gopherjs
$ export PATH=/tmp/go/bin:$PATH
$ cd go2
$ mkdir -p src/testpackage/
$ cat src/testpackage/main.go 
package main

func main() {

println("Hello, world!")

}

$ export GOPATH=/tmp/go2
$ go build testpackage
$ ./testpackage 
Hello, world!
$ gopherjs build testpackage
../../usr/lib/go/src/runtime/error.go:58:7: undeclared name: efaceOf
```

shurcooL:
This change brings two major features:

- It allows the `gopherjs` binary to be completely standalone and not require GopherJS source code to be in $GOPATH. Only $GOROOT is required.
- It allows the source code of `gopherjs` to be vendored in a Go project, and the vendored copy of `gopherjs` will be able to compile Go code successfully.

The approach taken to make these things possible is as follows.

As of commit ab917e062e6a7940608ee3c9d422382784f5a7ab, the natives (GopherJS-specific overrides for the Go standard library packages) are embedded in the GopherJS build system via a virtual filesystem. We re-generate the VFS whenever natives are modified.

Doing that made it possible to compile most parts of the Go standard library with GopherJS without needing the `github.com/gopherjs/gopherjs` repository in $GOPATH. However, without the `github.com/gopherjs/gopherjs/js` and `github.com/gopherjs/gopherjs/nosync` packages, compiling the Go standard library is not possible. Those two packages are core to the GopherJS compiler and are needed to build `runtime`, etc.

Therefore, since natives are already embedded, it makes a lot of sense to also embed the 2 small additional `github.com/gopherjs/gopherjs/js` and `github.com/gopherjs/gopherjs/nosync` packages analogously. Then, the VFS is used whenever the source of those two packages is accessed.

As a result, a compiled `gopherjs` binary can successfully compile Go code even if the GopherJS repository is not located in $GOPATH.

Additionally, the `natives` and `gopherjspkg` packages are very friendly towards being vendored. That means the entire GopherJS project can be vendored in another Go project, and then used to compile Go code.

Fixes #462.
Fixes #415.


hajimehoshi:
I'd be happy if you could tell me the steps to test this.

compiler/gopherjspkg/doc.go
```diff
@@ -0,0 +1,12 @@
...
+// Core GopherJS packages are packages that are critical for GopherJS compiler
+// operation. They are needed to build the Go standard library with GopherJS.
+// Currently, they include:
+//
+// 	github.com/gopherjs/gopherjs/js
+// 	github.com/gopherjs/gopherjs/nosync
+//
+package gopherjspkg
+
+//go:generate vfsgendev -source="github.com/gopherjs/gopherjs/compiler/gopherjspkg".FS -tag=gopherjsdev
```
hajimehoshi: Where does `vfgendev` come from?

shurcooL: See https://github.com/gopherjs/gopherjs/wiki/Developer-Guidelines#code-generation.

hajimehoshi:
```sh
NEWGOPATH=/tmp/gopherjs

rm -rf $NEWGOPATH
GOPATH=$NEWGOPATH go get github.com/gopherjs/gopherjs
(cd $NEWGOPATH/src/github.com/gopherjs/gopherjs; git checkout origin/embed-core-pkgs)
mkdir -p $NEWGOPATH/src/github.com/hajimehoshi/gopherjstest
echo 'package main                                                                                                                                                                                          
                                                                                                                                                                                                            
func main() {                                                                                                                                                                                               
    js.Global.Get("console").Call("log", "Hello")                                                                                                                                                           
}' > $NEWGOPATH/src/github.com/hajimehoshi/gopherjstest/main.go
mkdir -p $NEWGOPATH/src/github.com/hajimehoshi/gopherjstest/vendor/github.com/gopherjs

# vendering                                                                                                                                                                                                 
cp -r $NEWGOPATH/src/github.com/gopherjs/gopherjs $NEWGOPATH/src/github.com/hajimehoshi/gopherjstest/vendor/github.com/gopherjs
rm -rf $NEWGOPATH/src/github.com/gopherjs/gopherjs

GOPATH=$NEWGOPATH go build github.com/hajimehoshi/gopherjstest
GOPATH=$NEWGOPATH $NEWGOPATH/bin/gopherjs run $NEWGOPATH/src/github.com/hajimehoshi/gopherjstest/main.go
```

failed with the message

```
cannot find package "github.com/gopherjs/gopherjs/js" in any of:
        /usr/local/go/src/github.com/gopherjs/gopherjs/js (from $GOROOT)
        /tmp/gopherjs/src/github.com/gopherjs/gopherjs/js (from $GOPATH)
```

Am I missing something? (Without vendering, it succeeded to say 'Hello'.)

myitcv:
> The way I tested was:

Should we be getting tests like the ones you describe in a script (or similar) as part of the repo? Automates the process of verifying that the behaviour is as described, makes it easier for you to test that the vendored version works with any subsequent commits/changes submitted to this PR, but also makes it easier for anyone else making future (unrelated) changes to verify that their changes haven't broken the vendor behaviour.

shurcooL:
@hajimehoshi I followed your edited instructions and got a different outcome. Notably, `gopherjs` didn't build because you forgot to vendor one of its new dependencies. After I fetched that dependency, it built successfully, and the `gopherjstest/main.go` program was executed by GopherJS successfully.

I think you made a mistake somewhere, and ended up using the `master` branch of GopherJS instead of this PR. I suggest you try those instructions a second time.

See the log from my terminal for reference:

<details><br>

```bash
Last login: Sun Apr  8 13:11:18 on ttys000
~ $ cd /tmp/
tmp $ ls
com.apple.launchd.0q00aXw85q	powerlog
com.apple.launchd.v3Mhw42Gmb
tmp $ NEWGOPATH=/tmp/go2
tmp $ ls $NEWGOPATH
ls: /tmp/go2: No such file or directory
tmp $ rm -rf $NEWGOPATH
tmp $ GOPATH=$NEWGOPATH go get github.com/gopherjs/gopherjs
tmp $ tree $NEWGOPATH | head
/tmp/go2
├── bin
│   └── gopherjs
└── src
    ├── github.com
    │   ├── fsnotify
    │   │   └── fsnotify
    │   │       ├── AUTHORS
    │   │       ├── CHANGELOG.md
    │   │       ├── CONTRIBUTING.md
tmp $ (cd $NEWGOPATH/src/github.com/gopherjs/gopherjs; git checkout origin/embed-core-pkgs)
Note: checking out 'origin/embed-core-pkgs'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 9b293ec... README: Remove note saying vendoring GopherJS is unsupported.
tmp $ mkdir -p $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest
tmp $ echo 'package main                                                                                                                                                                                          
>                                                                                                                                                                                                             
> import "github.com/gopherjs/gopherjs/js"                                                                                                                                                                    
>                                                                                                                                                                                                             
> func main() {                                                                                                                                                                                               
>     js.Global.Get("console").Call("log", "Hello")                                                                                                                                                           
> }' > $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/main.go
tmp $ tree /tmp/go2/src/foobarbaz/
/tmp/go2/src/foobarbaz/
└── hajimehoshi
    └── gopherjstest
        └── main.go

2 directories, 1 file
tmp $ mkdir -p $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs
tmp $ ls go2/src/
foobarbaz	github.com	golang.org
tmp $ cp -r $NEWGOPATH/src/github.com $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/vendor
tmp $ cp -r $NEWGOPATH/src/golang.org $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/vendor
tmp $ cd go2
go2 $ ls
bin	src
go2 $ ls src/
foobarbaz/  github.com/ golang.org/ 
go2 $ ls src/foobarbaz/hajimehoshi/gopherjstest/vendor/
github.com	golang.org
go2 $ rm -rf $NEWGOPATH/src/github.com/gopherjs/gopherjs
go2 $ ls -la bin/gopherjs 
-rwxr-xr-x  1 Dmitri  staff  12955388 Apr  8 13:24 bin/gopherjs
go2 $ GOPATH=$NEWGOPATH go install -v foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs
src/foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/build/build.go:27:2: cannot find package "github.com/shurcooL/httpfs/vfsutil" in any of:
	/tmp/go2/src/foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/shurcooL/httpfs/vfsutil (vendor tree)
	/usr/local/go/src/github.com/shurcooL/httpfs/vfsutil (from $GOROOT)
	/tmp/go2/src/github.com/shurcooL/httpfs/vfsutil (from $GOPATH)
go2 $ (mkdir /tmp/go2/src/foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/shurcooL; cd /tmp/go2/src/foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/shurcooL; git clone https://github.com/shurcooL/httpfs)
Cloning into 'httpfs'...
remote: Counting objects: 197, done.
remote: Total 197 (delta 0), reused 0 (delta 0), pack-reused 197
Receiving objects: 100% (197/197), 32.35 KiB | 1.08 MiB/s, done.
Resolving deltas: 100% (82/82), done.
go2 $ GOPATH=$NEWGOPATH go install -v foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler/prelude
foobarbaz/hajimehoshi/gopherjstest/vendor/golang.org/x/sys/unix
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/neelance/sourcemap
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/kisielk/gotool/internal/load
foobarbaz/hajimehoshi/gopherjstest/vendor/golang.org/x/tools/go/buildutil
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler/astutil
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler/typesutil
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/spf13/pflag
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler/vendor/github.com/neelance/astrewrite
foobarbaz/hajimehoshi/gopherjstest/vendor/golang.org/x/tools/go/gcimporter15
foobarbaz/hajimehoshi/gopherjstest/vendor/golang.org/x/tools/go/types/typeutil
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler/filter
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler/analysis
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler/gopherjspkg
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler/natives
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/shurcooL/httpfs/vfsutil
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/kisielk/gotool
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/compiler
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/internal/sysutil
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/fsnotify/fsnotify
foobarbaz/hajimehoshi/gopherjstest/vendor/golang.org/x/crypto/ssh/terminal
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/spf13/cobra
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs/build
foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs/gopherjs
go2 $ GOPATH=$NEWGOPATH go build foobarbaz/hajimehoshi/gopherjstest
go2 $ GOPATH=$NEWGOPATH $NEWGOPATH/bin/gopherjs run $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/main.go
Hello
go2 $ 
```

</details>

hajimehoshi:
@shurcooL OK, I've succeeded to say 'hello' with your fix, thank you!

```sh
NEWGOPATH=/tmp/gopherjs

rm -rf $NEWGOPATH
GOPATH=$NEWGOPATH go get -d github.com/gopherjs/gopherjs
(cd $NEWGOPATH/src/github.com/gopherjs/gopherjs; git checkout origin/embed-core-pkgs)
GOPATH=$NEWGOPATH go get github.com/shurcooL/httpfs/vfsutil
GOPATH=$NEWGOPATH go install github.com/gopherjs/gopherjs

# Create the main package                                                                                                                                                                                   
mkdir -p $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest
echo 'package main                                                                                                                                                                                          
                                                                                                                                                                                                            
func main() {                                                                                                                                                                                               
    js.Global.Get("console").Call("log", "Hello")                                                                                                                                                           
}' > $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/main.go
mkdir -p $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/vendor/github.com/gopherjs

# Vendering                                                                                                                                                                                                 
cp -r $NEWGOPATH/src/github.com $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/vendor
cp -r $NEWGOPATH/src/golang.org $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/vendor
rm -rf $NEWGOPATH/src/github.com/gopherjs/gopherjs                                                                                                      

GOPATH=$NEWGOPATH go build foobarbaz/hajimehoshi/gopherjstest
GOPATH=$NEWGOPATH $NEWGOPATH/bin/gopherjs run $NEWGOPATH/src/foobarbaz/hajimehoshi/gopherjstest/main.go
```

build/build.go
```diff
@@ -47,9 +56,52 @@ func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
 		),
 		ReleaseTags: build.Default.ReleaseTags,
 		CgoEnabled:  true, // detect `import "C"` to throw proper error
+
+		IsDir: func(path string) bool {
+			if strings.HasPrefix(path, gopherJSRoot+string(filepath.Separator)) {
+				path = filepath.ToSlash(path[len(gopherJSRoot):])
+				if fi, err := vfsutil.Stat(gopherjspkg.FS, path); err == nil {
```
hajimehoshi: This condition is repeated four times, so how about creating a function?

build/build.go
```diff
@@ -91,17 +145,17 @@ func importWithSrcDir(path string, srcDir string, mode build.ImportMode, install
 	case "crypto/x509", "os/user":
 		// These stdlib packages have cgo and non-cgo versions (via build tags); we want the latter.
 		bctx.CgoEnabled = false
+	case "github.com/gopherjs/gopherjs/js", "github.com/gopherjs/gopherjs/nosync":
```
hajimehoshi: Isn't it ok to check in the above way like `strings.HasPrefix(path, gopherJSRoot+string(filepath.Separator)` for consistency?

compiler/gopherjspkg/fs.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/shurcooL/httpfs/filter"
+)
+
+// FS is a virtual filesystem that contains core GopherJS packages.
+var FS = filter.Keep(
+	http.Dir(importPathToDir("github.com/gopherjs/gopherjs")),
+	func(path string, fi os.FileInfo) bool {
+		return path == "/" ||
+			path == "/js" || strings.HasPrefix(path, "/js/") ||
+			path == "/nosync" || strings.HasPrefix(path, "/nosync/")
```
hajimehoshi: Not only `github.com/gopherjs/js`, but also `github.com/gopherjs/js/foobar` would be a special path?

hajimehoshi:
And as @myitcv suggests, we should have a test for this.

mpl:
@shurcooL I haven't looked at the code at all, but with this PR it seems I was able to build Perkeep without our temp GOPATH hackery, so LGTM as far as I'm concerned. thanks!

build/build.go
hajimehoshi: Sorry but I meant `strings.HasPrefix(path, gopherJSRoot+string(filepath.Separator)` or `path = filepath.ToSlash(path[len(gopherJSRoot):])`. Aren't they repeated?

build/build.go
hajimehoshi: OK, thanks. I feel like this set is duplicated with `FS` definition in `fs.go`. Would it be possible to unify these sets into one?

shurcooL:
Improved failing test output in commit 8a9d21b0b1d346773ce412b48b6da808a9ed823e:

```
$ go test -v -run=TestGopherJSCanBeVendored
=== RUN   TestGopherJSCanBeVendored
vendoring github.com/gopherjs/gopherjs/js package is not supported, see https://github.com/gopherjs/gopherjs/issues/415
--- FAIL: TestGopherJSCanBeVendored (5.19s)
	gorepo_test.go:40: exit status 1
FAIL
exit status 1
FAIL	github.com/gopherjs/gopherjs/tests	5.190s
```

compiler/gopherjspkg/fs.go
shurcooL: I made the filter more precise in 2ad4a5b, so it won't pick up any subdirectories now.

build/build.go
shurcooL: The "core" GopherJS packages are enumerated and documented in the package comment of `./compiler/gopherjspkg` package (see `compiler/gopherjspkg/doc.go` file).

Here, the `./build` package implements that. It uses import paths.

In `gopherjspkg.FS`, the code picks up those packages via relative directory paths.

I don't think anything more needs to be done, this is very simple, and trying to make it more DRY isn't worth the effort (I'm usually all for it, but it needs to be worthwhile).

build/build.go
shurcooL: I see, thanks for clarifying.

I've considered it, but I think keeping it as is is more readable and simpler. The duplication is locally contained in the same place. Factoring it out into a helper moves the relevant code further away from where it's being used, and there's just too little code here to factor out efficiently.

shurcooL:
I filed issue #792 for some potential followup work.

tests/gorepo_test.go
```diff
@@ -22,3 +25,21 @@ func TestGoRepositoryCompilerTests(t *testing.T) {
...
+	}
+
+	cmd := exec.Command("bash", "gopherjsvendored_test.sh")
+	cmd.Stderr = os.Stdout
+	got, err := cmd.Output()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if want := "hello using js pkg\n"; string(got) != want {
+		t.Errorf("got != want:\ngot:\n%s\nwant:\n%s", got, want)
```
hajimehoshi: I'd write the message mentioning the script like `gopherjsvendored_test.js failed`

tests/gorepo_test.go
```diff
@@ -22,3 +25,21 @@ func TestGoRepositoryCompilerTests(t *testing.T) {
...
+
+// Test that GopherJS can be vendored into a project, and then used to build Go programs.
+// See issue https://github.com/gopherjs/gopherjs/issues/415.
+func TestGopherJSCanBeVendored(t *testing.T) {
+	if runtime.GOARCH == "js" {
+		t.Skip("test meant to be run using normal Go compiler (needs os/exec)")
+	}
+
+	cmd := exec.Command("bash", "gopherjsvendored_test.sh")
+	cmd.Stderr = os.Stdout
```
hajimehoshi: Why not catching the stderr and dumping it as an error message?

tests/gopherjsvendored_test.sh
```diff
@@ -0,0 +1,51 @@
...
+cp -r $(go list -e -f '{{.Dir}}' github.com/fsnotify/fsnotify)  "$tmp/src/example.org/hello/vendor/github.com/fsnotify/fsnotify"
+cp -r $(go list -e -f '{{.Dir}}' github.com/kisielk/gotool)     "$tmp/src/example.org/hello/vendor/github.com/kisielk/gotool"
+cp -r $(go list -e -f '{{.Dir}}' github.com/neelance/sourcemap) "$tmp/src/example.org/hello/vendor/github.com/neelance/sourcemap"
+cp -r $(go list -e -f '{{.Dir}}' github.com/shurcooL/httpfs)    "$tmp/src/example.org/hello/vendor/github.com/shurcooL/httpfs"
+cp -r $(go list -e -f '{{.Dir}}' github.com/spf13/cobra)        "$tmp/src/example.org/hello/vendor/github.com/spf13/cobra"
+cp -r $(go list -e -f '{{.Dir}}' github.com/spf13/pflag)        "$tmp/src/example.org/hello/vendor/github.com/spf13/pflag"
+cp -r $(go list -e -f '{{.Dir}}' golang.org/x/crypto)           "$tmp/src/example.org/hello/vendor/golang.org/x/crypto"
+cp -r $(go list -e -f '{{.Dir}}' golang.org/x/sys)              "$tmp/src/example.org/hello/vendor/golang.org/x/sys"
+cp -r $(go list -e -f '{{.Dir}}' golang.org/x/tools)            "$tmp/src/example.org/hello/vendor/golang.org/x/tools"
```
hajimehoshi: If GopherJS's dependencies are changed, we should update these items, right? At least, should we leave a comment?

shurcooL: Yeah. Perhaps this script can be improved in the future to infer dependencies more automatically (e.g., with `go list -f '{{join .Deps "\n"}}' github.com/gopherjs/gopherjs/...` or so).

There's a comment on top saying `# Vendor GopherJS and its dependencies into hello project.`, is it not clear?

tests/gorepo_test.go
shurcooL: The test is considered failed if the exit code is non-zero, or if the stdout output is not exactly the expected "hello using js pkg\n" value.

Stderr can be anything, it shouldn't fail the test. It's just information.

For example, if you run the test without the sourcemap module, it'll print "gopherjs: Source maps disabled. Install source-map-support module for ..." message to stderr.

tests/gopherjsvendored_test.sh
myitcv: Feels like we should make this change now; it's brittle having these hard-coded, particularly when using `go list` is something that we're already doing elsewhere and is easy to make work in this situation.

hajimehoshi: Agreed with @myitcv . This doesn't look so hard.

tests/gorepo_test.go
hajimehoshi: Hm, as I am not sure GopherJS's test policy, I don't have a strong opinion on it. Then, that's fine :-)

tests/gorepo_test.go
hajimehoshi: Makes sense.

tests/gorepo_test.go
```diff
@@ -22,3 +25,21 @@ func TestGoRepositoryCompilerTests(t *testing.T) {
...
 }
+
+// Test that GopherJS can be vendored into a project, and then used to build Go programs.
+// See issue https://github.com/gopherjs/gopherjs/issues/415.
+func TestGopherJSCanBeVendored(t *testing.T) {
+	if runtime.GOARCH == "js" {
+		t.Skip("test meant to be run using normal Go compiler (needs os/exec)")
+	}
+
+	cmd := exec.Command("bash", "gopherjsvendored_test.sh")
```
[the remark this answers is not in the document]
shurcooL: Did you see the commit message of 49f2b5ca9ae7415de29eb86c95d2cecaca9d606f?

However, I think I know what it'll take to make it work with `sh`. It might just be a matter of removing the `SIG` prefix from the signal names.

compiler/version_check.go
```diff
@@ -6,4 +6,4 @@ package compiler
 const ___GOPHERJS_REQUIRES_GO_VERSION_1_10___ = true
 
 // Version is the GopherJS compiler version string.
-const Version = "1.10-2"
+const Version = "1.10-3"
```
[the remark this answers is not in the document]
shurcooL: I don't see a difference, but sure, I can take out this commit and apply it on `master` after this PR lands.

compiler/gopherjspkg/doc.go
shurcooL: That's out of scope of this PR. This `gopherjspkg` package is the same in structure as `natives`.

tests/gorepo_test.go
shurcooL: I can make it something like this, hope it's more clear:

```
unexpected stdout from gopherjsvendored_test.sh:
	got:
	unexpected output 123
	
	want:
	hello using js pkg
```

compiler/version_check.go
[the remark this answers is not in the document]
myitcv: The difference being that if this PR needs to be reverted at a later date we don't want to revert the version change. Keeping the release as a separate PR ensures we have that ability.

tests/gorepo_test.go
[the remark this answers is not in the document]
myitcv: Is it not cleaner to just write this test as `// +build ignore`-d `.go` file in any case? That way we don't need to worry about shells, traps etc.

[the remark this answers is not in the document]
shurcooL: I often prefer Go over other tools, but a bash script seemed like a good fit for this task. It was based on the bash script in [playground repo](https://github.com/gopherjs/gopherjs.github.io/blob/b3fd139784228a7f0371aa48599b6377bac7fbaf/playground/update.sh), which has been used with success for some years.

It's already done and it works well. I don't see a need to spend much more time on it.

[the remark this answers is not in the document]
myitcv: Will leave it up to you. It might be, however, that addressing https://github.com/gopherjs/gopherjs/pull/787/files/2ad4a5b9a30e5f1f31c643c2a2ac27543a5ddb58#r181624969 would be easier/clearer in a `.go` file.

tests/gopherjsvendored_test.sh
shurcooL: Done.

shurcooL:
I've rebased this PR on top of latest `master` and addressed all outstanding comments. PTAL.

(Previous PR version was 2ad4a5b9a30e5f1f31c643c2a2ac27543a5ddb58).

tests/gopherjsvendored_test.sh
```diff
@@ -0,0 +1,48 @@
...
+    pkgDir=$(go list -f '{{.Dir}}' "$1")
+    # Copy all files (not directories), other than ones that start with "." or "_".
+    for f in $(find -H "$pkgDir" -maxdepth 1 -name "[^._]*" -type f); do
+        cp "$f" "$2"
+    done
+}
+
+# Make a hello project that will vendor GopherJS.
+mkdir -p "$tmp/src/example.org/hello"
+echo 'package main
```
myitcv: Just a minor thought, but the heredoc syntax often reads a bit cleaner in these multi-line situations because the redirect is on the same line:

```bash
cat <<EOD > "$tmp/src/example.org/hello/main.go"
package main

func main() {
    js.Global.Get("console").Call("log", "hello using js pkg")
}
EOD 
```

tests/gorepo_test.go
```diff
@@ -1,15 +1,18 @@
...
 import (
 	"os"
 	"os/exec"
+	"runtime"
 	"testing"
 )
 
 // Go repository basic compiler tests, and regression tests for fixed compiler bugs.
 func TestGoRepositoryCompilerTests(t *testing.T) {
+	if runtime.GOARCH == "js" {
```
myitcv: What's the thinking behind dropping the `// +build !js` build tag here?

tests/lowlevel_test.go
```diff
@@ -15,6 +14,10 @@ import (
 //
 // See https://github.com/gopherjs/gopherjs/issues/279.
 func TestTimeInternalizationExternalization(t *testing.T) {
+	if runtime.GOARCH == "js" {
```
myitcv: Same question on why the build tag isn't sufficient in this situation and why you prefer the `t.Skip`

tests/gopherjsvendored_test.sh
shurcooL: Thanks for the suggestion. I wasn't familiar with this syntax, but I'll consider it next time.

tests/gorepo_test.go
shurcooL: The motivation is written in the commit message of d0d69c08b6af458cebb60d3295f643e4a471029d:

> Rely on runtime.GOARCH and t.Skip to skip tests that shouldn't be run
> with GopherJS. It's more clear and reliable than a file-wide build tag.

`t.Skip` is exists per-test (rather than for the entire file) and can offer an individual message for the skip reason. This was better in this context. Build tags were used previously because I didn't get the idea that `t.Skip` could be used instead until now.

shurcooL:
Thanks for the reviews!

I'm going to merge this and bump up the GopherJS compiler version to 1.10-3, the first version to support vendoring GopherJS into other Go projects.

myitcv:
@shurcooL - would it make sense to also review and merge https://github.com/gopherjs/gopherjs/pull/669 prior to the new version?

shurcooL:
There isn't a need, that PR is orthogonal. We can bump up to 1.10-4 for it if needed.

myitcv:
> There isn't a need, that PR is orthogonal. We can bump up to 1.10-4 for it if needed.

Yes I know it's orthogonal, but I was proposing bundling together a number of fixes into one release. Or would you prefer to have one point release per fix/feature?

shurcooL:
Let's discuss that outside this PR. I've already effectively allocated 1.10-3 for this fix by bumping to 1.10-3 in the earlier version of this PR, so that release has already been decided.

We can reconsider the approach for future versions.

shurcooL:
As a followup step, I've checked that the GopherJS Playground works with the latest version (it does) and updated it in https://github.com/gopherjs/gopherjs.github.io/commit/cbdec896dfc8a335278926fc7d6a7e2c43ccdb05.

PR #791, opened 2018-04-15
compiler/prelude: move prelude to separate .js files

myitcv:
I've been working with the prelude quite a lot recently. One of the really painful things with the current implementation is that the JavaScript for the prelude exists as a `const` string in a `.go` file, which makes properly writing and formatting the Javascript itself very difficult.

This PR splits the prelude into separate `.js` files therefore and adds a `formatpreludejs.go` `go generate`-er to format those `.js` files in place.

PR #794, opened 2018-04-18
compiler/prelude: format .js files using prettier

myitcv:
**DO NOT MERGE**

_Based on #791 - this will need a rebase once #791 is merged._

See the diff this PR introduces on top of #791 [here](https://github.com/myitcv/gopherjs-compiler/compare/js_file_prelude...myitcv:prettier_js)

This PR introduces automatic formatting of the prelude's now separate
.js files. Using prettier has the same argument as using gofmt on .go
source code.

As part of this we make the go generate step for the prelude include a
formatpreludejs.go step.

The resulting prelude_min.go file is slightly different now, but post
#791 we can see these
differences are entirely and solely attributable to prettier's adjusting
of the .js files.


flimzy:
See https://github.com/gopherjs/gopherjs/pull/791#issuecomment-1002102777 for the future plans on this issue.

PR #797, opened 2018-04-18
ci: upgrade to CircleCI 2

### LINKED ISSUE #766: Deal with CircleCI 1.0 End of Life on August 31, 2018.
See https://circleci.com/blog/sunsetting-1-0/:

> TL;DR: After August 31, 2018, CircleCI 1.0 will no longer be available for Linux and macOS users.

We're currently using CircleCI 1.0 for this project (/cc @neelance):

image

Need to deal with it by the time it's shut down. Options include:

- migrate to CircleCI 2.0 (if it's still free)
- migrate to another CI

myitcv:
Based on the instructions https://circleci.com/docs/2.0/migrating-from-1-2/, this is as straightforward an upgrade as possible, i.e. with as few variations as possible.

We also move to an exclusion list approach for the standard library tests we want to ignore. This is an entirely derived list based on the current exclusions with the exception of `os/user` which is added in order to make the tests pass. Further work can be done on this exclusion list in future PRs (i.e. removing exclusions that should not be there).

Fixes #766.


myitcv:
@shurcooL @hajimehoshi this is ready for review.

myitcv:
Closing because this was superseded by https://github.com/gopherjs/gopherjs/pull/864

PR #798, opened 2018-04-20
WIP: Ability to call overridden functions from the std library

r-l-x:
Gopherjs overrides some functions from the go standard library. The original implementation from the standard library gets removed from the compilation process.

This PR implements the ability to retain designated overridden functions from the go standard library.
#558 is a typical use case for this (try an optimized way to get the result, but if this doesn't work, fallback to the standard library).

Here is how to use it:
If a function in `compiler/natives` overrides a function from the standard library and has the `gopherjs:keep_overridden` directive in the function comment, then the overridden function from the standard library will be compile, and can be called under the name `_gopherjs_overridden_<function_name>`.

A typical use would look like (in a file in `compiler/natives`):
```
//gopherjs:keep_overridden
// This function overrides foo() from the standard library
func foo() {
  // try the optimized function
  if err := optimized_foo(); err != nil {
    // this didn't work, fallback to the standard library implementation
    _gopherjs_overridden_foo()
  }
}

```


myitcv:
Ah, presumably https://github.com/r-l-x/gopherjs/blob/bcfda31b2bc688a9dce31c1402629ad44d949d14/compiler/natives/src/crypto/sha256/sha256.go#L28-L38 is such an example?

r-l-x:
@myitcv Yes exactly.

myitcv:
My suggestion would be we use this directive for at least one function as part of this PR to demonstrate it actually works, passes tests, minifies properly etc. 

Then we can add additional functions etc as proposed in #558 and anything else in future PRs.

myitcv:
Sorry for the noise in my last few comments/reviews...

r-l-x:
Well, actually, #558 is the prefect exemple for this (and it's based on this branch). @shurcooL suggested issuing a separate PR for this, so that's what I did, but if you prefer considering #558 directly, I don't mind.

It's not easy to add a "minimalist" use for this feature without risking breaking something or adding useless code.

myitcv:
> It's not easy to add a "minimalist" use for this feature without risking breaking something or adding useless code.

I think this is the key. Either there needs to be a test or some minimal use of what's being committed in this PR, some way we can "know" it's working.

r-l-x:
Ok, I just [used this](https://github.com/gopherjs/gopherjs/pull/798/commits/5cd0971517beac70dd107ffed0bcec2200e737d6#diff-86969dd7844e85288656189eedd46882R10) in the gopherjs regexp test "TestOnePassCutoff".

This test was systematically skipped, because apparently it may raise "Maximum call stack size exceeded" errors. I rewrote the test using `gopherjs:keep_overridden`, so that the original test is run and only in case of a panic, the test gets skipped. This allows the person running the test if the test was actually in success or not, without making the gopherjs tests fail in case of error. It wouldn't be possible to do this without this PR (appart from duplicating the code from the standard library).

And actually both on my configuration and on [circleci](https://circleci.com/gh/gopherjs/gopherjs/1545) the test runs well (so it seems like it could be re-enabled):
```
=== RUN   TestOnePassCutoff
--- PASS: TestOnePassCutoff (0.08s)
```

myitcv:
> Ok, I just used this in the gopherjs regexp test "TestOnePassCutoff".

Perfect.

> And actually both on my configuration and on circleci the test runs well (so it seems like it could be re-enabled):

Thanks for flagging. I suspect this got fixed by https://github.com/gopherjs/gopherjs/pull/687. I've raised https://github.com/gopherjs/gopherjs/issues/801 to follow up on re-enabling this and other tests.

build/build.go
```diff
@@ -187,6 +187,23 @@ func ImportDir(dir string, mode build.ImportMode, installSuffix string, buildTag
 	return &PackageData{Package: pkg, JSFiles: jsFiles}, nil
 }
 
+// Test if we find the '//gopherjs:keep_overridden' comment
+func findKeepOverridenComment(doc *ast.CommentGroup) bool {
+	if doc == nil {
+		return false
+	}
+	for _, comment := range doc.List {
+		text := comment.Text
```
myitcv: Might I suggest the following is slightly clearer in its intention?

```go
text := strings.TrimRightFunc(comment.Text, unicode.IsSpace)
if text == "//gopherjs:keep_overridden" {
  return true
}
```

i.e. you're looking for a line that contains only the special comment, save any trailing space.

r-l-x: I actually took this part from the go pragma directive parser:
https://github.com/golang/go/blob/master/src/cmd/compile/internal/gc/noder.go#L1435
It seems like you can add text after a go pragma directive, and it will be ignored, and I think it's a good idea to do the same here.

r-l-x:
I just rebased to fix conflicts and fixed a typo in the code.

build/build.go
myitcv: Got it. I think you'll find that is effectively the parse phase of the compiler; so it has to deal with all types of pragma. [Many of which take arguments](https://gist.github.com/myitcv/e4202a9513bed098262444ed0f4f3625).

In this case we've already parsed into the AST and we know we have a comment group.

And specifically, I'm guessing you want `//gopherjs:keep_overridden` to be used without arguments, because there are none. Hence it doesn't make sense for anything to follow that special comment.

r-l-x: Considering the [pragma function from the standard library]( https://github.com/golang/go/blob/20c98dc396d1cb8c3a9776f1d330feed2f2a132a/src/cmd/compile/internal/gc/noder.go#L1397), my understanding is that pragmas expecing arguments (`go:linkname`, `go:cgo_*`) are treated in the non-default cases of the switch. The default case, from which this code is taken, only treats pragmas with no arguments (`go:nointerface`, `go:noescape`, etc., see the implementation of [pragmaValues](https://github.com/golang/go/blob/20c98dc396d1cb8c3a9776f1d330feed2f2a132a/src/cmd/compile/internal/gc/lex.go#L52))
So it looks like the standard library allows putting any text (so basically a comment) after a pragma without argument. I think it's appropriate to have the same behaviour here (even if I agree that it's not the typical use case).

myitcv: > So it looks like the standard library allows putting any text (so basically a comment) after a pragma without argument. I think it's appropriate to have the same behaviour here (even if I agree that it's not the typical use case).

You're absolutely right, thanks for correcting me.

I'm easy either way; in this PR I think you can be very specific and prescriptive because you're defining the pragma so it's in your gift to do what you want 😄 - to my mind doing so keeps the documentation (on how to use your pragma) simple as well as the implementation. 

But, as I said earlier, it's a minor point in any case.

myitcv:
This LGTM 👍

nevkontakte:
@r-l-x yes, renaming to `gopherjs:keep-original` sounds good to me. If you don't think you'll have to do the rebase in the near future, please let me know, I can try to do it myself.

PR #800, opened 2018-04-20
Move js package tests into tests package.

### LINKED ISSUE #792: Move tests in js package to tests package.
After PR #787 lands, the [`js`](https://godoc.org/github.com/gopherjs/gopherjs/js) and [`nosync`](https://godoc.org/github.com/gopherjs/gopherjs/nosync) packages will be considered "core" and embedded into the GopherJS compiler via `compiler/gopherjspkg` package. `gopherjspkg` needs to be regenerated whenever `js` or `nosync` changes (not often).

To make everything work, I embedded the entire `js`, `nosync` packages, including their tests, and did https://github.com/gopherjs/gopherjs/pull/787/commits/c964badbd94f063b99e21a368c78324bbc529b00 to make their tests run.

When reviewing #779, I've realized we might want to modify/add tests in `js` package more often. Having to regenerate `gopherjspkg` is inconvenient (e.g., especially so when in the process of writing/modifying a test). It also increases the size of the compiler, and I don't want to feel pressure to keep the tests smaller because of that.

Finally, there's always been an odd split of tests in `js` package and tests in `tests` package. Sometimes it's not clear where to add a test.

By moving `js` tests from `js` package into `tests`, all these problems are resolved.

All of `js` tests are external, so I think its should be easy. I think it's better than trying to make `js` normal package embedded but its tests not embedded.

I think a good way of thinking `github.com/gopherjs/gopherjs/js` package (after #787) is that it's somewhat like `C` pseudo-package (related to `cgo`), and it wouldn't make sense to try to put tests inside `C`.

shurcooL:
This way, all GopherJS-specific tests are in tests package, rather than
having an arbitrary split. None of the package js tests were internal
to the package.

This helps a future change where the js package becomes embedded into
the compiler. It makes it easier to add and modify GopherJS tests
without having to regenerate the js package.

Resolves #792.

### DIFF
--- a/tests/js_test.go
+++ b/tests/js_test.go
@@ -1,6 +1,6 @@
 // +build js
 
-package js_test
+package tests_test
 
 import (
 	"fmt"


myitcv:
As discussed @shurcooL 👍 from me.

shurcooL:
I'm going to merge this and rebase #787 on top. That way, none of the GopherJS tests will be embedded into the core packages, which has benefits (smaller compiler size, easier to modify tests).

PR #789, opened 2018-04-10
compiler: vendor golang.org/x/tools/go/gcimporter15

myitcv:
As of 5dfd8930f264888dd5fe26820a0ac605a384dd33
golang.org/x/tools/go/gcimporter15 has been removed (in line with a
promise to remove this old API). This CL vendors a copy of
golang.org/x/tools/go/gcimporter15 as of
14d5b80f954f510ba9bcd241821b6f4828498e1a
(5dfd8930f264888dd5fe26820a0ac605a384dd33~1) minus *_test.go files and
the testdata directory (which contained binary files).

This is effectively a sticking plaster until we migrate, per the
deprecation notice, to golang.org/x/tools/go/gcexportdata.

Fixes #788

PR #803, opened 2018-04-21
[WIP] compiler: remove dependency on legacy gcimporter15

myitcv:
**DO NOT MERGE**

**DO NOT REVIEW**

This is a follow up to #789 that removes the sticking plaster in that PR,
and switches the compiler to using golang.org/x/tools/go/gcexportdata as
directed here
https://github.com/golang/tools/blob/14d5b80f954f510ba9bcd241821b6f4828498e1a/go/gcimporter15/gcimporter.go#L13-L14.

This means we can also remove the temporary vendor of gcimporter15.


myitcv:
@shurcooL @hajimehoshi this is now ready for review.

compiler/compiler.go
```diff
@@ -243,7 +243,8 @@ func ReadArchive(filename, path string, r io.Reader, packages map[string]*types.
 	}
 
 	var err error
-	_, packages[path], err = gcimporter.BImportData(token.NewFileSet(), packages, a.ExportData, path)
+	b := bytes.NewReader(a.ExportData)
+	packages[path], err = gcexportdata.Read(b, token.NewFileSet(), packages, path)
```
shurcooL: `b` is not a descriptive variable name. That's often okay in short functions, but it doesn't work well in this context.  It's also typically associated with `[]byte` type, not `io.Reader`. I see `r` is already taken and I don't think it's a good idea to reuse it.

I think it's better to either use a longer descriptive name here:

```Go
exportData := bytes.NewReader(a.ExportData)
packages[path], err = gcexportdata.Read(exportData, token.NewFileSet(), packages, path)
```

Which would mirror the `gcexportdata.Write ` code below.

But I think this line is not that long, that it's even better to just inline it:

```Go
packages[path], err = gcexportdata.Read(bytes.NewReader(a.ExportData), token.NewFileSet(), packages, path)
```

Either of those is okay.

compiler/package.go
```diff
@@ -164,7 +164,10 @@ func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, impor
 	}
 	importContext.Packages[importPath] = typesPkg
 
-	exportData := gcimporter.BExportData(nil, typesPkg)
+	exportData := new(bytes.Buffer)
+	if err := gcexportdata.Write(exportData, nil, typesPkg); err != nil {
+		return nil, fmt.Errorf("failed to write export data: %v", err)
+	}
 	encodedFileSet := bytes.NewBuffer(nil)
```
shurcooL: This line uses a very outdated style of initializing a new `bytes.Buffer`, no one writes it like that in anymore. [`bytes.NewBuffer`](https://godoc.org/bytes#NewBuffer) documentation hints at this:

> In most cases, new(Buffer) (or just declaring a Buffer variable) is sufficient to initialize a Buffer.

It's not surprising because this line was written 4 years ago.

I know it's not your code, but it's the line directly following your code, and I think it's not nice how just 4 lines above it is a different style is used to do the same thing:

```Go
exportData := new(bytes.Buffer)
// ... 3 lines
encodedFileSet := bytes.NewBuffer(nil)
```

Let's make this better by updating it so they both use the same style.

I would slightly prefer the `var buf bytes.Buffer` (and using `&` operator) style over `buf := new(bytes.Buffer)`:

```Go
var exportData bytes.Buffer
if err := gcexportdata.Write(&exportData, nil, typesPkg); err != nil {
	return nil, fmt.Errorf("failed to write export data: %v", err)
}
var encodedFileSet bytes.Buffer
if err := fileSet.Write(json.NewEncoder(&encodedFileSet).Encode); err != nil {
```

But if you feel strongly about using `new(bytes.Buffer)` for both variables (and not using `&` operator), I'm okay with that too.

Also, I know this file has yet another instance of `bytes.NewBuffer(nil)`, but it's fine to leave that alone, it's too far from the affected code and hence out of scope of this PR.

compiler/vendor/golang.org/x/tools/go/gcexportdata/gcexportdata.go
```diff
@@ -0,0 +1,97 @@
...
+// problems for developer tools that use it, since they must be able to
+// consume the outputs of the gc compiler both before and after a Go
+// update such as from Go 1.7 to Go 1.8.  Because this package lives in
+// golang.org/x/tools, sites can update their version of this repo some
+// time before the Go 1.8 release and rebuild and redeploy their
+// developer tools, which will then be able to consume both Go 1.7 and
+// Go 1.8 export data files, so they will work before and after the
+// Go update. (See discussion at https://github.com/golang/go/issues/15651.)
+//
+package gcexportdata // import "golang.org/x/tools/go/gcexportdata"
```
shurcooL: Removing the now-unused old `gcimporter15` package from `vendor` directory is the right thing to do. However, we shouldn't vendor the new `gcexportdata` package.

This project currently follows the pattern of vendoring as few packages as possible, only the ones that _have_ to be vendored. The new `gcexportdata` package is not deprecated (nor removed from upstream) and doesn't need to be vendored at this time.

That means when newer versions of unvendored dependencies are released, installing the latest version of GopherJS will automatically use the latest versions of said dependencies. If the dependencies are vendored, they're not updated until we manually do so.

At this time, we prefer the reduced maintenance burden by not vendoring when it's not required. This PR should follow the current pattern and not deviate from it.

In the PR description, you said:

> we instead vendor golang.org/x/tools/go/gcexportdata and
> golang.org/x/tools/go/internal/gcimporter to avoid hassle for people
> upgrading to this version of GopherJS.

I don't understand what hassle you're referring to. To update to a newer version of GopherJS, per the [installation instructions](https://github.com/gopherjs/gopherjs#installation-and-usage), people do `go get -u github.com/gopherjs/gopherjs`. That automatically gets the latest versions of dependencies. Please let me know if I'm misunderstanding something, but I don't believe there is an increased hassle for users. (And if there is, we should deal with this outside of this PR if it affects other non-vendored packages too.)

So, please remove the newly vendored packages from this PR, just keep the deletion of `gcimporter15`.

compiler/compiler.go
myitcv: Done.

compiler/package.go
myitcv: Done.

myitcv:
> Let's wait a few days before merging this. I'd like to have a chance to test this change locally a bit to gain more confidence.

Ok, will leave it to you to merge when you're comfortable. 

Incidentally, if there are tests/checks that you perform locally that help to give you confidence that this PR "works" we should look to add those tests/checks (or equivalent tests/checks) to the CI; that way we can just merge changes in the future, confident that things will just "work". 

> We'll also increase the compiler version after merging.

Ok.

shurcooL:
> Incidentally, if there are tests/checks that you perform locally that help to give you confidence that this PR "works" we should look to add those tests/checks (or equivalent tests/checks) to the CI; that way we can just merge changes in the future, confident that things will just "work".

Nope, what I had in mind is just casual testing by trying it locally, and seeing if anything unexpected comes up (e.g., when upgrading from the older version, etc.). I think we have complete test coverage already.

shurcooL:
Tested locally, not seeing any issues. Merging and bumping version.

PR #807, opened 2018-04-26
compiler: fix variadic args not being nil when zero length.

### LINKED ISSUE #806: Omitted variadic function parameter should be nil
### Example
```go
func(s ...string) {
    if s == nil {
        fmt.Println("s is nil")
    }
}()
```
With Go compiler `s == nil` ([playground](https://play.golang.org/p/G2jPMo5sT-0))
With GopherJS `s != nil` ([playground](https://gopherjs.github.io/playground/#/1jQnzf995r))

### Spec
https://golang.org/ref/spec#Passing_arguments_to_..._parameters

myitcv:
Fixes #806.

### DIFF
--- a/tests/compiler_test.go
+++ b/tests/compiler_test.go
@@ -10,6 +10,6 @@ func TestVariadicNil(t *testing.T) {
 	}
 
 	if got := printVari(); got != nil {
-		t.Errorf("expected printVari() to be %v; got: %v", nil, got)
+		t.Errorf("printVari(): got: %#v; want %#v.", got, nil)
 	}
 }
--- a/tests/compiler_test.go
+++ b/tests/compiler_test.go
@@ -12,4 +12,18 @@ func TestVariadicNil(t *testing.T) {
 	if got := printVari(); got != nil {
 		t.Errorf("printVari(): got: %#v; want %#v.", got, nil)
 	}
+
+	{
+		var want []string
+		if got := printVari(want...); got != nil {
+			t.Errorf("printVari(want...): got: %#v; want %#v.", got, nil)
+		}
+	}
+
+	{
+		want := []string{}
+		if got := printVari(want...); got == nil || len(got) != len(want) {
+			t.Errorf("printVari(want...): got: %#v; want %#v.", got, want)
+		}
+	}
 }
--- a/tests/compiler_test.go
+++ b/tests/compiler_test.go
@@ -1,0 +1,17 @@
+// +build js
+
+package tests
+
+import (
+	"testing"
+)
+
+func TestVariadicNil(t *testing.T) {
+	printVari := func(strs ...string) []string {
+		return strs
+	}
+
+	if v := printVari(); v != nil {
+		t.Errorf("expected printVari() to be %v; got: %v", nil, v)
+	}
+}
--- a/tests/compiler_test.go
+++ b/tests/compiler_test.go
@@ -1,5 +1,3 @@
-// +build js
-
 package tests
 
 import (
@@ -11,7 +9,7 @@ func TestVariadicNil(t *testing.T) {
 		return strs
 	}
 
-	if v := printVari(); v != nil {
-		t.Errorf("expected printVari() to be %v; got: %v", nil, v)
+	if got := printVari(); got != nil {
+		t.Errorf("expected printVari() to be %v; got: %v", nil, got)
 	}
 }
--- a/compiler/utils.go
+++ b/compiler/utils.go
@@ -95,6 +95,10 @@ func (c *funcContext) translateArgs(sig *types.Signature, argExprs []ast.Expr, e
 		varargType = sig.Params().At(paramsLen - 1).Type().(*types.Slice)
 	}
 
+	if sig.Variadic() && len(argExprs) == 0 {
+		return []string{fmt.Sprintf("%s.nil", c.typeName(varargType))}
+	}
+
 	preserveOrder := false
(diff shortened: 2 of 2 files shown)


tests/compiler_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+func TestVariadicNil(t *testing.T) {
+	printVari := func(strs ...string) []string {
+		return strs
+	}
+
+	if v := printVari(); v != nil {
+		t.Errorf("expected printVari() to be %v; got: %v", nil, v)
+	}
+}
```
hajimehoshi: Isn't `tests/misc_test.go` a proper place?

bzub:
I tried this out and everything looks good on my end. Thank you.

tests/compiler_test.go
myitcv: The `tests` package needs a bit of a cleanup at some point. The segmentation is, I agree, not great. I even be tempted to have subpackages to better segment things. 

But that belongs in a later PR. Not critical either way in my opinion whether this goes in `misc_test.go` or a new file.

tests/compiler_test.go
```diff
@@ -0,0 +1,17 @@
+// +build js
```
shurcooL: We use `t.Skip` rather than build constraints for skipping tests meant for `GOARCH=js` only, unless the build constraint is necessary for things to compile.

In this case, this test shouldn't be skipped on non-`js`; I want to see it run and pass using `gc` as well (just like all other tests in this package that which can pass with `gc`).

So, remove this line.

tests/compiler_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"testing"
+)
+
+func TestVariadicNil(t *testing.T) {
+	printVari := func(strs ...string) []string {
+		return strs
+	}
+
+	if v := printVari(); v != nil {
```
shurcooL: It'd be nice to be more consistent and stick with the `got, want` test style:

```Go
if got, want := printVari(), []string(nil); got != want {
	t.Errorf("printVari() returned: %v, want: %v", got, want)
}
```

tests/compiler_test.go
myitcv: Done.

tests/compiler_test.go
myitcv: I'm intentionally using untyped `nil` here; but I've renamed the `got` variable.

myitcv:
Thanks @shurcooL. I've pushed up a commit that addresses your feedback.

tests/compiler_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"testing"
+)
+
+func TestVariadicNil(t *testing.T) {
+	printVari := func(strs ...string) []string {
+		return strs
+	}
+
+	if got := printVari(); got != nil {
+		t.Errorf("expected printVari() to be %v; got: %v", nil, got)
```
shurcooL: I want us to be more consistent with the style used in [most Go tests](https://github.com/golang/go/search?q=want), which is to print the "got" value first, the "want" value second. Mind changing it?

```Go
t.Errorf("printVari(): got: %#v, want: %#v", got, []string(nil))
```

Using `%#v` instead of `%v` has the advantage that it actually differentiates nil slice vs non-nil empty slice. `%v` prints both as `[]`.

myitcv: Done. Just pushed up a commit that does this. 

Please can I ask in return that you start a wiki, that we can then build on, which documents the conventions you would like to follow?

shurcooL: Thanks. But why did you replace `[]string(nil)` with just `nil`? It has a [different behavior](https://play.golang.org/p/6f4GhqZc0Dn). Not critical, just pointing out.

> Please can I ask in return that you start a wiki, that we can then build on, which documents the conventions you would like to follow?

We already have exactly that. Please see https://github.com/gopherjs/gopherjs/wiki/Developer-Guidelines#style.

shurcooL: (Actually, https://play.golang.org/p/L31k_UZViaQ is a better example of what I was pointing out.)

tests/compiler_test.go
hajimehoshi: I was wondering why `compiler_test.go` didn't exist until this PR.

myitcv:
@hajimehoshi indeed, `compiler_test.go` didn't exist before; I took the opportunity of this PR to start the process of splitting things up a bit. Could have waited until a later PR, but I didn't think it was too big a deal to start sooner rather than later!

hajimehoshi:
Yeah, I don't mean we should not `compiiler_test.go` now, but rather I wanted to know the current direction. I'm also fine with splitting this :-)

shurcooL:
I wonder if we forgot to consider the case where an empty but non-nil slice is passed to a variadic function. I.e., the 4th case in this snippet: https://play.golang.org/p/357N56KuktI.

shurcooL:
I tested it, and that case is handled correctly. But, I think it'd be great to add a test case for it (as far as I can see, we don't have one). Mind adding one @myitcv? (Otherwise, I can add it before merging.)

myitcv:
> Mind adding one @myitcv?

No problem @shurcooL I've pushed up a couple of extra tests.

paralin:
Merge?

flimzy:
I've rebased this here: https://github.com/gopherjs/gopherjs/tree/fix_806

PR #812, opened 2018-05-02
compiler/natives/src/golang.org/x/mobile/gl: Add implementation of OpenGL ES 2+3 APIs using WebGL.

BenLubar:
Based on golang/mobile@c909788f9903c3fa409bdba28f909907d69fb3d1.

Adds `gl.JSCanvas`, which returns the canvas associated with the context.


flimzy:
If I understand, you're extending the standard library to add an exported method?

BenLubar:
Yes. The alternative is to set up a global WebGL context somehow in the code that calls this package.

The native implementations of packages that use this set up EGL using cgo before calling NewContext, so the new exported JSCanvas function would only be used by gopherjs versions of those packages and generally not the actual application calling OpenGL functions.

flimzy:
Well, I don't have any authority on this project, but I am pretty sure that modifying the standard library is not something GopherJS will be doing.

I have no familiarity with WebGL, though, so I'm not the one to offer alternatives that would be more acceptable.

hajimehoshi:
IMHO reimplementing (semi-)standard library is not what GopherJS aims. Have you discussed this change with anyone before?

BenLubar:
> However, given that Go is gaining WebAssembly support in 1.11,

I did not know this. As long as there's a way to run external JavaScript from Go when it's compiled to WebAssembly, the way I'd suggest is to just scrap this code and rewrite it once 1.11 is out.

flimzy:
This is a great effort, but really does not belong in the GopherJS compiler. I would encourage you to publish this under your own GitHub repo if you desire.

PR #824, opened 2018-05-25
Check for fields in externalize, internalize

dave:
Check for fields before accessing first struct field in externalize and internalize functions.

Fixes this bug: https://play.jsgo.io/308c189e795b573511cbc4d4606af6b0a60a065a

Note: I don't have node so I can't generated the minified prelude.


compiler/prelude/jsmapping.go
```diff
@@ -113,6 +113,9 @@ var $externalize = function(v, t) {
         }
         return searchJsObject(v.$get(), t.elem);
       case $kindStruct:
+        if (t.fields.length == 0) {
```
hajimehoshi: `===` would be better

compiler/prelude/jsmapping.go
```diff
@@ -347,6 +350,9 @@ var $internalize = function(v, t, recv) {
       case $kindPtr:
         return searchJsObject(t.elem);
       case $kindStruct:
+        if (t.fields.length == 0) {
```
hajimehoshi: ditto

hajimehoshi:
> Note: I don't have node so I can't generated the minified prelude.

Oh I missed this comment. Then I'll try this later. If anyone can do this now, I'm also fine :-)

dave:
Oooh I'll change `==` to `===`...

hajimehoshi:
ping @shurcooL

shurcooL:
Sorry about the delay.

The code in `jsmapping.go` looks good.

There have been changes to how prelude is stored by now, so we'll need to rebase and regenerate it (as was mentioned in comments above) in order to make it mergeable. We should also add 2 tests for this bug/behavior.

hajimehoshi:
Ah right. Thank you.

@dave I was wondering if you still can't do `go generate` on your environment. I'd do later if you want.

dave:
This should be ready to merge now I think?

hajimehoshi:
I think yes.

hajimehoshi:
Oh wait, @dmitshur suggested that we should add tests.

> We should also add 2 tests for this bug/behavior.

flimzy:
@dave We're looking at all old PRs, and see that this one is still outstanding. Unfortunately, the bug link in the original description has died.  Are you able to either describe the bug this fixes, or provide a new reproduction case?  Thanks.

flimzy:
Superceded by #1194

PR #845, opened 2018-07-20
Switch to 1.11beta2

dvic:


dmitshur:
According to the CI failure, the current `go1.11` code is no longer compatible with beta 2. It will need to be updated.

PR #846, opened 2018-07-24
natives/runtime: add pkgpath method for _type

egonelbre:


dmitshur:
Thanks for PR @egonelbre! This looks like the right fix. I’ll try to get to fully reviewing and merging it within the next few days.

PR #852, opened 2018-08-20
compiler/prelude: Keep nil slice nil when slicing.

### LINKED ISSUE #851: Nil slice becomes non-nil after slicing.
According to the Go spec on [slice expressions](https://golang.org/ref/spec#Slice_expressions):

> If the sliced operand of a valid slice expression is a `nil` slice, the result is a `nil` slice.

```Go
var s []int
s = s[0:0]
fmt.Println(s == nil) // should be true
```

GopherJS doesn't currently implement that. Slicing a `nil` slice makes it non-`nil`:

- https://play.golang.org/p/9U18mKBuQ3L
- https://gopherjs.github.io/playground/#/9U18mKBuQ3L

I discovered this bug as part of the work to support Go 1.11, because this bug started to cause `go/doc` tests to fail (see https://github.com/gopherjs/gopherjs/issues/836#issuecomment-414159715).

dmitshur:
According to the Go specification:

> If the sliced operand of a valid slice expression is a nil slice,
> the result is a nil slice.

This behavior was not implemented and not caught previously. Slicing a nil slice would incorrectly make it an empty but non-nil slice. This change fixes that.

The minified prelude was regenerated with:

	go generate ./compiler/prelude

Fixes #851.

### DIFF
--- a/tests/misc_test.go
+++ b/tests/misc_test.go
@@ -641,16 +641,24 @@ func TestSlicingNilSlice(t *testing.T) {
 	t.Run("StaysNil", func(t *testing.T) {
 		var s []int
 		s = s[:]
+		if s != nil {
+			t.Errorf("nil slice became non-nil after slicing with s[:]: %#v, want []int(nil)", s)
+		}
+		s = nil
 		s = s[0:0]
+		if s != nil {
+			t.Errorf("nil slice became non-nil after slicing with s[0:0]: %#v, want []int(nil)", s)
+		}
+		s = nil
 		s = s[0:0:0]
 		if s != nil {
-			t.Errorf("nil slice became non-nil after slicing: %#v", s)
+			t.Errorf("nil slice became non-nil after slicing with s[0:0:0]: %#v, want []int(nil)", s)
 		}
 	})
 	t.Run("Panics", func(t *testing.T) {
 		defer func() {
 			if err := recover(); err == nil || !strings.Contains(err.(error).Error(), "slice bounds out of range") {
-				t.Error("slicing nil slice out of range didn't panic as expected")
+				t.Error("slicing nil slice out of range didn't panic, want panic")
 			}
 		}()
 		var s []int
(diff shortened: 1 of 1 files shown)


tests/misc_test.go
```diff
@@ -635,3 +635,32 @@ func TestTypeConversion(t *testing.T) {
...
 	}
 }
+
+// See https://github.com/gopherjs/gopherjs/issues/851.
+func TestSlicingNilSlice(t *testing.T) {
+	t.Run("StaysNil", func(t *testing.T) {
+		var s []int
+		s = s[:]
+		s = s[0:0]
+		s = s[0:0:0]
```
hajimehoshi: I'd add the same `if s != nil` above this

tests/misc_test.go
```diff
@@ -635,3 +635,32 @@ func TestTypeConversion(t *testing.T) {
...
+		if s != nil {
+			t.Errorf("nil slice became non-nil after slicing: %#v", s)
```
hajimehoshi: I know the expectation, but would it be better to reword this including expectation?

tests/misc_test.go
```diff
@@ -635,3 +635,32 @@ func TestTypeConversion(t *testing.T) {
...
+			}
+		}()
+		var s []int
+		s = s[5:10]
+	})
+	t.Run("DoesNotBecomeNil", func(t *testing.T) {
+		var s = []int{}
+		s = s[:]
+		if s == nil {
+			t.Errorf("non-nil slice became nil after slicing: %#v", s)
```
hajimehoshi: ditto

tests/misc_test.go
dmitshur: Are you suggesting this in case one of the assigments makes `s` non-nil, but the following line makes it nil again, and we don't catch it?

dmitshur:
Done, PTAL.

tests/misc_test.go
hajimehoshi: As `s` is expected to be non-nil, you don't have to reassign nil, but probably reassiging would be better. Thanks!

PR #865, opened 2018-10-03
Go 1.11.1 support (version bump to GopherJS 1.11-2).

### LINKED ISSUE #862: reflect package from Go 1.11.1 doesn't build with GopherJS 1.11-1.
Go 1.11.1 has been released [today](https://groups.google.com/forum/#!topic/golang-announce/pFXKAfoVJqw). It includes some changes to the `reflect` package. The overrides that GopherJS 1.11-1 makes to the `reflect` package are no longer compatible, causing it to not build:

```
$ go version
go version go1.11.1 darwin/amd64
$ gopherjs version
GopherJS 1.11-1
$ gopherjs test reflect
/usr/local/go/src/reflect/value.go:673:60: cannot use t (variable of type *rtype) as *funcType value in argument to funcLayout
/usr/local/go/src/reflect/value.go:336:21: cannot use methodReceiver(op, v, int(v.flag) >> flagMethodShift) (value of type *rtype) as *funcType value in assignment
```

dmitshur:
Go 1.11.1 has been released recently. It includes some changes to the
reflect package, including:

-	golang/go@a2f1c8e2ad723589eed8f01056e29f6156869e53
-	golang/go@3afa9dfa9b00c17c8ad0e42b220ec37329631f2c
-	golang/go@58c9bd9bfbedda966d2ae687f3e23f67ec889bc4

The second commit makes some changes to internal reflect code that
makes GopherJS reflect overrides no longer compatible, causing the
package to no longer build:

	$ gopherjs build reflect
	/usr/local/go/src/reflect/value.go:673:60: cannot use t (variable of type *rtype) as *funcType value in argument to funcLayout
	/usr/local/go/src/reflect/value.go:336:21: cannot use methodReceiver(op, v, int(v.flag) >> flagMethodShift) (value of type *rtype) as *funcType value in assignment

Specifically, the type of return value t of function methodReceiver
was changed from *rtype to *funcType in commit golang/go@3afa9dfa9b00c17c8ad0e42b220ec37329631f2c.

This change updates the native reflect override for methodReceiver
function and Value.call method accordingly.

Regenerate natives with:

	go generate ./compiler/natives

Use Go 1.11.1 in CI.

Bump GopherJS version to GopherJS 1.11-2.

Fixes #862. /cc @bradfitz @katiehockman


bradfitz:
LGTM. It's not a goal for GopherJS to still also support Go 1.11(.0) is it? (Doesn't seem worthwhile if it's difficult.)

dmitshur:
> It's not a goal for GopherJS to still also support Go 1.11(.0) is it? (Doesn't seem worthwhile if it's difficult.)

It would be a goal, but I didn't see any easy way of making it happen (there are only [build tags](url) for `go1.10`, `go1.11`, etc., not minor point releases), and figured it's fine to require latest patch version of Go.

But now that I think about it, it wouldn't be too hard to special-case it in `github.com/gopherjs/gopherjs/build` package. I'll try that this time.

dmitshur:
I'm not completely happy with the complexity this adds, but the benefit of having GopherJS 1.11-2 be able to build and install regardless if the user has Go 1.11.1 or Go 1.11(.0) seems worthwhile. PTAL.

PR #755, opened 2018-02-17
Go 1.10 support (version bump to GopherJS 1.10-1).

shurcooL:
This PR adds support for Go 1.10, which [has been released today](https://blog.golang.org/go1.10).

It is passing CI ✅ on Linux (see [here](https://circleci.com/gh/gopherjs/gopherjs/1344)), and all tests are passing on macOS for me locally (aside from an unrelated upstream issue https://github.com/golang/go/issues/23881). It's fully ready for testing and review!

(I have not tested it on Windows and can't guarantee support there.)

I've been testing the `go1.10` branch on all of my GopherJS packages during the last few weeks, and there are no known show-stoppers at this time.

The aspirational plan is to merge this sometime next week, after it has a chance to go through review and additional testing, unless major show-stopping issues are discovered and take longer to fix. Minor non-critical issues don't need to be resolved as part of this PR, they can be worked on post-release. But any viable opportunities to improve code can be applied.

Please see issue #725 for context. Each individual commit has a commit message that describes what it's doing.

Resolves #725.

PR #552, opened 2016-12-13
Go 1.8 support (version bump to GopherJS 1.8-1).

shurcooL:
This is a WIP PR to track work to support Go 1.8. `go1.8beta1` has been released.

Each commit is bite-sized, with rationale and description, and fixes a specific issue.

It's now at a stage where only some of the standard library tests are failing. I'll make a table to track remaining work.

**Update:** All failing tests have been fixed, but there are some known TODOs remaining in code.

- [x] `database/sql/driver`
	<details>
		<summary>Test Failure Details</summary>
		<pre>--- FAIL: TestValueConverters (0.10s)
		types_test.go:77: test 20: driver.defaultConverter(driver.i(9)) = 9 (driver.i); want 9 (int64)
		types_test.go:77: test 23: driver.defaultConverter(driver.bs([1])) = [1] (driver.bs); want [1] ([]uint8)
	FAIL
	FAIL	database/sql/driver	0.623s</pre>
	</details>

- [x] `hash/crc32`

- [x] `math`

- [x] `io`

- [x] `reflect`

- [x] `runtime`

- [x] `net/rpc/jsonrpc`

	<details>
		<summary>Test Failure Details</summary>
		<pre>/home/ubuntu/gopherjs/test.342850936:4
	Error.stackTraceLimit=Infinity;var $global,$module;if(typeof window!=="undefined"){$global=window;}else if(typeof self!=="undefined"){$global=self;}else if(typeof global!=="undefined"){$global=global;$global.require=require;}else{$global=this;}if($global===undefined||$global.Array===undefined){throw new Error("no global object found");}if(typeof module!=="undefined"){$module=module;}var $packages={},$idCounter=0;var $keys=function(m){return m?Object.keys(m):[];};var $flushConsole=function(){};var $throwRuntimeError;var $throwNilPointerError=function(){$throwRuntimeError("invalid memory address or nil pointer dereference");};var $call=function(fn,rcvr,args){return fn.apply(rcvr,args);};var $makeFunc=function(fn){return function(){return $externalize(fn(this,new($sliceType($jsObjectPtr))($global.Array.prototype.slice.call(arguments,[]))),$emptyInterface);};};var $mapArray=function(array,f){var newArray=new array.constructor(array.length);for(var i=0;i<array.length;i++){n

	Error: runtime error: native function not implemented: sync.runtime_SemacquireMutex
	    at $callDeferred (/home/ubuntu/gopherjs/test.342850936:4:28973)
[...]

Issue #500, opened 2016-08-21
Proposal: Add version string to communicate GopherJS's mature status, better tracking of release history.


shurcooL:
I'd like us to add a version string to the GopherJS compiler. Given how stable and mature it is by now, it should've had one a while ago. This is similar to how Go 1.0 was eventually a stable version name.

Nothing will change regarding how to download/update to latest version. `master` branch will still have the latest "patch" version and we will not be tagging those releases (think of each commit to `master` branch as a patch release). So still `go get -u github.com/gopherjs/gopherjs` to get latest GopherJS. No changes to that - this is purely to help communicate and track history.

The version scheme is:

```
x.x-y
```

Where x.x is the major and minor version of Go, e.g., 1.7 for today's Go 1.7. It'd be 1.6 for Go 1.6.3.

y is the version of GopherJS compiler for its version of Go. It will start at 1 and can increment for each significant change/addition/improvement/breaking change. **It will correspond to new entries in the release history that note what has changed.** It's a way for us to communicate externally the changes in a more summarized way than individual commits.

So, first release for Go 1.7 will be GopherJS 1.7-1. Then 1.7-2, 1.7-3, etc.

Once 1.8 comes out and GopherJS supports it, the version will bump up to 1.8-1.

```
1.7-1
1.7-2
1.7-3
1.8-1
1.8-2
1.9-1
...
```

Note that you can/should always use latest Go patch version. So it's fine to use Go 1.7.1 or 1.7.3 whenever those come out with GopherJS 1.7-1.

I've discussed this with @neelance and he was in agreement; we iterated on the versioning scheme to arrive at the above solution.

neelance:
@shurcooL Close?

shurcooL:
Yes, this is done now via #501 and the blog post.

PR #900, opened 2019-03-03
Support Go1.12

### LINKED ISSUE #887: Support for Go 1.12
When trying to go get goherjs/gopherjs using Go 1.12 (one commit newer than go1.12beta1, golang/go@9ed9df6ca2), I get the following error:

```
$ go get -u github.com/gopherjs/gopherjs
# github.com/gopherjs/gopherjs/compiler
goget/src/github.com/gopherjs/gopherjs/compiler/compiler.go:20:9: undefined: ___GOPHERJS_REQUIRES_GO_VERSION_1_11___
```

As I did not find any open issues about this, I just wanted to write about the workaround for this error I'm currently using until gopherjs gets support for Go 1.12.

```
$ cd $GOPATH/src/github.com/gopherjs/gopherjs
$ git remote add myitcv https://github.com/myitcv/gopherjs
$ git fetch myitcv
$ git checkout go1.12
$ go get ./...
```

This successfully compiles GopherJS for Go 1.12.

```
$ gopherjs version
GopherJS 1.12.0
```

Cheers,
Robin

hajimehoshi:
This PR fixes implementation and tests in order to support Go 1.12.

Fixes #887 

PTAL


circle.yml
```diff
@@ -22,8 +22,8 @@ jobs:
     - run: go generate github.com/gopherjs/gopherjs/compiler/prelude
     - run: diff -u <(echo -n) <(git status --porcelain)
     - run: diff -u <(echo -n) <(gofmt -d .)
-    - run: go tool vet *.go # Go package in root directory.
-    - run: for d in */; do echo $d; done | grep -v tests/ | grep -v third_party/ | xargs go tool vet # All subdirectories except "tests", "third_party".
+    - run: go vet *.go # Go package in root directory.
+    - run: for d in */; do echo ./$d...; done | grep -v ./doc | grep -v ./test | grep -v ./node | xargs go vet # All subdirectories except "doc", "tests", "node*".
```
neelance: Please explain this change.

compiler/natives/src/syscall/syscall_unix.go
```diff
@@ -44,7 +44,7 @@ var syscallModule *js.Object
 var alreadyTriedToLoad = false
 var minusOne = -1
 
-func syscall(name string) *js.Object {
+func syscall_byname(name string) *js.Object {
```
neelance: Maybe `syscallByName`?

compiler/version_check.go
```diff
@@ -1,9 +1,9 @@
...
+// +build !go1.13
 
 package compiler
 
-const ___GOPHERJS_REQUIRES_GO_VERSION_1_11___ = true
+const ___GOPHERJS_REQUIRES_GO_VERSION_1_12___ = true
 
 // Version is the GopherJS compiler version string.
-const Version = "1.11-2"
+const Version = "1.12-wip"
```
neelance: This shouldn't be `wip`.

compiler/natives/src/reflect/reflect.go
```diff
@@ -550,21 +550,62 @@ type mapIter struct {
...
-	k := iter.keys.Index(iter.i)
-	return unsafe.Pointer(js.Global.Call("$newDataPointer", iter.m.Get(k.String()).Get("k"), jsType(PtrTo(iter.t.Key()))).Unsafe())
+func mapiterkey(it unsafe.Pointer) unsafe.Pointer {
+	iter := (*mapIter)(it)
+	var kv *js.Object
+	if iter.last != nil {
+		kv = iter.last
+	} else {
+		// Compare the index and the size of the actual key set, and check if the iterator is already exhausted.
+		if iter.i >= js.Global.Call("$keys", iter.m).Length() {
```
neelance: Calling `$keys` again does not look right at first glance. Which bug is this supposed to fix?

circle.yml
hajimehoshi: * `go tool vet` is deprecated and `go vet` is the substitute. ([Error](https://circleci.com/gh/gopherjs/gopherjs/1672))
* It looked like `go vet` required the relative paths for local packages. I thought this line tries to `go vet` against part of the local directories. Is that correct?

compiler/natives/src/reflect/reflect.go
hajimehoshi: * [`reflect.TestMapIterDelete1`](https://github.com/golang/go/blob/4fb900e9ca1f08c57b074e7bf6a7eab90b92c898/src/reflect/all_test.go#L6686-L6700)

`Value` tries to access `mapiterkeys` at the second and later `Next` call ([code](https://github.com/golang/go/blob/4fb900e9ca1f08c57b074e7bf6a7eab90b92c898/src/reflect/value.go#L1197), and `iter.keys` can be already not synced with `$keys` by deleting keys in a loop.

On second thought, my fix might not work with the below code:

```go
iter := ValueOf(m).MapRange()
for iter.Next() {
    delete(m, iter.Key())
}
if len(m) != 0 {
    panic("not reached")
}
```

Let me think...

compiler/natives/src/syscall/syscall_unix.go
hajimehoshi: Done.

compiler/version_check.go
hajimehoshi: Done.

hajimehoshi:
I found another problem: the test doesn't work well on macOS (e.g. `go run -tags=gopherjsdev . test --short reflect` takes forever). I'll fix this later before merging this.

circle.yml
dmitshur: This change makes sense to me, as @hajimehoshi described above.

A few questions/suggestions:

1. I suspect the `go tool vet *.go` line can be simplified even further to just `go vet`, or `go vet .`, to run vet on the Go package with the import path `github.com/gopherjs/gopherjs`. It was `*.go` previously because `go tool vet` did not accept import paths, only files and directories.

2. Does `doc` directory need to be skipped? It has just 2 .md files, so the import path pattern `./doc/...` should not be matching any packages.

compiler/version_check.go
dmitshur: @neelance Why not? I think it should be `wip` while this PR is in active development and review, and changed to `1.12-1` only when it's about to be merged.

That way, when you run `gopherjs version` and get `1.12-1`, it's guaranteed to be the final version and not an older in-development one.

It's also how we've done the last 3+ releases. See https://github.com/gopherjs/gopherjs/pull/853/commits/8b6e87829e4f5245fde933d574a6cd89ec008d9b for example.

circle.yml
hajimehoshi: > 1.

Done.

> 2.

`go vet ./doc/...` fails with the exit code non-zero, and I thought I needed to avoid.

Would `go vet ./...` work btw?

compiler/version_check.go
hajimehoshi: Sooner or later we'll change `1.12` before merging, so I think changing this now should be fine.

circle.yml
hajimehoshi:
```
hajimehoshi@Hajimes-MacBook-Pro ~/go/src/github.com/gopherjs/gopherjs
$ go vet ./docs/...
go: warning: "./docs/..." matched no packages
no packages to vet

hajimehoshi@Hajimes-MacBook-Pro ~/go/src/github.com/gopherjs/gopherjs
$ echo $?
1
```

hajimehoshi:
In the current situation:

```
hajimehoshi@Hajimes-MacBook-Pro ~/go/src/github.com/gopherjs/gopherjs
$ go build . && env GOOS=linux ./gopherjs test -v github.com/gopherjs/gopherjs/tests/main
gopherjs: Source maps disabled. Install source-map-support module for nice stack traces. See https://github.com/gopherjs/gopherjs#gopherjs-run-gopherjs-test.                                                                             
=== RUN   TestNotRunMain
--- PASS: TestNotRunMain (0.00s)
PASS
fatal error: all goroutines are asleep - deadlock!
FAIL    github.com/gopherjs/gopherjs/tests/main 0.268s
hajimehoshi@Hajimes-MacBook-Pro ~/go/src/github.com/gopherjs/gopherjs

$ go build . && env GOOS=darwin ./gopherjs test -v github.com/gopherjs/gopherjs/tests/main
gopherjs: Source maps disabled. Install source-map-support module for nice stack traces. See https://github.com/gopherjs/gopherjs#gopherjs-run-gopherjs-test.                                                                             
fatal error: all goroutines are asleep - deadlock!
FAIL    github.com/gopherjs/gopherjs/tests/main 0.246s
```

Now I'm investigating why the tests don't work on Darwin.

dmitshur:
Did you install source-map-support module and syscall module? The ones mentioned at https://github.com/gopherjs/gopherjs#gopherjs-run-gopherjs-test. They used to be more optional but by now I think they're mandatory for some tests.

compiler/natives/src/reflect/reflect.go
hajimehoshi: Fixed.

hajimehoshi:
> Did you install source-map-support module and syscall module?

No, but I think the cause was that the `write` implementation was not used on Darwin and I fixed syscall infrastructure for Darwin.

hajimehoshi:
Now all the known issues have been solved. Please take another look. Thank you!

myitcv:
Just adding some notes based on us applying roughly the same patch to https://github.com/myitcv/gopherjs/pull/41:

* I think the `*_js.go` files need to be renamed and build constraints used
* Because the CircleCI [config file](https://github.com/gopherjs/gopherjs/blob/d547d1d9531ed93dbdebcbff7f83e7c876a1e0ee/circle.yml#L30-L138) specifies the standard library tests to run (as opposed to a [set negation approach](https://github.com/myitcv/gopherjs/blob/bb0d1242f166dd8ec2e65818f364f2ff32d176cc/.circleci/config.yml#L84)) this PR is not running tests in some of the new standard library packages, crucially `internal/fmtsort`, so the fact the tests are passing for this PR is (I think) currently a false positive

```diff
$ diff -wu 1.11 1.12
--- 1.11        2019-03-08 18:46:10.181963351 +0000
+++ 1.12        2019-03-08 18:46:02.570274117 +0000
@@ -88,6 +88,8 @@
 index/suffixarray
 internal/bytealg
 internal/cpu
+internal/fmtsort
+internal/goroot
 internal/nettrace
 internal/poll
 internal/race
@@ -99,6 +101,27 @@
 internal/testenv
 internal/testlog
 internal/trace
+internal/x/crypto/chacha20poly1305
+internal/x/crypto/cryptobyte
+internal/x/crypto/cryptobyte/asn1
+internal/x/crypto/curve25519
+internal/x/crypto/hkdf
+internal/x/crypto/internal/chacha20
+internal/x/crypto/poly1305
+internal/x/net/dns/dnsmessage
+internal/x/net/http/httpguts
+internal/x/net/http/httpproxy
+internal/x/net/http2/hpack
+internal/x/net/idna
+internal/x/net/internal/nettest
+internal/x/net/nettest
+internal/x/text/secure
+internal/x/text/secure/bidirule
+internal/x/text/transform
+internal/x/text/unicode
+internal/x/text/unicode/bidi
+internal/x/text/unicode/norm
+internal/xcoff
 io
 io/ioutil
 log
@@ -143,6 +166,7 @@
 runtime/cgo
 runtime/debug
 runtime/internal/atomic
+runtime/internal/math
 runtime/internal/sys
 runtime/pprof
 runtime/pprof/internal/profile
@@ -167,22 +191,3 @@
 unicode/utf16
 unicode/utf8
 unsafe
-vendor/golang_org/x/crypto/chacha20poly1305
-vendor/golang_org/x/crypto/cryptobyte
-vendor/golang_org/x/crypto/cryptobyte/asn1
-vendor/golang_org/x/crypto/curve25519
-vendor/golang_org/x/crypto/internal/chacha20
-vendor/golang_org/x/crypto/poly1305
-vendor/golang_org/x/net/dns/dnsmessage
-vendor/golang_org/x/net/http/httpguts
-vendor/golang_org/x/net/http/httpproxy
-vendor/golang_org/x/net/http2/hpack
-vendor/golang_org/x/net/idna
-vendor/golang_org/x/net/internal/nettest
-vendor/golang_org/x/net/nettest
-vendor/golang_org/x/text/secure
-vendor/golang_org/x/text/secure/bidirule
-vendor/golang_org/x/text/transform
-vendor/golang_org/x/text/unicode
-vendor/golang_org/x/text/unicode/bidi
-vendor/golang_org/x/text/unicode/norm
```

hajimehoshi:
@myitcv Thank you! For the above diff, the original master's circle.yml doesn't include a lot of tests like `internal/cpu`. Would it be possible to decide to care such packages in another PR if needed?

> I think the *_js.go files need to be renamed and build constraints used

Done.

myitcv:
@hajimehoshi 

> For the above diff, the original master's circle.yml doesn't include a lot of tests like internal/cpu. Would it be possible to decide to care such packages in another PR if needed?

I'd suggest switching to the negation-based approach. We can even do that ahead of the Go 1.12 changes. 

It's a simple, easy to understand change that means we won't have to manually keep track of new packages. Packages like `internal/fmtsort` which, per https://github.com/myitcv/gopherjs/pull/41, seem to have caught an issue.

hajimehoshi:
> I'd suggest switching to the negation-based approach. We can even do that ahead of the Go 1.12 changes.

A blacklist instead of the current whitelist makes sense.

.std_test_pkg_exclusions
```diff
@@ -0,0 +1,86 @@
...
+go/importer
+go/internal/gccgoimporter
+go/internal/gcimporter
+go/internal/srcimporter
+go/types
+hash
+image/color/palette
+image/internal/imageutil
+internal/cpu
+internal/fmtsort
```
myitcv: A better starting point for this blacklist is the list of packages that were not tested as part of the 1.11 cycle (hence why I suggested we could/should do that as part of a separate, simple PR). 

That would then highlight that you're including `internal/fmtsort` in this file, i.e. it would be an additional package we're ignoring for Go 1.12.

I still think we should understand why this package in particular is failing, because it's fundamental to the sorting changes in the `fmt` package and other places:

```
Package fmtsort provides a general stable ordering mechanism for maps, on behalf of the fmt and text/template packages.
```

hajimehoshi: Sure. As I merged your PR, I'll be investigating why `internal/fmtsort` tests failed. I agree `fmtsort` is so fundamental that we need to care.

hajimehoshi: The root cause is that getting pointer values from a slice is now counterintuitive

https://gopherjs.github.io/playground/#/jvnZoFsINA

hajimehoshi: Ha, nice, my guess is that conversion an integer from a pointer (in `reflect`?) causes this issue

https://gopherjs.github.io/playground/#/fN4zm6Wj4W

hajimehoshi: https://github.com/gopherjs/gopherjs/blob/master/compiler/expressions.go#L982 I think this is the culprit

hajimehoshi: I made the tests pass with these fixes:

* For pointers and channels, the current GopherJS cannot treat them as well for sorting. Skip this.
* GopherJS cannot treat nil key well in a map. Skip this.

compiler/version_check.go
dmitshur: I would prefer to see the `-wip` suffix when the version is in development. Removing it and replacing it with `-1` should be done when the merge is happening.

This is consistent with how the version of Go toolchain gets bumped only when the release is happening, not sooner. E.g., see https://github.com/golang/go/commit/05e77d41914d247a1e7caf37d7125ccaa5a53505.

circle.yml
dmitshur: > Would `go vet ./...` work btw?

I don't think it would, because the intention is to skip "tests" directory from being vetted.

compiler/version_check.go
hajimehoshi: Just to confirm, will we squash these commits when merging?

dmitshur: No, we've never done that in the past. Squashing would create a giant commit without a meaningful commit message.

It has always been a merge commit, typically after rebasing the PR on top of latest `master` branch and cleaning up its history a bit.

See previous PRs and their merge commits:

- Go 1.11 support - PR #853 - merge commit https://github.com/gopherjs/gopherjs/commit/0210a2f0f73c96103378b0b935f39868e5731809
- Go 1.10 support - PR #755 - merge commit - https://github.com/gopherjs/gopherjs/commit/4e9f423448a0b0462d41b077c43bb9f2a5d103c6
- Go 1.9 support - PR #651 - merge commit https://github.com/gopherjs/gopherjs/commit/f7c56533adfac99e12ccdccfec04a466d2a778da
- Go 1.8 support - PR #552 - merge commit https://github.com/gopherjs/gopherjs/commit/b9bcb1da229a59cc1e1d168401662cb6450aae08
- Go 1.7 support - PR #495 - merge commit https://github.com/gopherjs/gopherjs/commit/b33dd36ad3f3b02fc01a860f94becdb4e9a95ddc
- Go 1.6 support - PR #404 - merge commit https://github.com/gopherjs/gopherjs/commit/f8510c9aef19750ec5a1178b33932a7f9defd197

GopherJS versions were proposed in issue #500 and implemented as of GopherJS 1.7-1 (after it was already on `master`). So if you look at PRs for 1.8 and newer, you'll find a "GopherJS 1.x-wip" -> "GopherJS 1.x-1" bump commit at the very end. E.g.:

- https://github.com/gopherjs/gopherjs/pull/853/commits/8b6e87829e4f5245fde933d574a6cd89ec008d9b - bump version from 1.11-wip to 1.11-1
- https://github.com/gopherjs/gopherjs/pull/755/commits/4ebbd33f7fb1fca27e0a0a413c2729ae8cb4d958 - bump version from 1.10-wip to 1.10-1
- https://github.com/gopherjs/gopherjs/pull/651/commits/fe63fb1b9957f490d5011c8bb0a254db64e6a138 - bump version from 1.9-wip to 1.9-1
- https://github.com/gopherjs/gopherjs/pull/552/commits/0ff5023f798422e0389990d6edff2b06e7861e4d - notable exception, but it's because the history of the `go1.8` branch was re-written before merged and the wip->1 bump was squashed into the initial version commit

hajimehoshi: Done.

hajimehoshi:
Friendly ping

compiler/natives/src/reflect/reflect.go
```diff
@@ -550,20 +550,63 @@ type mapIter struct {
 	m    *js.Object
 	keys *js.Object
 	i    int
+	last *js.Object
```
dmitshur: It'll be helpful to document this new field.

compiler/natives/fs_vfsdata.go
```diff
@@ -21,583 +21,635 @@ var FS = func() http.FileSystem {
...
 		},
 		"/src/net/http/http.go": &vfsgen۰CompressedFileInf
```
dmitshur: This file shouldn't be added.

compiler/natives/fs_vfsdata.go
```diff
@@ -21,583 +21,635 @@ var FS = func() http.FileSystem {
...
+		},
+		"/src/internal/fmtsort/fmtsort_test.go~": &vfsgen۰FileInfo{
```
dmitshur: Same here.

compiler/version_check.go
dmitshur: When you're ready to merge the PR, push a commit that updates the version to `1.12-1` just before.

compiler/natives/fs_vfsdata.go
dmitshur: GitHub seems to be displaying this comment at an incorrect spot in the discussion tab (it's not the same as in the "Files changed" tab). I left it on line 617, but it's showing it on line 346 here. 😕

To be clear, the two files I was referring to are:

- `/src/testing/testing_test.go~`
- `/src/internal/fmtsort/fmtsort_test.go~`

hajimehoshi: Nice catch: done.

Can we fix the generator to skip such files?

hajimehoshi:
@dmitshur Thank you for reviewing! I appreciate your careful work.

Should we squash the commits, or not? My preference is to squash them, but I'd like to obey the policy of GopherJS.

dmitshur:
When you ask if we should squash the commits, can you clarify what exactly you mean? I can't answer your question directly because I'm not sure what you meant, so I'll describe this in full.

When merging this PR, we should use the "create a merge commit" merge strategy:



When creating the merge commit, we can follow the same pattern as in previous merge commits, e.g., https://github.com/gopherjs/gopherjs/commit/0210a2f0f73c96103378b0b935f39868e5731809. You'll have to edit the subject and body fields accordingly. E.g.:

image

We should not use the "squash and merge" merge strategy. Doing so would squash all commits in this PR into a single commit. Such a commit would be massive, it would include all generated code, and it's not viable to meaningfully describe the changes that went into adding Go 1.12 support.

However, before doing the merge, it is completely acceptable and in fact preferable to re-write the history to clean it up and squash redundant commits. At the very least:

- all the generated commits can be squashed into a single commit at the end that re-generates all data; see [third-last commit](https://github.com/gopherjs/gopherjs/pull/853/commits) in PR #853 that added Go 1.11 support

If any of the other commits are redundant or WIP (e.g., where you tried one thing, but ended up overwriting it a future commit) and can be squashed into a single logical commit with a descriptive commit message, that's great too.

Of course, re-writing the history involves a force push to this PR branch, so be careful, and do a diff of the final result against the starting point (i.e., the latest commit is a02f42a28d940ae027c87396698d9820cc5b28ec right now) to ensure the end result is unchanged.

Finally, feel free to look over the previous PRs to get an idea of how it was done historically. I've linked them all in https://github.com/gopherjs/gopherjs/pull/900#discussion_r264015568.

Cleaning up the PR history and writing more descriptive commit messages is great, but optional. Feel free to do as much or as little before merging this PR. Thanks again @hajimehoshi!

compiler/natives/fs_vfsdata.go
dmitshur: Yes, it's Go code, and it's possible to implement any logic for what files to include/exclude. See here:

https://github.com/gopherjs/gopherjs/blob/847fc94819f9d5a4e7154e2203b36f6dbace6f48/compiler/natives/fs.go#L15-L21

I'm not sure if it's worth it, but I'll let you decide.

hajimehoshi:
> We should not use the "squash and merge" merge strategy. 

What I wanted to ask is whether we should use this strategy. Sorry for my unclear question. Thanks!

hajimehoshi:
I removes some redundant commits, but I couldn't integrate go-generating into one commit. That's not perfect, but I believe that's acceptable.

PR #902, opened 2019-03-09
ci: use black list instead of white list for standard library tests

myitcv:
Using a black list is a more effective mechanism for expressing those
standard library packages we don't want to test because it means we
don't need to manually check whether new standard library packages have
been added that would otherwise need to be added to the white list.

See the discussion in https://github.com/gopherjs/gopherjs/pull/900 for
more context.


dmitshur:
`grep -v -x -f .std_test_pkg_exclusions` is behaving differently on macOS compared to Linux. :(

On Linux (working as expected):

```
$ grep --version
grep (GNU grep) 2.25
[...]
$ echo -e "fmt\nruntime\nruntime/cgo" | grep -v -x -f .std_test_pkg_exclusions
fmt
```

On macOS (not filtering out `runtime/cgo`):

```
$ grep --version
grep (BSD grep) 2.5.1-FreeBSD
$ echo -e "fmt\nruntime\nruntime/cgo" | grep -v -x -f .std_test_pkg_exclusions
fmt
runtime/cgo
```

PR #903, opened 2019-03-09
ci: upgrade to Go 1.11.5

myitcv:
So that we can then see the true diff that the Go 1.12 changes in
#900 will introduce.


hajimehoshi:
Thank you!

```
# go run run.go -- fixedbugs/issue28688.go
exit status 2
Error: unknown shorthand flag: 'g' in -gcflags=-d=softfloat
Usage:
  gopherjs run [gofiles...] [arguments...] [flags]

Flags:
      --color         colored output
  -h, --help          help for run
      --localmap      use local paths for sourcemap
  -m, --minify        minify generated code
  -q, --quiet         suppress non-fatal warnings
      --tags string   a list of build tags to consider satisfied during the build
  -v, --verbose       print the names of packages as they are compiled


FAIL	fixedbugs/issue28688.go	0.006s
```

Is this an intended error?

myitcv:
@hajimehoshi - the way the error is reported looks like another issue, but this test can itself be ignored because it looks related to checking runtime optimisations. Hence I've added it to the list of `notApplicable` tests that can be ignored.

PR #907, opened 2019-04-03
Remove vendored from misc_test.go

hajimehoshi:
`vendored` package was introduced at f8510c9aef19750ec5a1178b33932a7f9defd197,
but now the package has gone. I presume that testing `vendored`
package no longer makes sense.


hajimehoshi:
Sorry but this was my misunderstanding. `tests/vendor/vendored` still exists.

PR #911, opened 2019-04-10
compiler/natives/src/time: disable TestSleep

hajimehoshi:
As `time.Now()` is not accurate enough, the test `time_test.TestSleep` is flaky. This PR disables the test.


dmitshur:
Minor, but it would be nicer to merge a small PR like this via the "squash and merge" strategy, so it would generate just 1 commit on master branch rather than two. This is because we're trying to follow the convention set by the Go project.

hajimehoshi:
Certainly. I'll be careful later. Thanks!

Issue #899, opened 2019-03-03
Unify GopherJS's `syscall/js` and GopherWasm


hajimehoshi:
@neelance suggested to unify [GopherWasm](https://github.com/gopherjs/gopherwasm) and GopherJS by making GopherJS recognize `syscall/js` as a wrapper like GopherWasm. This would enable the users to replace

```
import "github.com/gopherjs/gopherwasm/js"
```

with

```
import "syscall/js"
```

while the programs still work both for GopherJS and Wasm. This simplifies things very much and encourages users to migrate from GopherJS to Wasm more smoothly than the current GopherWasm.

Any thoughts?

hajimehoshi:
I'll get started soon.

inliquid:
@hajimehoshi thanks for your great work for gopherjs, may I ask in this context what would you recommend if I'm actively using `honnef.co/go/js/dom` and `honnef.co/go/js/xhr`, while keeping in mind possible wasm move in future? Should I stick with these libs or is there any solution which better suits in the wasm space? Maybe some support is planned by `honnef.co` maintainers?

hajimehoshi:
Hi, I have no idea about these projects. Please ask the owners.

inliquid:
Well it's not about these projects, but generally about your vision of moving towards wasm in the context of this particular adoption keeping in mind that many ppl still (and will) be using other libs as well. Ok your point is clear.

johanbrandhorst:
@inliquid I don't wish to speak for @dominikh but in the past he has expressed his intent not to spend anymore time on adapting the existing libraries to webassembly. I think there are alternative efforts to have a syscall/js-based DOM interface such as https://github.com/dennwc/dom. Ideally for the DOM specifically we would generate Go bindings directly from the WebIDL spec like Rust does. @dennwc was doing some work in this area last year sure if that effort is still ongoing. Would you be interested in contributing to this?

inliquid:
Hi @johanbrandhorst! Thank you for detailed comments, I will follow mentioned repo activity and try to contribute when possible.

johanbrandhorst:
Will we get another release with this?

hajimehoshi:
I guess GopherJS does not have a 'release' policy. Not sure though.

dmitshur:
The release policy is we increment the version number by 1 whenever there's a notable or large change. It's just for convenience, so people can tell if their GopherJS version includes said notable/large change (and we can ask them in bug reports).

[...]

PR #914, opened 2019-04-30
compiler: update version to 1.12-2

hajimehoshi:
Introducing `syscall/js` (#899) is so significant that we update
the version.


hajimehoshi:
Please take a look :)

PR #933, opened 2019-08-11
build, compiler/natives/src/reflect: remove Go 1.11-specific code

dmitshur:
GopherJS 1.12 is out by now, so it's safe to remove this code that
was added to deal with differences between Go 1.11.0 and Go 1.11.1.

There should not be any change in behavior.

Regenerate `./compiler/natives`.

Updates #862.


dmitshur:
> There should not be any change in behavior, except the internal and
> temporary `go1.11.1` build tag will no longer be set by GopherJS.
> It was never intended to be used externally, so hopefully nobody did.

Never mind, that tag was never available externally. It was added to `nativesContext`, which is the context used to import std package overrides from the `natives` filesystem.

build/build.go
```diff
@@ -316,13 +316,6 @@ func parseAndAugment(bctx *build.Context, pkg *build.Package, isTest bool, fileS
 		},
 	}
 
-	// reflect needs to tell Go 1.11 apart from Go 1.11.1 for https://github.com/gopherjs/gopherjs/issues/862,
-	// so provide it with the custom go1.11.1 build tag whenever we're on Go 1.11.1 or later.
-	// TODO: Remove this ad hoc special behavior in GopherJS 1.12.
-	if runtime.Version() != "go1.11" {
-		nativesContext.ReleaseTags = append(nativesContext.ReleaseTags, "go1.11.1")
-	}
-
```
hajimehoshi: As `runtime.Version() != "go1.11"` becomes always true, `nativesContext.ReleaseTags = append(nativesContext.ReleaseTags, "go1.11.1")` should always be kept. Is this correct?

dmitshur: It should not be kept, because it's no longer needed. This is described in the commit PR description and my comment above.

hajimehoshi: OK I understood this was originally not needed.

compiler/natives/src/reflect/reflect.go
```diff
@@ -694,6 +694,39 @@ func Copy(dst, src Value) int {
 	return js.Global.Call("$copySlice", dstVal, srcVal).Int()
 }
 
+func methodReceiver(op string, v Value, i int) (_ *rtype, t *funcType, fn unsafe.Pointer) {
```
hajimehoshi: I'm not sure this PR's intention. Does this PR remove unnecessary things? I was wondering why you added new code.

dmitshur: There were two copies of these functions, one for Go 1.11.0 in reflect_go111.go and another for Go 1.11.1+ in reflect_go1111.go.

The copy in reflect_go111.go is no longer ever needed, because we don't support Go 1.11.0 in GopherJS 1.12. This PR removes it. It moves the actively used code from reflect_go1111.go back into reflect.go, since there's no need for it to be in a separate file anymore.

It effectively reverts PR #865, which added complexity that was needed at the time to fix issue #865.

dmitshur:
Thank you for reviewing.

PR #935, opened 2019-08-18
Syscalls does not yet support version 12.0.0

pjdufour:
Syscalls are not compiling on Ubuntu for NodeJS 12.  This PR simply updates the docs to show how to use an older version of Node with syscalls.

### DIFF
--- a/doc/syscalls.md
+++ b/doc/syscalls.md
@@ -13,14 +13,24 @@ The JavaScript environment of a web browser is completely isolated from your ope
 
 ### Node.js on Linux and macOS
 
-GopherJS has support for system calls on Linux and macOS. Before running your code with Node.js, you need to install the system calls module. The module is compatible with Node.js version 10.0.0 (or newer). If you want to use an older version you can opt to not install the module, but then system calls are not available.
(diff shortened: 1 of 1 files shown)


doc/syscalls.md
```diff
@@ -13,14 +13,24 @@ The JavaScript environment of a web browser is completely isolated from your ope
 
 ### Node.js on Linux and macOS
 
-GopherJS has support for system calls on Linux and macOS. Before running your code with Node.js, you need to install the system calls module. The module is compatible with Node.js version 10.0.0 (or newer). If you want to use an older version you can opt to not install the module, but then system calls are not available.
+GopherJS has support for system calls on Linux and macOS. Before running your code with Node.js, you need to install the system calls module. The module is compatible with Node.js versions from 10.0.0 to 11.15.0 (Node.js version 12 is not yet supported).  If you want to use an older version you can opt to not install the module, but then system calls are not available.
```
flimzy: I would suggest changing this to:

version >= 10.0.0 and < 12.0.0

to remain valid if 11.15.1, or 11.16.0 are ever released.

Issue #947, opened 2019-10-20
Experiment: use go/packages instead of go/build for easier build system support


nevkontakte:
Earlier this year, Go Team presented [golang.org/x/tools/go/packages](https://godoc.org/golang.org/x/tools/go/packages) as build system-agnostic interface for tooling authors. At the moment GopherJS uses go/build as a main way to load packages, which lacks Go Modules support and is strongly tied to the `go` tool.

In theory, switch to go/packages should offer several advantages:

  * Transparent build system integration, lesser chance to be affected by changes in go build tooling. As a corollary, this should provide Go Module support "for free".
  * An avenue for integrating GopherJS with other build systems (bazel.io and friends).
  * A possibility to remove a lot of custom code from the gopherjs/build package, which is handling loading packages in different ways.

How much of that proves to be true, is yet to be seen, but I'll start an experiment.

nevkontakte:
Citint my comment from the pull request:

> A somewhat late update: I was able to make go/packages work, but it required a number of inelegant hacks and did not actually reduce complexity nearly as much as I thought it would, so I am no longer pursuing this PR. I might try to retrofit some fragments of refactoring I did as a part or it into the existing code, but it will depend on whether a viable fork of GopherJS emerges.
>
> For the posterity, the main difficulty I encountered was that go/packages is really designed to suit needs of developer tooling and editors, rather than a third-party compiler. In particular, GopherJS augments many stdlib packages on the fly to make them work in the browser environment. And even though go/packages provides some hooks for transforming/patching AST on the fly, it doesn't provide enough context about the package for GopherJS to do its thing. I did manage to do it with a bunch of hacks and heuristics, but they are fragile and complex, which undermined the original goal of this PR.

I wouldn't say there's no merit to this idea, but in practice it doesn't seem easy to implement either, so I am declaring this experiment concluded for the time being.

PR #948, opened 2019-10-20
Experimental support for go/packages in GopherJS build system

nevkontakte:
As described in #947, go/packages might be an opportunity for GopherJS to shed some maintenance burden and gain better build system support at the same time.

I'm opening this PR mostly as an early preview of my progress, it is **not** ready to be merged just yet. I'm writing the code in such a way that it would be able to coexist with the current implementation until we decide it is good enough to be the default.

The end goal for this PR is to reach feature parity for `gopherjs build` command. At the moment, it is able to produce sensible-ish JS artifacts, but I haven't implemented support for `natives` VFS as well as standard library packages pruning.

I'll be updating this PR as I go.

### DIFF
--- a/build/v2/augmenter.go
+++ b/build/v2/augmenter.go
@@ -1,0 +1,175 @@
+package build
+
+import (
+	"fmt"
+	"go/ast"
+	"go/build"
+	"go/parser"
+	"io"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/gopherjs/gopherjs/compiler/natives"
+	"golang.org/x/tools/go/ast/astutil"
+	"golang.org/x/tools/go/packages"
+)
+
+// SymbolFilter implements top-level symbol pruning for augmented packages.
+//
+// GopherJS provides custom implementations of some parts of standard library, which are added
+// on top of the original stdlib sources. To avoid compilation errors due to duplicate symbols,
+// SymbolFilter can Collect top-level symbols from overlay sources and then Prune conflicting
+// symbols from the original sources.
+type SymbolFilter map[string]bool
+
+func (sf SymbolFilter) funcName(d *ast.FuncDecl) string {
+	if d.Recv == nil || len(d.Recv.List) == 0 {
+		return d.Name.Name
+	}
+	recv := d.Recv.List[0].Type
+	if star, ok := recv.(*ast.StarExpr); ok {
+		recv = star.X
+	}
+	return recv.(*ast.Ident).Name + "." + d.Name.Name
+}
+
+// traverse top-level symbols within the file and prune top-level symbols for which keep() returned
+// false.
+//
+// This function is functionally very similar to ast.FilterFile with two differences: it doesn't
+// descend into interface methods and struct fields, and it preserves imports.
+func (sf SymbolFilter) traverse(f *ast.File, keep func(name string) bool) {
+	astutil.Apply(f, func(c *astutil.Cursor) bool {
+		switch d := c.Node().(type) {
+		case *ast.File: // Root node.
+			return true
+		case *ast.FuncDecl: // Child of *ast.File.
+			if !keep(sf.funcName(d)) {
+				c.Delete()
+			}
+		case *ast.GenDecl: // Child of *ast.File.
+			return c.Name() == "Decls"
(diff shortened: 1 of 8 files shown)


nevkontakte:
By the way, any early feedback is very welcome :-)

nevkontakte:
I have been looking for ways how GopherJS's standard library augmentations could be implemented on top of go/packages.

As far as I can tell there are three ways how GopherJS modifies the standard library:

 1. Replaces "sync" package with "nosync" implementation. For a reason I don't entirely understand, this is only applied to [some packages in standard library](https://github.com/gopherjs/gopherjs/blob/master/build/build.go#L388), but not anything else.
 2. [Prunes some files](https://github.com/gopherjs/gopherjs/blob/master/build/build.go#L163-L175) from standard library packages. AFAICT the pruned files contain code that is not GopherJS-compatible (or necessary) in some ways.
 3. [Replaces at AST level](https://github.com/gopherjs/gopherjs/blob/master/build/build.go#L259) some of the definitions with GopherJS-compatible implementations.

To support those in go/packages implementation there are a few tool available:

 1. Static overlays in Config.Overlay option. This allows either replacing the entire contents of the file with something else or injecting a new file, assuming it is for a known package.
 2. Intervention at file-level AST parsing stage with a custom Config.ParseFile function. Unfortunately, there's no package-level metadata available at this stage, and order of invocation is not guaranteed.
 3. Custom post-processing after all files were loaded and parsed, somewhat similar to what GopherJS is currently doing.

Ideally I'd like to offload as much work to go/packages as possible, but its interface has been heavily geared towards serving needs of LSP and editors' integration, and fancy augmentation that GopherJS is doing is somewhat difficult. Another advantage of go/packages is that it is going a lot of work in parallel. 

My experiments show so far that pruning files at ParseFile() stage works pretty well. The only inelegant aspect is inferring standard library package name from a bare file path, but with a help of runtime.GOROOT() that seems to work okay, at least for the standard go build system.

Replacing "sync" with "nosync" should be also very easy with ParseFile().

What is proving challenging is augmenting packages with natives' implementations. Static overlays won't apply in any straightforward way because files in natives don't map onto "real" files in most cases. Overlays+ParseFile could be used to do AST rewriting, but we'd need to make sure that natives implementations are loaded first and prune conflicting definitions from other files, which is going to be ugly. Post-processing seems the most reasonable options so far but it would require us to do type-checking and loading natives the old-fashioned way instead of letting go/packages do the work for us.

So my current plan is to try to do augmentation after `go/packages` does its work and do type-checking afterwards and see how much code it would be.

nevkontakte:
Ran into another issue with go/packages: https://github.com/golang/go/issues/35331

This is pretty unfortunate because until it's fixed I have to let go/packages to do its type-checking which limits a degree to which I can mess with the code being loaded at the parse stage to make sure it's type-correct. That type check is also redundant because gopherjs will do its own on the augmented code anyway.

I'll wait a bit and see what Go Team's response would be and decide which way I want to work around this :)

nevkontakte:
For those who might be following this PR, a quick update. For a long time I was stuck on mysterious import errors, which turned out to be another bug in go/packages: https://github.com/golang/go/issues/35331#issuecomment-557818093. Luckly, the fix seems to be simple enough, so I'm back to porting stdlib augmentation logic.

nevkontakte:
Another quick update. I have not abandoned this effort, but I was stuck for a while on implementing standard library augmentation. I think I solved the most difficult part and am making forward progress again :)

nevkontakte:
A somewhat late update: I was able to make go/packages work, but it required a number of inelegant hacks and did not actually reduce complexity nearly as much as I thought it would, so I am no longer pursuing this PR. I might try to retrofit some fragments of refactoring I did as a part or it into the existing code, but it will depend on whether a viable fork of GopherJS emerges.

For the posterity, the main difficulty I encountered was that `go/packages` is really designed to suit needs of developer tooling and editors, rather than a third-party compiler. In particular, GopherJS augments many stdlib packages on the fly to make them work in the browser environment. And even though go/packages provides some hooks for transforming/patching AST on the fly, it doesn't provide enough context about the package for GopherJS to do its thing. I did manage to do it with a bunch of hacks and heuristics, but they are fragile and complex, which undermined the original goal of this PR.

PR #950, opened 2019-11-19
Close file

CallanTaylor:


tests/run.go
```diff
@@ -340,6 +340,7 @@ func goFiles(dir string) []string {
 		}
 	}
 	sort.Strings(names)
+	f.Close()
```
dmitshur: `f` is never used after the `f.Readdirnames` call, so this line can be moved to be immediately after it:

```Go
...
dirnames, err := f.Readdirnames(-1)
f.Close()
check(err)
...
```

That will be cleaner and more consistent with code in `ioutil.ReadDir` that does a similar task (see [here](https://github.com/golang/go/blob/cfe2ab42e764d2eea3a3339aac1eaff97520baa0/src/io/ioutil/ioutil.go#L99)).

PR #956, opened 2020-01-21
fixes blocking when using function literals

### LINKED ISSUE #955: compiler/analysis/info.go function literal call site not recorded for later marking
The function analysis.AnalyzePkg doesn't get the blocking right when a function literal calls another function that only later is determined to require blocking. The call site of the function literal is not retained for later handling.

A short main.go program that demonstrates the problem.
```go
package main

import "runtime"

func main() {
	foo()
	println("main done")
}

// Problem: foo is not recognized as a blocking function.
//
// TLDR; function literals aren't handled identically to defined functions
// in analysis/info.go and a call stack is not kept for later use.
//
// Short cause analysis: the function literal called in foo is not determined
// to be blocking until later in analysis/info.go when it is determined that bar
// is blocking.  The FuncLit itself is recognized as blocking later on once bar
// has been analyzed, but not foo: while foo was being analyzed the call site
// stack within it was not saved for later use. So the call site effectively
// falls through and is left as unblocking, so foo is left as unblocking, so
// main is left as unblocking.
func foo() {
	func() { bar() }()
	println("foo done")
}

func bar() {
	runtime.Gosched()
}
```

FrankReh:
When a function literal is used that calls another function and
that other function has yet to be analyzed, the call site for
the function literal is not retained, so it's not processed later,
so the function that was calling the function literal is not made
blocking itself, when it should be.

A new field is used in info.go. It's name kept short to keep gofmt
from readjusting surrounding lines. It's not the most descriptive name
as it stands.

A short test case that demonstrates the problem can be found in
issue #955.

Fixes #955.

### DIFF
--- a/compiler/analysis/info.go
+++ b/compiler/analysis/info.go
@@ -32,6 +32,7 @@ type FuncInfo struct {
 	Blocking      map[ast.Node]bool
 	GotoLabel     map[*types.Label]bool
 	LocalCalls    map[*types.Func][][]ast.Node
+	lFLitCalls    map[*FuncInfo][][]ast.Node
 	ContinueStmts []continueStmt
 	p             *Info
 	analyzeStack  []ast.Node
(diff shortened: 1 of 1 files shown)


nevkontakte:
Hi @FrankReh, we are finally working through the backlog of issues and pull requests. Sorry it took so long!

Can you please rebase this PR on the latest version of master? Also it would be nice to add your test case `tests` package to make sure a regression doesn't happen. https://github.com/gopherjs/gopherjs/blob/master/tests/misc_test.go would probably be a good place for it.

nevkontakte:
@FrankReh I went ahead and rebased this change myself as https://github.com/gopherjs/gopherjs/pull/1115. Thank you for your contribution and providing a minimal reproduction example. It made investigating and fixing the bug a lot easier for me. I will close this PR in favor of #1115.

PR #960, opened 2020-01-26
fix check int overflows error

visualfc:
fix check int overflows error

example:
```
func main() {
	js.Global().Get("document").Call("write", 1<<40)
}
```
gopherjs build output:
`main.go:8:44: 1 << 40 (untyped int constant 1099511627776) overflows int`


nevkontakte:
GopherJS simulates 32 bit environment, so this error seems to be working as intended.

Issue #894, opened 2019-01-29
Health and future of the GopherJS open source project


dmitshur:
I'd like to open this tracking issue to open a discussion about the future of GopherJS, and the health of the project. The goal here is to try to answer the question: what does the future and roadmap look like for GopherJS? This is something that affects both the users and contributors to the project.

## Project Health

Right now, the latest GopherJS 1.11-2 supports Go 1.11.x and works on macOS and Linux. On Windows or other OSes, it requires setting GOOS to one of two supported values, as documented at https://github.com/gopherjs/gopherjs#installation-and-usage, and system calls can’t work (which means `gopherjs test` can’t be used on those platforms). It can be described as functional, but with known limitations.

There are many open issues and PRs that haven’t been fully reviewed and merged. In the last few months, there hasn’t been much activity in that regard. This is indicative of declining health of an open source project.

At this time, there are 4 people with push rights to the main GopherJS repository: @neelance, myself, @hajimehoshi, and @myitcv as of recently (issue #799). To land a PR into `master`, it needs to go through code review and get a LGTM from at least one other maintainer. Richard has been quite busy working on Go WebAssembly and elsewhere. I have a full time job now, leaving me with less time to spend on GopherJS than before. Hajime and Paul have other commitments too. Being an open source project, we’re all volunteers working in spare time, so the resources we can dedicate to the project are limited.

From my side, I can say that I’m committed to doing my best to keep GopherJS running in its current state: on macOS and Linux, and compatible with the current Go releases (1.11 right now, 1.12 in a few months, 1.13, 1.14, etc.) for the foreseeable future. I can’t spend much more time on fixing the remaining issues and reviewing incoming PRs, at least nowhere near what I’d like.

I think it would be very helpful for contributors to have a better idea of what kind of PRs they can/should send and if it has a chance of being reviewed.

## Ways To Proceed

We have a few possible strategies to choose from about what to do next. I will enumerate them below, and make a proposal for a suggested plan of action below. However, if there exists a better option that we haven't thought of, your input will be greatly appreciated.

### Option A: Do Nothing

I want to make it clear that this is not a good option. However, this is the default outcome that happens unless we actively work towards one of the other options.

This option is to do nothing and let things keep going as they are. This isn’t great, there are many stale PRs and it’s unclear to contributors if they should spend time creating new PRs, and whether their PR will be properly reviewed and eventually merged or closed.

[...]

PR #966, opened 2020-02-08
support being built and used with newer versions of Go

### LINKED ISSUE #941: support for being built and used with newer version of Go
Go 1.13 is out by now, but GopherJS doesn't support it yet (that work is tracked in issue #929). I'd like to work on a change that can be made in parallel, unrelated to Go 1.13 support, but it should help with the current situation and buy us more time.

The change can be briefly described as "to allow GopherJS 1.12-n to operate alongside with arbitrary Go distribution versions, rather than hard-requiring it to be Go 1.12.x".

/cc @neelance @hajimehoshi @myitcv

## Background

At this time, it's only possible to build and use GopherJS with one major version of Go at a time. Currently, GopherJS 1.12-n requires Go 1.12.x. This is enforced by a build-time check implemented via build constraints. See:

https://github.com/gopherjs/gopherjs/blob/d3ddacdb130fcd23f77a827e3b599804730be6b5/compiler/version_check.go#L1-L2

The check is necessary because the GopherJS requires a 1.12.x Go distribution. It needs it to be able to read the source code for all of the standard library packages (i.e., the .go files in GOROOT/src), which GopherJS needs to compile.

However, the check doesn't have to be done at compile-time. We can do it at runtime instead, which would enable doing things like:

```
$ go version
go version go1.13 darwin/amd64
$ go get -u github.com/gopherjs/gopherjs

$ gopherjs build some.example/project
GopherJS 1.12-2 requires a Go 1.12.x distribution, but found version go1.13

$ GOROOT=$HOME/sdk/go1.12.9 gopherjs build some.example/project
<success>
```

# Alternatives

### 1. Make it easier to specify a GOROOT value that only GopherJS uses

Let's assume the user wants to have Go 1.13 as their default Go distribution, say in `/usr/local/go`, and a Go 1.12.9 distribution in their `$HOME/sdk/go1.12.9` directory (this is easy by doing `go get golang.org/dl/go1.12.9` and then `go1.12.9 download`).

That means they can either build GopherJS using the Go 1.12.9 distribution so the default `runtime.GOROOT()` value in the `gopherjs` binary is `$HOME/sdk/go1.12.9` (**edit:** based on comments from @inliquid below, this option may not be working correctly today), or explicitly set the `GOROOT` environment variable to `$HOME/sdk/go1.12.9` when running the `gopherjs` binary:

```
GOROOT=$HOME/sdk/go1.11.13 gopherjs build some.example/project
```

However, this has two problems/limitations:

1. Setting `GOROOT` manually for each `gopherjs` invocation is annoying and easy to forget, or accidentally set wrong `GOROOT` for main Go distribution

2. Correct GopherJS-specific `GOROOT` will also need to be set when using GopherJS to compile in-memory (e.g., with [`github.com/shurcooL/gopherjslib`](https://godoc.org/github.com/shurcooL/gopherjslib) or [`github.com/shurcooL/go/gopherjs_http`](https://godoc.org/github.com/shurcooL/go/gopherjs_http) packages)

To avoid both these problems, I propose adding a new GopherJS-specific environment variable:

```
GOPHERJS_GOROOT - if set, GopherJS uses this value as the default GOROOT value,
                  instead of using the system GOROOT as the default GOROOT value
```

The idea would be that if you want to use GopherJS with a mis-matched version of the Go distribution, you permanently set `GOPHERJS_GOROOT` to point to a matching Go distribution version and forget about it (until next major GopherJS version). It won't interfere with your normal Go distribution.

### 2. Embed all of the required files from GOROOT/src into GopherJS itself

Similar to how we already have `natives` and `gopherjspkg` file systems, we can have a third one: a copy of the source code of Go standard library packages (and potentially their tests).

I've explored this idea. It is certainly possible, but the concern is about the size of the generated assets_vfsdata.go file because the Go standard library is quite sizable.

I took GOROOT/src from Go 1.12.9 and trimmed it as follows:

- deleted all of cmd/ directory, GopherJS doesn't need it
- deleted all non .go files (there are many cgo files, assembly files, etc., GopherJS doesn't need them)
- deleted all testdata directories
- deleted all _test.go files (they were pretty small, didn't make a big difference)

The end result size of the src directory was about 15.3 MB. Compressed with gzip it becomes just 4.8 MB, which is pretty small. vfsgen uses gzip compression but encodes bytes in Go byte slice literals, which have an efficiency of 1:4, so the assets_vfsdata.go file ends up being 18.8 MB. If we're checking that into the git repository, it's not small, and it'd need to change whenever we add support for a new Go major version (or minor version).

The advantage of this approach is that it'd be very user friendly: users just `go get -u github.com/gopherjs/gopherjs` on any version of Go and it just works, no mucking around with environment variables.

The disadvantage is the large assets_vfsdata.go file, and complexity when it comes to testing GopherJS (we need the _test.go files and testdata directories to test GopherJS on std lib packages).

I think this is something we should explore more in the future, I don't think we should implement it now.

## Proposal

For now, I'm leaning towards implementing alternative 1: moving the Go distribution version check from build-time to run-time, and adding support for the `GOPHERJS_GOROOT` environment variable. It would have to be documented, and the error messages should mention it.

Thoughts?

dmitshur:
This change removes the build-time check that GopherJS was built
with Go version 1.12, and replaces it with a runtime check instead.

Add and document a new GopherJS-specific environment variable:

```
GOPHERJS_GOROOT - if set, GopherJS uses this value as the default GOROOT value,
                  instead of using the system GOROOT as the default GOROOT value
```

The main reason that GopherJS required Go 1.12 to be built in the
past was as a way of ensuring that the Go 1.12 standard library
code is available on disk, which GopherJS 1.12-2 needs to build
Go code. Now that the Go distribution can provided to GopherJS
with convenience via the GOPHERJS_GOROOT environment variable,
we can allow GopherJS to be built with a newer version of Go,
as long as a Go 1.12 distribution is provided via GOPHERJS_GOROOT.

Update GopherJS installation instructions to support being installed
with Go 1.12 and newer versions of Go.

The Go distribution version check is added to build.NewSession.
That check may fail, and so it was necessary to add an error return
value to the signature of build.NewSession. The build API is deemed
semi-internal, so a breaking API change is unfortunate but acceptable.

Regenerate ./compiler/natives with:

	go generate ./...

Fixes #941


dmitshur:
/cc @hajimehoshi Feel free take a look if you have time (and it's okay if not, due to #894).

If there aren't major problems with this, I plan to merge this later today, and follow up if any problems are discovered in the near future.

I've been using the less polished pre-force-push version of this change locally for many months now and didn't spot any problems.

compiler/version_check.go
```diff
@@ -1,9 +1,25 @@
...
+
+// CheckGoVersion checks the version of the Go distribution
+// at goroot, and reports an error if it's not compatible
+// with this version of the GopherJS compiler.
+func CheckGoVersion(goroot string) error {
+	v, err := ioutil.ReadFile(filepath.Join(goroot, "VERSION"))
+	if err != nil {
+		return fmt.Errorf("GopherJS %s requires a Go 1.12.x distribution, but failed to read its VERSION file: %v", Version, err)
+	}
+	if !bytes.Equal(v, []byte("go1.12")) && !bytes.HasPrefix(v, []byte("go1.12.")) {
```
hajimehoshi: I'd want `strings.TrimSpace` for `v` just in case.

This check would fail with e.g., `go1.12rc1`.

dmitshur: Ok, I'll make it just check for a "go1.12" prefix. The code will change before we need to worry about it matching Go 1.120 falsely.

hajimehoshi: `/^go1.12([.a-z].+)?/` maybe?

dmitshur: Done. I think the prefix check should be sufficient for now.

hajimehoshi: As this string comes from a file, I'd want `strings.TrimSpace`.

hajimehoshi: On the second thought, `HasPrefix` would be enough to ignore the tailing new line char.

dmitshur: Thanks for suggesting `strings.TrimSpace`. I've considered it, but I don't think it's needed. In a normal Go installation, there are never leading spaces.

tests/run.go
```diff
@@ -311,14 +313,6 @@ func main() {
 	}
 }
 
-func toolPath(name string) string {
```
hajimehoshi: My understanding is that this function was not called from anywhere. Is that correct?

dmitshur: Yep, it was unused, so I deleted it.

It's unused in the [upstream `test/run.go`](https://github.com/golang/go/blob/cfe2ab42e764d2eea3a3339aac1eaff97520baa0/test/run.go) file too. I plan to send a CL to remove it from there too after the tree opens up.

dmitshur:
Thanks for reviewing.

I'll be watching the issue tracker and follow up if anything comes up.

dmitshur:
I'll bump the (minor) version of GopherJS and advertise this change on Twitter.

Issue #969, opened 2020-02-15
ReleaseTags of default build context is inherited from Go version used to build GopherJS, which may be newer than 1.12


flimzy:
Running `gopherjs` with the new `GOPHERJS_GOROOT` option sets build tags according to the running version of Go, not the compatible version of the standard library.

To reproduce, run GopherJS 1.12 with Go 1.13 (or newer, presumably):

```shell
$ go version
go version go1.13 linux/amd64
$ gopherjs version
GopherJS 1.12-3
```

and create a simple `main.go` as so:

```golang
// +build go1.13

package main 

func main() {
}
```

Then attempt to compile it, and notice it works.

```shell
$ gopherjs build .
```

It should fail:

```shell
$ go version
go version go1.12 linux/amd64
$ go build .
can't load package: package .: build constraints exclude all Go files in /home/jonhall/go/src/github.com/flimzy/tes
```

flimzy:
I believe this is the result of compiling `gopherjs` with Go 1.13, and thus pulling in the Go 1.13 version of `go/build`.

Two possible solutions might be:

1. Update the documentation to compile GopherJS against Go 1.12
2. Find some way to vendor the Go 1.12 stdlib needed for compiling GopherJS

dmitshur:
Thanks for reporting this @flimzy. I hadn't anticipated this, and it's helpful to know about it.

I'll need some time to think about what's the best solution.

Can you share if this caused some real-world breakage for you, or did you only realize this may cause problems in theory?

dmitshur:
Ok, here's an example of how having a mismatched ReleaseTags could cause a problem.

Imagine someone has a valid Go package with Go 1.12 and 1.13+ code:

```Go
// +build !go1.13

// doesn't use errors.Unwrap, errors.Is and errors.As, since they're new to 1.13
```

```Go
// +build go1.13

// uses errors.Unwrap, errors.Is and errors.As
```

Then GopherJS 1.12-3 will build that package correctly if it was built with Go 1.12, but fail to do so if it was built with Go 1.13+.

So, I think the fight fix is to hardcode the ReleaseTags of the default GopherJS build context to correspond to the Go standard library version that it supports the most, namely Go 1.12 at this time.

flimzy:
> Can you share if this caused some real-world breakage for you, or did you only realize this may cause problems in theory?

[...]

PR #970, opened 2020-02-15
Build our own list of release tags

flimzy:
This is a possible fix for #969.  It seems like an ugly hack to me, but it solves the immediate problem.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -56,6 +56,14 @@ func (e *ImportCError) Error() string {
 // are loaded from gopherjspkg.FS virtual filesystem rather than GOPATH.
 func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
 	gopherjsRoot := filepath.Join(DefaultGOROOT, "src", "github.com", "gopherjs", "gopherjs")
+	var releaseTags []string
+
+	// Modification of https://golang.org/src/go/build/build.go#L303, to exclude
+	// newer versions of Go.
+	for i := 1; i <= 12; i++ {
+		releaseTags = append(releaseTags, "go1."+strconv.Itoa(i))
+	}
+
 	return &build.Context{
 		GOROOT:        DefaultGOROOT,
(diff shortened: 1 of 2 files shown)


dmitshur:
This may or may not be the right change depending on what we decide the ideal behavior should be. Let's discuss in issue #969 first.

build/build.go
```diff
@@ -67,7 +75,7 @@ func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
 			"netgo",  // See https://godoc.org/net#hdr-Name_Resolution.
 			"purego", // See https://golang.org/issues/23172.
 		),
-		ReleaseTags: build.Default.ReleaseTags,
+		ReleaseTags: releaseTags,
```
dmitshur: I'd like to suggest we simplify and reduce allocations by reusing `build.Default.ReleaseTags`. On Go 1.12+ and newer, it's guaranteed to have exactly the values we want, there is just more of them on 1.13 and newer. So, something like:

```
ReleaseTags: build.Default.ReleaseTags[:12],
```

But I think it may be a good idea to move the 12 to be a constant in `compiler/version_check.go`, so all numbers that need to change when adding support for a newer Go standard library are together in one place. In that case:

```
ReleaseTags: build.Default.ReleaseTags[:compiler.GoMajorVersion],
```

And we should re-add a `// +build go1.12` constraint to `version_check.go`, because attempting to build GopherJS with 1.11 or older is not supported and we wan't want to try to catch all the possible problems that it might cause.

@flimzy Are you okay with making these changes?

flimzy: @dmitshur: I believe I've made all suggested changes in this PR now. Let me know if you'd like to see anything else, and if you'd like me to squash commits.

dmitshur: Everything looks good, thank you. No need to squash, we'll do that as part of merging this PR.

PR #971, opened 2020-02-20
add Go modules support

visualfc:
Add Go modules build/run support and test. ( project use `go env GOMOD` lookup go.mod)
add fastmod for check Go modules and lookup package from go.mod. 
The fastmod is read-only for go.mod, please 'go mod tidy' first if need.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -550,11 +550,6 @@ func (s *Session) checkMod(pkg *PackageData) (err error) {
 	return nil
 }
 
-func (s *Session) CheckModFromDir(dir string) (err error) {
-	s.mod, err = fastmod.LoadPackage(dir, s.bctx)
-	return
-}
-
 // BuildContext returns the session's build context.
 func (s *Session) BuildContext() *build.Context { return s.bctx }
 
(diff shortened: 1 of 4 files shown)


build/build.go
```diff
@@ -525,6 +538,18 @@ func NewSession(options *Options) (*Session, error) {
...
+	if pkg.Goroot {
+		s.mod = nil
+	} else {
+		s.mod, _ = fastmod.LoadPackage(pkg.Dir, s.bctx)
+	}
+}
+
+func (s *Session) CheckModFromDir(dir string) {
+	s.mod, _ = fastmod.LoadPackage(dir, s.bctx)
+}
```
nevkontakte: I rather like this PR, but these two functions make the code rely on side-effects, which is quite error prone and brittle. Perhaps module info should be added to PackageData?

Note: I'm not a GopherJS maintainer, so my words carry limited weight :)

visualfc: https://github.com/gopherjs/gopherjs/pull/971/commits/246fd7500e264213537cdcd26a2cbde0b61ded59 
simplify mod check and clear code.

PR #978, opened 2020-05-26
- convert thrown errors to go errors

pjebs:
You cans see that having a number at the end of the function is idiomatic.

https://golang.org/pkg/reflect/#Value.Slice3

### DIFF
--- a/js/js.go
+++ b/js/js.go
@@ -49,7 +49,7 @@ func (o *Object) SetIndex(i int, value interface{}) { o.object.SetIndex(i, value
 // Call calls the object's method with the given name.
 func (o *Object) Call(name string, args ...interface{}) *Object { return o.object.Call(name, args...) }
 
-// Call calls the object's method with the given name. If object's method throws a Javascript
+// Call2 calls the object's method with the given name. If object's method throws a Javascript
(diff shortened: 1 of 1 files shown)


pjebs:
@dmitshur @flimzy Why is the unit test failing?

pjebs:
@neelance

pjebs:
It's not about how easy it is to create a workaround function.
If we're programming JS applications in Go, we want JS functions to behave and look like Go code. That means Exceptions need to be converted to error return values.

flimzy:
I think I agree with @nevkontakte. Making a new function version for every function that might panic will be very noisy, and cumbersome to use.

My preference is for a helper function that wraps a function that may panic, and returns an error.  I think that is more maintainable, and also may look a bit more like a JS Try/Catch block, which is pretty close to what it is.  Maybe we even call it `Try()`?

pjebs:
I can do that

flimzy:
Based on the last comment, it seems that an alternative approach was agreed on.  Further, this PR has now sat idle for about a year and a half, so I think it's fair to close it, and apply the alternate approach in a new PR.  If anyone disagrees, we can of course re-open it.

Issue #989, opened 2021-02-21
Go 1.16 standard library support


nevkontakte:
Go 1.16 has been [released](https://blog.golang.org/go1.16) just a few days ago, and the latest Go version that GopherJS supported was 1.12, so we need to catch up. Note that this issue covers only the standard library and language support, Go Modules are tracked in https://github.com/gopherjs/gopherjs/issues/855.

After speaking with @flimzy and other folks we've decided that it's best to jump straight Go 1.16 version, without providing separate releases for intermediate versions. The major reasons behind this choice are:

 1. Making multiple additional releases is a lot of work that we really can't afford now.
 2. If we support Go 1.16 we should still be able to build Go code for earlier versions without any significant issues thanks to the Go backward compatibility policy.
 3. The value of these releases will also quickly diminish as the time passes.

In terms of how to proceed we have two options:

 1. Pull in changes from @visualfc's fork (https://github.com/goplusjs/gopherjs), which already pretty much supports everything we need. However, I would like to get @visualfc's permission before we do that to make sure there won't be any bad feelings left between the two projects.
     - As a side note, this won't be a very straight-forward pull since @visualfc's fork uses different import paths and whatnot, but still it is much, much easier.
 2. As a plan B, we could also reimplement everything ourselves. This will be a lot of work and create more divergence between the forks, but is still doable.

visualfc:
> 1. Pull in changes from @visualfc's fork (https://github.com/goplusjs/gopherjs), which already pretty much supports everything we need. However, I would like to get @visualfc's permission before we do that to make sure there won't be any bad feelings left between the two projects.
>    
>    * As a side note, this won't be a very straight-forward pull since @visualfc's fork uses different import paths and whatnot, but still it is much, much easier.

use https://github.com/goplusjs/gopherjs any source code no problem.

https://github.com/goplusjs/gopherjs natives use tags for easy support high version. ( It's easier than you think )
```
compiler/natives/src/os
- os.go
- go116_os.go    // +build go1.16
```

nevkontakte:
@visualfc thanks for the permission, I appreciate that a lot.

Supporting multiple older versions of the library is nice, but it also has a cost that grows over time. If a certain library function changes its behavior between two versions, we'd have to keep two copies of it. The more versions of the standard library we support going back in time, the more redundant code we'll have to carry with us, maintain it, and fix bugs in.

[...]

PR #990, opened 2021-02-28
Support compiling Go 1.16 `runtime` with GopherJS.

nevkontakte:
The vast majority of the code (even in errors.go) is not used by
GopherJS, so starting with this commit the package is entirely defined
by our own implementation in natives. As a bonus, this allowed to drop
a few unexported symbols that existed purely to let other unused
functions to compile.

Updates #989


compiler/natives/src/runtime/runtime.go
```diff
@@ -12,19 +12,23 @@ const GOOS = sys.GOOS
...
+	RuntimeError()
 }
-func (t *_type) pkgpath() string {
-	return ""
+
+// A TypeAssertionError explains a failed type assertion.
+type TypeAssertionError struct{}
+
+func (*TypeAssertionError) RuntimeError() {}
+func (e *TypeAssertionError) Error() string {
```
flimzy, line 30:
```suggestion
func (*TypeAssertionError) Error() string {
```

compiler/natives/src/runtime/runtime.go
```diff
@@ -88,27 +90,22 @@ type Frame struct {
...
 
 func Goexit() {
 	js.Global.Get("$curGoroutine").Set("exit", true)
 	js.Global.Call("$throw", nil)
 }
 
-func GOMAXPROCS(n int) int {
-	return 1
-}
+func GOMAXPROCS(n int) int { return 1 }
```
flimzy, line 100:
```suggestion
func GOMAXPROCS(int) int { return 1 }
```

flimzy (approved):
LGTM. A couple argument variables can optionally be omitted, if we wish, but totally unimportant.

PR #994, opened 2021-03-11
Add `internal/reflectlite` package support.

nevkontakte:
This package is meant to be a cut-down version of the `reflect` package,
which is used by packages like `errors` to avoid a cyclic dependency
with `reflect`.

This code is based on goplusjs/gopherjs@8e608e5d486b4e9f5142c6c6822649856f936421.

The natives code of the reflectlite package is a subset of its
counterpart from reflect.  Unfortunately, this requires us to duplicate
a substantial chunk of code with all the maintenance problems that stem
from it. I have considered two alternative approaches, none of which
worked out:

  - Automatically rewrite imports from `internal/reflectlite` to
    `reflect`. Unfortunately, this creates cyclic dependencies and
    causes compilation to fail.
  - Redefine types in reflectlite as aliases of their reflect
    counterparts. This both causes a cyclic dependency and creates a
    divergence between GopherJS and Go standard library. TinyGo seems to
    be doing that somehow, but I suspect that they are able to handle
    cyclic dependencies.

As a bottom line, duplicating a subset of reflect code is the lesser
evil that is available to us. In the long term we could try and converge
our reflect implementation with the upstream, so that this is no longer
our problem.

Updates #989.

Co-authored-by: visualfc <visualfc@gmail.com>


compiler/natives/src/internal/reflectlite/export_test.go
```diff
@@ -0,0 +1,34 @@
...
+func TField(typ Type, i int) Type {
+	t := typ.(*rtype)
+	if t.Kind() != Struct {
+		panic("reflect: Field of non-struct type")
+	}
+	tt := (*structType)(unsafe.Pointer(t))
+	return StructFieldType(tt, i)
+}
+
+// Field returns the i'th struct field.
```
flimzy, line 27: Not sure, maybe?

```suggestion
// Field returns the type of the i'th struct field.
```

compiler/natives/src/internal/reflectlite/reflectlite.go
```diff
@@ -0,0 +1,973 @@
...
+	}
+	if len > cap {
+		panic("reflect.MakeSlice: len > cap")
+	}
+
+	return makeValue(typ, js.Global.Call("$makeSlice", jsType(typ), len, cap, js.InternalObject(func() *js.Object { return jsType(typ.Elem()).Call("zero") })), 0)
+}
+
+func TypeOf(i interface{}) Type {
+	if !initialized { // avoid error of uint8Type
```
flimzy, line 346: When does this happen?

compiler/natives/src/internal/reflectlite/reflectlite.go
```diff
@@ -0,0 +1,973 @@
...
+	used(interfaceType{})
+	used(mapType{})
+	used(ptrType{})
+	used(sliceType{})
+	used(structType{})
+	used(imethod{})
+	used(structField{})
+
+	initialized = true
+	uint8Type = TypeOf(uint8(0)).(*rtype) // set for real
```
flimzy, line 31: Is there magic happening that I'm not seeing, that requirese this to happen in `init()`?

compiler/natives/src/internal/reflectlite/reflectlite.go
```diff
@@ -0,0 +1,973 @@
...
+				methods: imethods,
+			})
+		case Map:
+			setKindType(rt, &mapType{
+				key:  reflectType(typ.Get("key")),
+				elem: reflectType(typ.Get("elem")),
+			})
+		case Ptr:
+			// p := (*ptrType)(unsafe.Pointer(rt))
+			// p.elem = reflectType(typ.Get("elem"))
```
flimzy, line 161: Dead code? Or TODO?

flimzy (approved):
Admitedly, a very high-level review.  I didn't read every function. I expect tests to tell us about correctness, more than review.

compiler/natives/src/internal/reflectlite/reflectlite.go
nevkontakte: Unfortunately, I do not know. This is basically a verbatim citation of the code in `reflect` natives, and it seems to work (judging by the fact that it passes tests).

This code is 7 years old and I fully expect that it could be improved significantly. All that "avoid dead code elimination" magic, for example, is highly suspicious to me. According to my understanding of the GopherJS compiler it should not be necessary, but it exists for some reason. That said, I help myself back from making any substantial refactorings right now, since it is a risky endeavor and I didn't want to mix it with the Go 1.16 support work.

compiler/natives/src/internal/reflectlite/reflectlite.go
nevkontakte: Same as above, this code was cargo-culted over from `reflect` and it seems to work, so I opted against touching it for now. Once GopherJS is fully up to date with Go 1.16 and is stable we could could try and experiment with removing potentially load-bearing artifacts :)

compiler/natives/src/internal/reflectlite/reflectlite.go
nevkontakte: Yep, it's a dead code I missed when adopting changes from @visualfc's fork. Removed.

nevkontakte:
Unfortunately, the inner workings of the GopherJS's reflection are not really documented and it is a fairly complex piece of code. Spending some time to document it and perhaps simplify would be a valuable contribution for sure…

Issue #993, opened 2021-03-11
node-syscall doesn't support struct-pointer arguments


nevkontakte:
I ran into this issue today while debugging `testing` package under Go 1.16. 

The `Fstatat()` function in `https://pkg.go.dev/internal/syscall/unix#Fstatat` package passes a pointer to `struct stat` as a third argument. Unfortunately, our node-syscall module doesn't seem to support passing structs into syscalls, only arrays or integers. I don't know if there is an easy fix, or how many syscalls like that there is, so I'm filing this issue primarily for reference purposes.

If we were to implement this, we need to be able to convert a Go struct baked by a JS object into a binary representation, pass that into the syscall, and then unpack it again. I am not sure if we currently have all necessary information at runtime to achieve that.

PR #995, opened 2021-03-12
Support `testing` package with Go 1.16 and re-enable most of its tests.

nevkontakte:
There were two major issues with the package:

  - os.RemoveAll() implementation on linux was using Fstatat syscall,
    which doesn't work with the node-syscall extension (see #993).
  - runtime.Callers() is not implemented by GopherJS, which causes
    testing package to report source code locations according to the JS
    output file rather than the original Go code. I had to skip a bunch
    of tests because of that, but at least most other tests can be
    tracked now.

Updates #989.

### DIFF
--- a/compiler/natives/src/testing/sub_test.go
+++ b/compiler/natives/src/testing/sub_test.go
@@ -1,0 +1,21 @@
+// +build js
+
+package testing
+
+func TestBenchmarkReadMemStatsBeforeFirstRun(t *T) {
+	t.Skip("runtime.ReadMemStats() is not supported by GopherJS.")
+}
+
+func TestTRun(t *T) {
+	// TODO(nevkontakte): This test performs string comparisons expecting to find
+	// sub_test.go in the output, but GopherJS currently reports caller
+	// locations as "test.<random_number>" due to minimal caller and source map support.
+	t.Skip("GopherJS doesn't support source maps sufficiently.")
+}
+
+func TestBRun(t *T) {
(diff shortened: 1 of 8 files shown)


compiler/natives/src/testing/sub_test.go
```diff
@@ -0,0 +1,19 @@
+package testing
+
+func TestBenchmarkReadMemStatsBeforeFirstRun(t *T) {
+	t.Skip("runtime.ReadMemStats() is not supported by GopherJS.")
+}
+
+func TestTRun(t *T) {
+	// TODO(nevkontakte): This test performs string comparisons expecting to find
+	// sub_test.go in the output, but GopherJS currently currently reports caller
```
flimzy, line 9:
```suggestion
	// sub_test.go in the output, but GopherJS currently reports caller
```

compiler/natives/src/testing/sub_test.go
```diff
@@ -0,0 +1,19 @@
...
+	// locations as "test.<random_number>" due to minimal caller and source map support.
+	t.Skip("GopherJS doesn't support source maps sufficiently.")
+}
+
+func TestBRun(t *T) {
+	// TODO(nevkontakte): This test performs string comparisons expecting to find
+	// sub_test.go in the output, but GopherJS currently currently reports caller
```
flimzy, line 16:
```suggestion
	// sub_test.go in the output, but GopherJS currently reports caller
```

flimzy (approved):

PR #992, opened 2021-03-08
Go1.16 support for several more stdlib packages

nevkontakte:
I'm slowly working towards making `gopherjs test` run. With this PR it is able to compile tests, but it fails at runtime because I haven't updated support reflectlite yet.

Packages covered in this PR:

 - math
 - sync
 - time
 - os
 - internal/syscall/unix

Issue #989.

PR #996, opened 2021-03-13
Support Go 1.16 `reflect` package.

nevkontakte:
This pull request addresses multiple issues that were causing test failures for the `reflect` package.

  - Better support for determining caller function name and files.
  - Stricter type checks in the `MakeFunc()` function.
  - Working implementation of the `StructOf()` function.
  - Fixed minor a divergence with the upstream Go in type signature string generation.
  - A few minor internal changes to match changes in the upstream.

See individual commit messages for more details.

Updates #992.


compiler/natives/src/reflect/reflect_test.go
```diff
@@ -59,40 +59,20 @@ func TestSelectOnInvalid(t *testing.T) {
...
-func TestStructOfTooManyFields(t *testing.T) {
-	t.Skip("StructOf")
+	// TODO(nevkontakte) Most of this test actually passes, but there is something
+	// about embedding fields with methods that can or can't be stored in an
+	// interface value directly that GopherJS does differently from upstream. As
+	// a result, GopherJS's implementation of StructOf() doesn't panic where
+	// upstream does. It seems to be a result of our implementation not propagating
+	// the kindDirectIface flag in struct types created by StructOf(), but at this
+	// point I wasn't able to figure out what that flag actually means in the
+	// GopherJS context or how it maps onto our own reflection implementation.
```
flimzy, line 74: Are these types of TODOs being documented somewhere, other than these comments? Should they be? Maybe a GH issue?

compiler/natives/src/runtime/runtime.go
```diff
@@ -60,13 +60,52 @@ func GOROOT() string {
 
 func Breakpoint() { js.Debugger() }
 
+var (
+	// JavaScript runtime doesn't provide access to low-level execution position
+	// conters, so we emulate them by recording positions we've encountered in
```
flimzy, line 65:
```suggestion
	// counters, so we emulate them by recording positions we've encountered in
```

?

compiler/natives/src/runtime/runtime.go
```diff
@@ -172,15 +211,39 @@ func SetFinalizer(x, f interface{}) {
...
+		return "<unknown>"
+	}
+	return f.name
+}
 
 func FuncForPC(pc uintptr) *Func {
-	return nil
+	ipc := int(pc)
+	if ipc >= len(positionCounters) {
+		// Since we are faking position counters, the only valida way to obtain one
```
flimzy, line 239:
```suggestion
		// Since we are faking position counters, the only valid way to obtain one
```

compiler/natives/src/runtime/runtime.go
```diff
@@ -172,15 +211,39 @@ func SetFinalizer(x, f interface{}) {
...
+		// is through a Caller() or Callers() function. If pc is out of positionCounters
+		// bounds it must have been obtained in some other way, which is unexpected.
+		// If a panic proves problematic, we can return a nil *Func, which will
+		// present itself as a generic "unknown" function.
+		panic("GopherJS: pc=" + itoa(ipc) + " is out of range of known position counters")
```
flimzy, line 244: If my reading of upstream is correct, panicking is what they do, too, so I expect this is fine.

https://golang.org/src/runtime/symtab.go?s=18483:18515#L698

flimzy (approved):

compiler/natives/src/reflect/reflect_test.go
nevkontakte: I have thought about this a bit, but did not come to a definitive conclusion. My thoughts so far:

 - Documenting this kind of problems in the code as a comment/todo makes sense because that's where people would be looking when debugging some kind of unexpected behavior. So a TODO is generally better discoverable.
 - Making such comments with a TODO makes it easy to find them with a simple grep command. Adding a username also points to a person who might be able to provide more context for the comment if the need arises.
 - In this particular instance the problem is fuzzy enough and seemingly unimpactful, so a GitHub issue is likely to linger unfixed ~forever. This will decrease signal/noise ratio and make issues harder to manager.
 
At least for now I lean towards filing issues for things that a definitely worth fixing or likely to impact users, but I can be convinced otherwise :)

compiler/natives/src/runtime/runtime.go
nevkontakte: I think this is different. In the upstream code they panic if there is no entry in some kind of data table, which I think is compiler-generated. In our case we panic if we try to look up info about a position that wasn't returned by `runtime.Caller()`.

compiler/natives/src/reflect/reflect_test.go
```diff
@@ -187,3 +167,18 @@ func init() {
 	// TODO: This is a failure in 1.11, try to determine the cause and fix.
 	typeTests = append(typeTests[:31], typeTests[32:]...) // skip test case #31
 }
+
+func TestConvertNaNs(t *testing.T) {
+	// This test is exactly the same as the upstream, except it uses a "quiet NaN"
+	// value instead of "signalling NaN". JavaScript apepars to coerce all NaNs
```
flimzy, line 173:
```suggestion
	// value instead of "signalling NaN". JavaScript appears to coerce all NaNs
```

flimzy (approved):
don't worry about comment typo corrections unless you're really bothered, or are making other changes anyway. It's still readable :)

compiler/natives/src/reflect/reflect_test.go
nevkontakte: Quite the opposite, I appreciate them a lot! For some reason my brain is incapable of proper spell checking on myself, so your corrections are very welcome. Also, having typos checked in forever is a bit embarrassing, so correcting them at review time is preferable 🙂

compiler/natives/src/reflect/reflect_test.go
```diff
@@ -187,3 +167,18 @@ func init() {
...
+	// value instead of "signalling NaN". JavaScript appears to coerce all NaNs
+	// into quiet ones, but for the purpose of this test either is fine.
+
+	const qnan uint32 = 0x7fc00001 // Originally: 0x7f800001.
+	type myFloat32 float32
+	x := V(myFloat32(math.Float32frombits(qnan)))
+	y := x.Convert(reflect.TypeOf(float32(0)))
+	z := y.Interface().(float32)
+	if got := math.Float32bits(z); got != qnan {
+		t.Errorf("signaling nan conversion got %x, want %x", got, qnan)
```
flimzy, line 182:
```suggestion
		t.Errorf("quiet nan conversion got %x, want %x", got, qnan)
```

Not sure if this change makes sense?

nevkontakte: Yup, thanks.

flimzy (approved):

PR #1009, opened 2021-03-29
Fix or skip newly added Go compiler tests in 1.16

nevkontakte:
This PR addresses test failures supplied by the Go repository itself. Some of them were testing features GopherJS doesn't support (primarily related to unsafe and memory management), and some were legitimate GopherJS bugs, which I fixed for the most part. See more details in commit messages.

I believe this should be the final PR for Go 1.16.2 standard library support. Phew.


compiler/version_check.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 // Version is the GopherJS compiler version string.
-const Version = "1.16-dev"
+const Version = "1.16.0+go1.16.2"
```
flimzy, line 13: Do we want to target the upstream patch version, or just the minor version?  I think in the past, GopherJS always targeted only a minor version. That is, GopherJS 1.12 was expected to work with any patch of Go 1.12.x.  But in theory there could be a patch version that breaks GopherJS, so maybe it's safer to target the patch version?  Does targeting a specific patch version actually do anything (other than advertise support for that patch)?

tests/run.go
```diff
@@ -127,6 +125,20 @@ var knownFails = map[string]failReason{
 	"fixedbugs/issue30977.go": {category: neverTerminates, desc: "does for { runtime.GC() }"},
 	"fixedbugs/issue32477.go": {category: notApplicable, desc: "uses runtime.SetFinalizer and runtime.GC"},
 	"fixedbugs/issue32680.go": {category: notApplicable, desc: "uses -gcflags=-d=ssa/check/on flag"},
+
+	// These are new tests in Go 1.13-1.16.
+	"fixedbugs/issue19113.go":  {category: lowLevelRuntimeDifference, desc: "JavaScript bit shifts by negative amount don't cause an exception"},
```
flimzy, line 130: Is this sort of thing worth mentioning in the compat table?

flimzy (approved):

compiler/version_check.go
nevkontakte: My understanding is that this constant reflects GopherJS version. I think it's a useful signal to show the latest Go version it was tested against, since incompatible changes in standard library internals may happen with patch releases. No a strong opinion though.

tests/run.go
nevkontakte: Yeah, this is a good idea, although I am again torn about where to put that note. docs/packages.md is supposed to be about packages and as a user I wouldn't expect to find runtime compatibility information there. On the other hand, creating a third document just for a few paragraphs' sake… Any thoughts?

nevkontakte: Maybe renaming packages.md into compatibility.md and putting all the bits together is a reasonable way to go, although I expect this will break a bunch of links, some of which are external, which is not great.

compiler/version_check.go
flimzy: Works for me.

tests/run.go
nevkontakte: Actually, I think I'll do this: keep packages.md and compiler.md with their narrow-scope purpose and create a more general compatibility.md as a central doc for this kind of thing, linking out to other docs when appropriate. After a bit of thought there's merit in describing the high-level state of things, as well as low-level details like that. Since it'll deserve its own review, I'll do that in a separate PR.

PR #1015, opened 2021-04-06
Add support for Go 1.16.3 standard library

### LINKED ISSUE #989: Go 1.16 standard library support

nevkontakte:
Merging go1.16-stdlib into the master to introduce Go 1.16.3 support into GopherJS 

We are skipping support for Go 1.13-1.15 and jumping straight to 1.16, since this is the current stable. Unfortunately, we don't have resources necessary to make separate releases for all intermediate Go versions that occurred since the previous GopherJS release.

Note: to the best of my knowledge, this should work with any 1.16.x release to date, but 1.16.3 is what we are testing against.

Fixes #989 :partying_face:


flimzy (approved):

paralin:
Awesome work!

PR #1017, opened 2021-04-10
Add go.mod go.sum files

paralin:
Run "go mod init github.com/gopherjs/gopherjs" to create the go.mod and go.sum files.


nevkontakte:
Thanks!

PR #1019, opened 2021-04-10
compiler/natives/src/syscall: support copy bytes to/from js

paralin:
This is an old addition that I've been using for a long time, maybe suitable to include in syscall

### DIFF
--- a/compiler/natives/src/syscall/js/js.go
+++ b/compiler/natives/src/syscall/js/js.go
@@ -281,3 +281,27 @@ func (e *ValueError) Error() string {
 type Wrapper interface {
 	JSValue() Value
 }
+
+// CopyBytesToGo copies bytes from the Uint8Array src to dst.
+// It returns the number of bytes copied, which will be the minimum of the lengths of src and dst.
+// CopyBytesToGo panics if src is not an Uint8Array.
+func CopyBytesToGo(dst []byte, src Value) int {
+	vlen := src.v.Length()
+	if dlen := len(dst); dlen < vlen {
+		vlen = dlen
+	}
+	copy(dst, src.v.Interface().([]byte))
+	return vlen
+}
+
+// CopyBytesToJS copies bytes from src to the Uint8Array dst.
+// It returns the number of bytes copied, which will be the minimum of the lengths of src and dst.
+// CopyBytesToJS panics if dst is not an Uint8Array.
+func CopyBytesToJS(dst Value, src []byte) int {
+	dt, ok := dst.([]byte)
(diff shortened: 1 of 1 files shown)


flimzy:
I'm not completely sure if this is accurate, but I was thinking that a goal of `syscall/js` was to retain compatibility with upstream's WASM support. If that's true, then these types of things may be long in some other library.

But again, I'm not 100% sure if this is a goal for `syscall/js`.

compiler/natives/src/syscall/js/js.go
```diff
@@ -281,3 +281,27 @@ func (e *ValueError) Error() string {
...
+
+// CopyBytesToGo copies bytes from the Uint8Array src to dst.
+// It returns the number of bytes copied, which will be the minimum of the lengths of src and dst.
+// CopyBytesToGo panics if src is not an Uint8Array.
+func CopyBytesToGo(dst []byte, src Value) int {
+	vlen := src.v.Length()
+	if dlen := len(dst); dlen < vlen {
+		vlen = dlen
+	}
+	copy(dst, src.v.Interface().([]byte))
```
nevkontakte, line 293: Could you add a type assertion in the `src` here to be consistent with `CopyBytesToJS`?

nevkontakte:
@flimzy unless I'm missing something, this change improves the actual compatibility with the upstream, no?

https://pkg.go.dev/syscall/js#CopyBytesToGo and https://pkg.go.dev/syscall/js#CopyBytesToJS are a part of the package interface, and AFAICT before this change they'd simply panic due to missing implementation of `copyBytesToJS` and `copyBytesToGo`, which in WASM are implemented as native JS functions.

Arguably, in GopherJS these functions are kind of a no-op, since they copy from one Uint8Array to another, but supporting them would make writing portable code easier for sure.

nevkontakte (approved):

flimzy:
> @flimzy unless I'm missing something, this change improves the actual compatibility with the upstream, no?
> 
> https://pkg.go.dev/syscall/js#CopyBytesToGo and https://pkg.go.dev/syscall/js#CopyBytesToJS are a part of the package interface, and AFAICT before this change they'd simply panic due to missing implementation of `copyBytesToJS` and `copyBytesToGo`, which in WASM are implemented as native JS functions.
> 
> Arguably, in GopherJS these functions are kind of a no-op, since they copy from one Uint8Array to another, but supporting them would make writing portable code easier for sure.

Oh, I missed that detail.  Disregard my old-man ramblings. :see_no_evil:

PR #1020, opened 2021-04-11
Future-proof untyped nil handling.

### LINKED ISSUE #1011: Compiler panic at tip Go due to assumption about nil basic types
Gopherjs breaks at ~[CL 284052](https://golang.org/cl/284052)~ [CL 289715](https://golang.org/cl/289715) in the Go repository, due to an assumption about nil basic types here:
https://github.com/gopherjs/gopherjs/blob/bed99a852dfeb7c8aad4ac454357a198750247e1/compiler/expressions.go#L738
(`t.Kind()` is now `types.UntypedNil`)

This is reproducible as follows:
```
> GO111MODULE=off go test -v -run=TestGoRepositoryCompilerTests ./tests                                                                                         
=== RUN   TestGoRepositoryCompilerTests                                                                             
goos: "linux", goarch: "js"                                                                                         
ok      fixedbugs/bug000.go     0.254s                                                                              
ok      fixedbugs/bug002.go     0.190s                                                                                                                                                                                                  
ok      fixedbugs/bug003.go     0.186s                                                                              
ok      fixedbugs/bug004.go     0.176s                                                                                                                                                                                                  
ok      fixedbugs/bug005.go     0.178s                                                                              
# go run run.go -- fixedbugs/bug006.go                                                                                                                                                                                                  
exit status 2                                                                                                       
panic: unexpected basic type [recovered]                                                                            
        panic: unexpected basic type                                                                                
                                                                                                                    
goroutine 1 [running]:                                                                                              
go/types.(*Checker).handleBailout(0xc00024a000, 0xc0000eed00)                                                       
        /usr/local/google/home/rfindley/src/go2/src/go/types/check.go:245 +0x9f                                   
panic(0x918020, 0xa87510)                                                                                       
        /usr/local/google/home/rfindley/src/go2/src/runtime/panic.go:965 +0x1c7
github.com/gopherjs/gopherjs/compiler.(*funcContext).translateExpr(0xc00052c2c0, 0xa9a310, 0xc00071c1c0, 0xa9a310)
        /usr/local/google/home/rfindley/dev/go/src/github.com/gopherjs/gopherjs/compiler/expressions.go:739 +0xf0d3
```

This appears to be a case of [Hyrum's law](https://www.hyrumslaw.com/) -- I'm not aware of documentation stating that a basic nil must be `types.UnsafePointer`, though please let me know if I've missed something.

Unless there is evidence that this implicit assumption about nil basic types is widespread, we'd like to land this change in `go/types` for go1.17.  It seems like a trivial fix, but I'm not familiar with the gopherjs compiler. Could someone take a look? Thanks.

CC @dmitshur

nevkontakte:
Upstream changes introduced in https://golang.org/cl/284052 introduced several new cases where `nil` identifier would be treated as untyped by `go/types` and an implicit type conversion would be expected. These changes future-proof GopherJS against this particular change.

I've also discovered an additional case, where untyped nil was causing a panic even with Go 1.16: `var _ *int = nil`, I fixed that as well.

I've manually tested these changes against Go `tip`: `go-tip install . && GOPHERJS_NO_VERSION_CHECK=true GOROOT="$(go-tip env GOROOT)" gopherjs test ./tests`.

This PR also includes a couple of changes that make debugging compiler panics and testing with unreleased versions a bit easier.

Fixes #1011.

### DIFF
--- a/compiler/package.go
+++ b/compiler/package.go
@@ -120,7 +120,20 @@ func (pi packageImporter) Import(path string) (*types.Package, error) {
 	return pi.importContext.Packages[a.ImportPath], nil
 }
 
-func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, importContext *ImportContext, minify bool) (*Archive, error) {
+func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, importContext *ImportContext, minify bool) (_ *Archive, err error) {
+	defer func() {
+		e := recover()
+		if e == nil {
+			return
+		}
+		if fe, ok := bailingOut(e); ok {
+			// Orderly bailout, return whatever clues we already have.
+			err = fe
+			return
+		}
+		// Some other unexpected panic, catch the stack trace and return as an error.
+		err = bailout(e)
+	}()
 	typesInfo := &types.Info{
 		Types:      make(map[ast.Expr]types.TypeAndValue),
 		Defs:       make(map[*ast.Ident]types.Object),
--- a/compiler/statements.go
+++ b/compiler/statements.go
@@ -22,6 +22,27 @@ func (fc *funcContext) translateStmtList(stmts []ast.Stmt) {
 }
 
 func (fc *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
+	defer func() {
+		err := recover()
+		if err == nil {
+			return
+		}
+		if _, yes := bailingOut(err); yes {
+			panic(err) // Continue orderly bailout.
+		}
+
+		// Oh noes, we've tried to compile something so bad that compiler paniced
+		// and ran away. Let's gather some debugging clues.
+		bail := bailout(err)
+		pos := stmt.Pos()
+		if fc.posAvailable && fc.pos.IsValid() {
+			pos = fc.pos
+		}
+		fmt.Fprintf(bail, "Occurred while compiling statement at %s:\n", fc.pkgCtx.fileSet.Position(pos))
+		ast.Fprint(bail, fc.pkgCtx.fileSet, stmt, ast.NotNilFilter)
+		panic(bail) // Initiate orderly bailout.
+	}()
+
 	fc.SetPos(stmt.Pos())
 
 	stmt = filter.IncDecStmt(stmt, fc.pkgCtx.Info.Info)
--- a/compiler/statements.go
+++ b/compiler/statements.go
@@ -386,7 +386,7 @@ func (fc *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 		case len(s.Lhs) == 1 && len(s.Rhs) == 1:
 			lhs := astutil.RemoveParens(s.Lhs[0])
 			if isBlank(lhs) {
-				fc.Printf("$unused(%s);", fc.translateExpr(s.Rhs[0]))
+				fc.Printf("$unused(%s);", fc.translateImplicitConversion(s.Rhs[0], fc.pkgCtx.TypeOf(s.Lhs[0])))
 				return
 			}
 			fc.Printf("%s", fc.translateAssign(lhs, s.Rhs[0], s.Tok == token.DEFINE))
--- a/compiler/utils.go
+++ b/compiler/utils.go
@@ -3,12 +3,14 @@ package compiler
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"go/ast"
 	"go/constant"
 	"go/token"
 	"go/types"
 	"net/url"
+	"runtime/debug"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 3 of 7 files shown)


compiler/package.go
```diff
@@ -120,7 +120,20 @@ func (pi packageImporter) Import(path string) (*types.Package, error) {
 	return pi.importContext.Packages[a.ImportPath], nil
 }
 
-func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, importContext *ImportContext, minify bool) (*Archive, error) {
+func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, importContext *ImportContext, minify bool) (_ *Archive, comError error) {
```
dmitshur, line 123: Consider modifying `err := recover()` to `e := recover()` and renaming `comError` to `err`, since that might look cleaner. `comError` only makes sense in the context of the defer, otherwise it is just the error that Compile returns, so `err` is a simple and clear name.

compiler/statements.go
```diff
@@ -22,6 +22,27 @@ func (fc *funcContext) translateStmtList(stmts []ast.Stmt) {
 }
 
 func (fc *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
+	defer func() {
+		err := recover()
+		if err == nil {
+			return
+		}
+		if _, yes := bailingOut(err); yes {
```
dmitshur, line 30: Consider also using `ok` instead of `yes` here, that's what most Go code does, making it more readable.

compiler/utils.go
```diff
@@ -738,3 +740,55 @@ func (st signatureTypes) Param(i int, ellipsis bool) types.Type {
...
+
+// FatalError is an error compiler panics with when it encountered a fatal error.
+//
+// FatalError implements io.Writer, which can be used to record any free-form
+// debugging details for human consumption. This information will be included
+// into String() result along with the rest.
+type FatalError struct {
+	cause interface{}
+	stack []byte
+	clues bytes.Buffer
```
dmitshur, line 752: Consider using `strings.Builder` here since it might be a better fit.

compiler/utils.go
```diff
@@ -738,3 +740,55 @@ func (st signatureTypes) Param(i int, ellipsis bool) types.Type {
...
+		return errors.New(s)
+	}
+	return fmt.Errorf("[%T]: %v", b.cause, b.cause)
+}
+
+// Write implements io.Writer and can be used to store free-form debugging clues.
+func (b *FatalError) Write(p []byte) (n int, err error) { return b.clues.Write(p) }
+
+func (b FatalError) Error() string {
+	buf := &bytes.Buffer{}
```
dmitshur, line 772: Similarly here.

compiler/version_check.go
```diff
@@ -19,6 +21,9 @@ const GoVersion = 16
 // at goroot, and reports an error if it's not compatible
 // with this version of the GopherJS compiler.
```
dmitshur, line 22: Consider documenting the change to CheckGoVersion's behavior here.

Perhaps it's also worth including at https://github.com/gopherjs/gopherjs#environment-variables.

compiler/version_check.go
```diff
@@ -19,6 +21,9 @@ const GoVersion = 16
...
 func CheckGoVersion(goroot string) error {
+	if nvc, err := strconv.ParseBool(os.Getenv("GOPHERJS_NO_VERSION_CHECK")); err == nil && nvc {
```
dmitshur, line 24: Maybe the name can be simplified to `GOPHERJS_VERSION_CHECK` to avoid the double negation? Since the error is being checked, it'll be fine when the env var is unset since `ParseBool` returns a non-nil error on the empty string. Another option is to use a non-negated verb like `GOPHERJS_SKIP_VERSION_CHECK`.

dmitshur (commented):
Thanks for working on this.

I don't have a chance to look very closely at the compiler changes, but was able to leave optional style suggestions for you to consider in case they're helpful.

compiler/statements.go
nevkontakte: I am aware of this convention, but was reluctant to follow it because in this particular case the meaning of "ok" is unclear: is it ok when we are panicing or when we aren't? "Yes" answers the question more directly, so I settled on it.

compiler/utils.go
nevkontakte: Done.

compiler/utils.go
nevkontakte: Done.

compiler/version_check.go
nevkontakte: Good point, I prefer `GOPHERJS_SKIP_VERSION_CHECK`, since it seems a bit more self-descriptive. I'm not much concerned about its length, since it should be only rarely needed by people working on GopherJS itself.

compiler/version_check.go
nevkontakte: Good point, done. I'll also add it to the developer instructions in the wiki once the PR is merged.

flimzy (approved):

Issue #1021, opened 2021-04-13
Proposal: materialize augmented GOROOT on disk before building


nevkontakte:
## Background

GopherJS ships with a set of "augmentations" to the standard library necessary to support JavaScript runtime. For user convenience, the augmentations are compiled into the `gopherjs` tool and are applied on the fly by modifying AST on the fly. Thanks to this, the user doesn't need to worry about patching Go distribution and/or maintaining separate GOROOTs for GopherJS and plain Go. Additionally, the user can switch between any number of GOROOTs without having to do extra steps.

However, this approach has several drawbacks:

1. These customizations are completely opaque to any kind of tooling: they can't be linted, vetted or used with any other tooling a user might want.
2. Transformations are done every time a build is performed, repeatedly re-doing the same work and making the build a bit slower. Overall this probably isn't so bad thanks to the incremental build cache though.
3. Such a sophisticated use case isn't well supported by the tooling Go provides (`go/build` and `golang.org/x/tools/go/packages` packages specifically). Specifically, `go/build` only works in GOPATH mode when build context has any of the file system access function are overridden; `go/packages` provides much more limited customization options: it either has to be done at ParseFile level, or by overriding whole contents of certain files in advance. So module support (#855) becomes much more complicated than it has to be.
4. This is more of a personal opinion, but mixing source augmentation and build process together makes the relevant code considerably harder to understand and maintain.

## Proposal

  - Before attempting a build, materialize GOROOT with all augmentations in a temporary directory (`os.TempDir()`).
    - GOROOT from the environment is used as the original.
    - Any files or directories from the original GOROOT that are not subject to augmentation are symlinked to the originals to conserve disk space.
    - Files that do require augmentation are processed and written on disk.
    - Path to the augmented GOROOT will include Go version from the original GOROOT as well as GopherJS version to allow for several different versions coexist on the same machine.
    - Optionally, some form of fingerprinting may be used to make sure that the augmented GOROOT is only regenerated when the inputs change.
    - Augmented GOROOT is considered disposable, and can be regenerated if it was deleted.
  - During the build the augmented GOROOT is used instead of the original.
  - `gopherjs root` subcommand is added so that the user can regenerate and get path to the augmented GOROOT for use with other tooling.

## Extras

[...]

PR #1026, opened 2021-05-15
natives: duplicate aliasing.go in x/crypto

benma:
This file is also present there, and `x/crypto` packages making use of
`x/crypto/internal/subtle` need the internal implementation as the one
in the stdlib.


nevkontakte:
Thanks for your investigation and this PR!

I am somewhat torn about it though… Technically, `golang.org/x/crypto/...` are not a part of the standard library, and it's infeasible for GopherJS to maintain patches for arbitrary libraries that may not be compatible upstream. I can see an argument that `golang.org/x` has somewhat of a special status in the ecosystem, but we are currently not testing new releases against any of those libraries, so it is likely to regress in future.

In particular, the plans we had for https://github.com/gopherjs/gopherjs/issues/1021 had an explicit assumption that we only augment standard library sources. Third-party libraries may or may not be available at that stage and specific versions will depend on a particular project.

@flimzy @dmitshur do you have any thoughts?

benma:
@nevkontakte thanks for the response.

Imho an exception is necessary here. Our project is running on GopherJS, we invested quite heavily in this stack. As you said, x/crypto has a special status, it is treated kind of like the stdlib.

> but we are currently not testing new releases against any of those libraries, so it is likely to regress in future.

We can add this particular internal to the test suite, and add some tests exercising chacha20poly1305 , so it wouldn't regress.

flimzy:
My thoughts, mostly off the top of my head, so I may be overlooking something:

1. For truly third-party packages, (i.e anything not in the stdlib and not in golang.org/x), the approach would be to ask the package maintainers to support GopherJS. I've had pretty good success with random third-party libraries accepting PRs to this end.  I wonder if golang.org/x would be willing to do the same?
2. If we do decide we need to augment this, can it be done in a separate library? Can we maintain a `gopherjs.org/x` repo, for example, so as not to complicate the core GopherJS package?

nevkontakte:
@benma my primary concern here is that there are only two active GopherJS maintainers and both of us have very limited time we can dedicate to it. It is barely enough to keep up with the existing scope and taking on more maintenance burden will likely put us back into the state the project was for the previous couple of years. If your team is actively using GopherJS, would you be interested in contributing some of your time to the project? Having an active group of contributors would allow us to be more ambitious about how much we can support.

Besides that, there's also a technical problem to consider. A significant technical difference between `golang.org/x` and the actual standard library is that they are released separately. GopherJS releases are very tightly coupled with Go releases, so for every Go version a matching release of GopherJS must be used. The problem is that a user can use absolutely any version of `golang.org/x/crypto` libraries with any version of Go and GopherJS. That would require us to maintain and ship a whole set of patches for different versions of the library, and build some kind of new infrastructure to detect and apply the correct patch — build tags will no longer be enough.

Echoing @flimzy's comment, maybe we could submit a patch to `golang.org/x/crypto` instead?

dmitshur:
Neither location is absolutely ideal. It's a patch that makes use of:
- internal GopherJS APIs (which may become different depending on GopherJS version)
API changes
- internal x/crypto APIs (which may become different depending on x/crypto version)

If added to GopherJS, then GopherJS maintainers need to keep up with x/crypto internal API changes. But if added to x/crypto, then x/crypto maintainers need to keep up with GopherJS internal API changes.

nevkontakte:
I was thinking that there could be a version of this patch that avoids pointer arithmetic, but also doesn't use GopherJS APIs. Or maybe we could try to improve GopherJS support for unsafe or reflect so that a patch wouldn't be necessary. I haven't looked at the surrounding code closely to tell for sure though.

benma:
I don't think that x/crypto is the right place to make changes - they should not be concerned with maintaining GopherJS compatibility, unless you can think of a trivial change that is not likely to regress on their end. 

> The problem is that a user can use absolutely any version of `golang.org/x/crypto` libraries with any version of Go and GopherJS.

While that is true, I don't see it as a great maintenance burden.

In my humble opinion, it wouldn't hurt to merge this PR as is, to make GopherJS work with x/crypto like it used to, and leave it at that. Chances are no new issues will come up, and if there will be, we can cross the bridge when we get there.

benma:
>  Or maybe we could try to improve GopherJS support for unsafe or reflect so that a patch wouldn't be necessary. I haven't looked at the surrounding code closely to tell for sure though.

That would indeed be the best, but I am not familiar enough with the internals to judge if that is possible.

nevkontakte:
Ok, here's my plan then:

 1. I'll wait until tomorrow on the off-chance that someone comes up with a better alternative.
 2. If no suggestions are made until then, I'll merge this PR as-is to unblock @benma and his colleagues, but with a significant caveat.

The caveat: we are unable to guarantee that this fix will work in the long term. I'll try to find a few hours to look at `x/crypto` code and see if there may be a patch I could send upstream that would not rely on GopherJS internals, but avoid pointer arithmetics as well. If I succeed — great, if not the following will likely happen:

 - The fix will exist as is, as long as it works.
 - If we decide to go forward with #1021 and we don't find a way to make this approach work with overlays for outside of the standard library, the fix will be removed from the overlays.
 - We *might* set up a forked version of `x/crypto` as a substitute, if we find a reasonably automated way to keep it up to date with the upstream. Cryptography in particular is one of those domains where being out of date can be an actual vulnerability, so if we find no automated solution to maintain a fork under the gopherjs umbrella, it will be up to the community to create and maintain one.

@flimzy @dmitshur please let me know what you think of this plan.

nevkontakte:
Alright, since there don't seem to be any objections, I am merging this PR. Thanks all!

benma:
Thanks a lot!

nevkontakte:
For posterity, I've looked into options for implementing `AnyOverlap()` without reliance on pointer arithmetics, and unfortunately I can't see any that would be sufficiently reliable to submit upstream. I found a discussion in https://groups.google.com/g/golang-nuts/c/_Pj9S_Ljp9o, which proposes comparing addresses of the last element of the underlying array, but one of the messages states that this will no longer be reliable if https://github.com/golang/go/issues/395 gets fixed, and it recently was.

So as dissatisfying as this might be, a fork is probably the only long-term option.

Issue #603, opened 2017-03-13
Invalid code generated for funcs with unnamed return parameters that defer blocking code.


nwidger:
https://gopherjs.github.io/playground/#/JcN1xwSTbp

If I run this Playground link in Firefox 52.0, the `should get here` alert dialog appears.  If I run the same link in Chrome 56.0.2924.87, it prints `calling alert(err)` but not `returned from alert(err)` and the following stack trace is generated:

```
Uncaught (in promise) TypeError: Cannot read property 'Error' of undefined
    at Object.main [as $blk] (eval at $b (playground.js:66), <anonymous>:2943:14)
    at Object.$init [as $blk] (eval at $b (playground.js:66), <anonymous>:2963:68)
    at fun (eval at $b (playground.js:66), <anonymous>:1477:37)
    at $goroutine (eval at $b (playground.js:66), <anonymous>:1475:19)
    at $runScheduled (eval at $b (playground.js:66), <anonymous>:1515:7)
    at $schedule (eval at $b (playground.js:66), <anonymous>:1531:5)
    at queueEntry (eval at $b (playground.js:66), <anonymous>:1686:11)
    at $send (eval at $b (playground.js:66), <anonymous>:1556:5)
    at $b (eval at $b (playground.js:66), <anonymous>:2895:70432)
    at v.$externalizeWrapper (eval at $b (playground.js:66), <anonymous>:1871:24)
```

Removing the `defer` statement after `http.Get` resolves the issue in Chrome:

https://gopherjs.github.io/playground/#/DK-J3eMXYr

Is using defer like this incorrect?  If it is I would expect to either see a compilation error or see the same issue in both Firefox and Chrome.

shurcooL:
I can reproduce in Chrome 57.0.2987.98.

This is definitely a bug and shouldn't be happening.

shurcooL:
It seems the value of `err` is somehow `undefined` (aka `js.Undefined`).

https://gopherjs.github.io/playground/#/Q5xYbr2_yf

That prints `undefined` in browser console in Chrome, which is very unexpected.

nwidger:
It definitely seems to be related to `io.Copy` being called in a `defer` statement.  This does not exhibit the same issue:

https://gopherjs.github.io/playground/#/d-r8POieip

shurcooL:
I suspect this may be related to the fix 50101461d59f26b13c876f57aab314ce55387671, and #381, #426, #493 issues it resolved. /cc @neelance

[...]

PR #1035, opened 2021-06-15
Refactor the analysis package.

nevkontakte:
This doesn't make any changes to the existing logic, but makes the code
a bit easier to read (or so I hope) and explains some of the logic I had
to reverse-engineer in the comments.

Overall this should make fixing
https://github.com/gopherjs/gopherjs/issues/603 a bit easier.

### DIFF
--- a/compiler/analysis/info.go
+++ b/compiler/analysis/info.go
@@ -11,72 +11,93 @@ import (
 
 type continueStmt struct {
 	forStmt      *ast.ForStmt
-	analyzeStack []ast.Node
+	analyzeStack astPath
+}
+
+func newContinueStmt(forStmt *ast.ForStmt, stack astPath) continueStmt {
+	cs := continueStmt{
+		forStmt:      forStmt,
+		analyzeStack: stack.copy(),
+	}
+	return cs
+}
+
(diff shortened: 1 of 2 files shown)


compiler/analysis/info.go
```diff
@@ -97,158 +122,206 @@ func AnalyzePkg(files []*ast.File, fileSet *token.FileSet, typesInfo *types.Info
...
 }
 
-func (c *FuncInfo) Visit(node ast.Node) ast.Visitor {
+type FuncInfo struct {
+	HasDefer bool
+	// Nodes are "flattened" into a switch-case statement when we need to be able
+	// to jump into an arbitrary position in the code with a GOTO statement, or
+	// resume a goroutine after a blocking call unblocks.
+	Flattened map[ast.Node]bool
+	// Blocking indicates that either the AST node itself or its may block
```
flimzy, line 131: I think this sentence is missing a word?  Doesn't quite make sense.

flimzy (approved):

compiler/analysis/info.go
nevkontakte: Good catch, fixed.

PR #1037, opened 2021-06-19
Fix undefined return when a deferred function blocks.

nevkontakte:
This change makes sure that whenever a function resumes execution after getting blocked by a deferred call, the return value is not lost. We do this by inserting a resumption point right before the return statement. If the return expression is not empty, we store it in a temporary variable to make sure it doesn't get evaluated twice.

### DIFF
--- a/compiler/analysis/info.go
+++ b/compiler/analysis/info.go
@@ -1,9 +1,11 @@
 package analysis
 
 import (
+	"fmt"
 	"go/ast"
 	"go/token"
 	"go/types"
+	"strings"
 
 	"github.com/gopherjs/gopherjs/compiler/astutil"
 	"github.com/gopherjs/gopherjs/compiler/typesutil"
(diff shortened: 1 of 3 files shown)


compiler/analysis/info.go
```diff
@@ -136,6 +166,8 @@ type FuncInfo struct {
 	GotoLabel map[*types.Label]bool
 	// List of continue statements in the function.
 	continueStmts []continueStmt
+	// List of return statements in the function.
+	returnStmts []astPath
```
flimzy, line 170: If this is only used in determining if the return may block due to a defer, should this become `returnStmtsAfterFirstDefer`?

I.e. in the common case of something like:

```golang
resp, err := http.Get(...)
if err != nil {
    return err
}
defer resp.Body.Close()

/// some time later
return ...
```

The first `return` would never be blocked by the defer.

flimzy (approved):
Looks reasonable, with the caveat that there is probably room for some streamlining.  I offered one possible suggestion along those lines. I haven't thought through the full implications, but I don't think it should block merging this. If it's useful, it can be a follow-up PR.

compiler/analysis/info.go
nevkontakte: I thought about that, but existence of goto defeats that logic. 

```go
resp, err := http.Get(...)
goto gotcha2
gotcha1:
if err != nil {
    return err
}
gotcha2:
defer resp.Body.Close()
goto gotcha1
```

Even though I suspect examples like this that also make sense would be rare, they might pop up in generated code. This could be solved with the use of control flow analysis, but that's something beyond my grasp just yet ¯\\\_(ツ)\_/¯

On a tangential note, cases like this make me wonder if there's merit to using something like https://pkg.go.dev/golang.org/x/tools/go/ssa instead of pure AST to generate code. It would certainly lead to less reader-friendly generated JavaScript, but might simplify the compiler? Or not 🤷‍♂️

Issue #855, opened 2018-08-26
module-aware building


hajimehoshi:
Now Go 1.11 introduces module-get, I suggest to change `gopherjs build` or other commands to be module-aware. For example, I'd want the below code to be workable:

```sh
mkdir foo
cd foo
go mod init example.com/m
go get github.com/gopherjs/gopherjs
gopherjs build -tags=example github.com/hajimehoshi/ebiten/examples/sprites@master
```

myitcv:
FWIW I have module-aware building in branch `go1.11` of my fork at https://github.com/myitcv/gopherjs

hajimehoshi:
@myitcv I discussed with @dmitshur , and we agreed that it is fine to merge your module-aware building. I'd be happy if you could make a PR. Thanks!

myitcv:
@hajimehoshi back in April I raised https://github.com/gopherjs/gopherjs/issues/799. As a follow up to that, and to discuss [my open PRs](https://github.com/gopherjs/gopherjs/pulls/myitcv), @dmitshur and I had a call in which Dmitri said that his plan was to "sunset" GopherJS with the arrival of experimental WASM support in Go 1.11. Hence he did not want to grant access to additional people or accept any more PRs, other than a PR he was working on for Go 1.11 support (and potentially in other exceptional circumstances). Hence Dmitri's suggestion was to do things in a fork, which is how https://github.com/myitcv/gopherjs came about.

Has that position now changed?

As I suggested to Dmitri on our original call, perhaps we could also get some clarification on the current position (with respect to the https://github.com/myitcv/gopherjs fork) added to the main README so anyone else considering contributing a PR is clear.

Thanks

hajimehoshi:
> Has that position now changed?

Thank you for elaborating and sorry for my lack of explanation. This position is not changed basically, but I asked @dmitshur that I am eager to have module support of GopherJS. @dmitshur permitted me that under the condition of my review and taking care of. But right, this change seemed to be conflicting with the current policy of GopherJS. We need to be on the same page.

IMHO, GopherJS is still useful for a while even after Wasm comes since Wasm still has several problems (e.g. OOM at mobile browsers https://github.com/golang/go/issues/27462, bigger binary size than GopherJS, etc.) so Wasm cannot become a complete substitute of GopherJS so far. Actually I cannot finish to switch to Wasm for my libs yet.

[...]

Issue #693, opened 2017-09-05
proposal: Limit support to one GOOS value (primarily for stdlib).


dmitshur:
The GopherJS compiler supports building output only for GOARCH=js value, and any GOOS value.

That means it will produce valid and different output for this Go package, depending on which GOOS value is used:

```Go
// +build plan9

package p

const Message = "It was a hot summer..."
```

```Go
// +build !plan9

package p

const Message = "It was a cold winter..."
```

Of course, it also supports cross-compiling, so one can produce different outputs by doing:

```bash
$ GOOS=plan9 gopherjs build
# hot summer

$ GOOS=solaris gopherjs build
# cold winter
```

I propose we limit support to just a single fixed GOOS value.

This is primarily meant to apply to the Go standard library, which we must manually apply overrides for in order to get it to build and work correctly for GOARCH=js. See https://github.com/gopherjs/gopherjs/commits/master/compiler/natives/src. However, I think we should apply for all packages for simplicity/consistency. Applying the limit to just Go stdlib but not 3rd party packages would be weird and unnecessary.

The motivation is simple. It costs a lot of developer time to support Go standard library for GopherJS. Right now, I don't think this project can afford to continue to support all GOOS values. It's an open source project, with no funding, and most of the progress is made by volunteers contributing in their spare time in after work hours.

The value gained from supporting multiple GOOS values is small. The cost is prohibitively high.

The compiler will continue to be able to (theoretically) compile for many GOOS values, we're just not going to support GOOS values other than a single one that we pick.

shurcooL:
This came up when I investigated what it'd take to fix #688.

I considered 3 alternative fixed GOOS values to use:

1. `darwin` - this is the same GOOS that the 2 primary GopherJS developers (@neelance and me) use when developing GopherJS and using it

2. `linux` - it's one of the most popular and generic `GOOS` values. Even `android` build tag piggybacks off it. Also, unlike `darwin`, it's open source. Given open source and proprietary, I'd rather we support an open source OS.

3. `nacl` - this is the GOOS environment that is most similar to `GOARCH=js`. In fact, there are a number of times where I base `GOARCH=js` implementations after NaCL ones. There are tests that could be simplified if we also targetted `nacl`.

Using `darwin` is the easiest and works best from my testing. It's what Richard and I have used all this time. I'm sure many other users also use it.

Using `linux` would require some work; it didn't work out of box for me.

[...]

PR #1042, opened 2021-07-18
Support Go Modules builds.

nevkontakte:
Fixes https://github.com/gopherjs/gopherjs/issues/855.

I took the approach of leveraging module support that's built into the `go/build` package. Doing so has two prerequisites:

 1. `go/build.Context` instance must not have its file system methods overridden.
 2. ReleaseTags must match the default.

Most of the changes in this PR tend to the #1. We were overriding FS access methods in order to make packages embedded in the VFS available during the build. Now we use two separate contexts chained together to access packages on the real FS and the embedded VFS. This refactoring is useful in the long run to make package loading logic more encapsulated in the build package.

The second prerequisite is somewhat bogus for reasons explained in the `build/versionhack` package. https://github.com/golang/go/issues/46856 is filed in the upstream for a better behavior, but until it's fixed, we will just work around it.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -1,3 +1,8 @@
+// Package build implements GopherJS build system.
+//
+// WARNING: This package's API is treated as internal and currently doesn't
+// provide any API stability guarantee, use it at your own risk. If you need a
+// stable interface, prefer invoking the gopherjs CLI tool as a subprocess.
 package build
 
 import (
@@ -8,7 +13,6 @@ import (
 	"go/scanner"
 	"go/token"
 	"go/types"
-	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -53,52 +57,13 @@ func (e *ImportCError) Error() string {
 // with GopherJS compiler.
 //
 // Core GopherJS packages (i.e., "github.com/gopherjs/gopherjs/js", "github.com/gopherjs/gopherjs/nosync")
-// are loaded from gopherjspkg.FS virtual filesystem rather than GOPATH.
-func NewBuildContext(installSuffix string, buildTags []string) *build.Context {
+// are loaded from gopherjspkg.FS virtual filesystem if not present in GOPATH or
+// go.mod.
+func NewBuildContext(installSuffix string, buildTags []string) XContext {
 	gopherjsRoot := filepath.Join(DefaultGOROOT, "src", "github.com", "gopherjs", "gopherjs")
-	return &build.Context{
-		GOROOT:        DefaultGOROOT,
-		GOPATH:        build.Default.GOPATH,
-		GOOS:          build.Default.GOOS,
-		GOARCH:        "js",
-		InstallSuffix: installSuffix,
-		Compiler:      "gc",
-		BuildTags: append(buildTags,
-			"netgo",            // See https://godoc.org/net#hdr-Name_Resolution.
-			"purego",           // See https://golang.org/issues/23172.
-			"math_big_pure_go", // Use pure Go version of math/big.
-		),
-		ReleaseTags: build.Default.ReleaseTags[:compiler.GoVersion],
-		CgoEnabled:  true, // detect `import "C"` to throw proper error
-
-		IsDir: func(path string) bool {
-			if strings.HasPrefix(path, gopherjsRoot+string(filepath.Separator)) {
-				path = filepath.ToSlash(path[len(gopherjsRoot):])
-				if fi, err := vfsutil.Stat(gopherjspkg.FS, path); err == nil {
-					return fi.IsDir()
-				}
-			}
-			fi, err := os.Stat(path)
-			return err == nil && fi.IsDir()
-		},
-		ReadDir: func(path string) ([]os.FileInfo, error) {
-			if strings.HasPrefix(path, gopherjsRoot+string(filepath.Separator)) {
-				path = filepath.ToSlash(path[len(gopherjsRoot):])
-				if fis, err := vfsutil.ReadDir(gopherjspkg.FS, path); err == nil {
-					return fis, nil
-				}
-			}
-			return ioutil.ReadDir(path)
-		},
-		OpenFile: func(path string) (io.ReadCloser, error) {
-			if strings.HasPrefix(path, gopherjsRoot+string(filepath.Separator)) {
-				path = filepath.ToSlash(path[len(gopherjsRoot):])
-				if f, err := gopherjspkg.FS.Open(path); err == nil {
-					return f, nil
-				}
-			}
-			return os.Open(path)
-		},
+	return &chainedCtx{
+		primary:   goCtx(installSuffix, buildTags),
+		secondary: embeddedCtx(&withPrefix{gopherjspkg.FS, gopherjsRoot}, installSuffix, buildTags),
 	}
 }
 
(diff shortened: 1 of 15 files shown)


aykevl:
Didn't look at the code, but this might be useful: https://github.com/tinygo-org/tinygo/pull/941
In TinyGo, I've basically constructed a modified GOROOT tree (in a temporary directory) and then used `go list -json -deps` directly.

nevkontakte:
Thanks! I actually did look at that exact PR at one point, and [materializing GOROOT](https://github.com/gopherjs/gopherjs/issues/1021) was behind the two of my first attempts to support Modules (this PR representing the third).

This did, however, prove somewhat tricky to get right:

  - Standard library augmentation GopherJS does is more complicated than TinyGo. The latter replaces whole packages (if I remember correctly), while GopherJS rewrites parts of AST. Writing this augmented AST back on disk and making sure it can be correctly parsed again revealed a whole host of problematic edge cases.
  - There are some complications around `syscall` package and how we deal with OS-dependent code introduce another layer of complexity. Fixing https://github.com/gopherjs/gopherjs/issues/693 would eliminate this problem, but at that point it looked a lot like a rabbit hole :)

You are welcome to check out https://github.com/nevkontakte/gopherjs/tree/go-tool/cmd/gopherjs-ng which has some of those experiments if you are curious.

I still like the idea of materializing augmented GOROOT on disk, and I'm going to continue working on it, but I felt like we need a working support for Modules before 1.17, which was rumored to deprecate GOPATH entirely. This is PR is the simplest solution to the problem I found, but it also doesn't address some of the issues which make our build system so complicated and weird.

nevkontakte:
Also worth mentioning that initially `go/build` did not support modules and was not meant to, but at some point (not sure when, but probably recently) to did get some level of it by doing that same `go list` under the hood. So maybe TinyGo implementation could be simplified a bit at this stage 🙂

flimzy:
This seems to fail for me when trying to build a project containing the `spf13/cobra` library.  Reproduction case in gist here: https://gist.github.com/flimzy/3485a8e31377c9b07d8622351813c0f7

nevkontakte:
I've fixed the panic, which was triggered when expanding patterns containing "/...". I also replaced the gotool package with an actual go tool invocation, since the package predated go modules and didn't know how to match them anyway.

aykevl:
>  * Standard library augmentation GopherJS does is more complicated than TinyGo. The latter replaces whole packages (if I remember correctly), while GopherJS rewrites parts of AST. Writing this augmented AST back on disk and making sure it can be correctly parsed again revealed a whole host of problematic edge cases.

Correct for TinyGo. I looked at what GopherJS did and intentionally did not do that. I made sure that only whole packages would be replaced, to simplify the build system. It makes working on the standard library for TinyGo a bit harder, but not by much. And it makes the build system a _lot_ simpler.

>   * There are some complications around `syscall` package and how we deal with OS-dependent code introduce another layer of complexity. Fixing [proposal: Limit support to one GOOS value (primarily for stdlib). #693](https://github.com/gopherjs/gopherjs/issues/693) would eliminate this problem, but at that point it looked a lot like a rabbit hole :)

That proposal seems like a very sensible thing to do!

What I do in TinyGo is that for every supported target that isn't in Go (that is, WASI and the various baremetal systems) I pick a single GOOS/GOARCH pair for the standard library to use. So for example, for WebAssembly in the browser (which is supported by the Go standard library) I use GOOS=js GOARCH=wasm as usual, but for WASI (which isn't supported by the Go standard library) I use GOOS=linux GOARCH=arm. This works very well in practice, only the syscall package and some packages that interact with the runtime (net, os, reflect, ...) need to be replaced for that to work. This means that most code in existence will think it's a different OS/arch, but in practice it'll usually run just fine. In the (few!) edge cases where it doesn't, they can use the `tinygo` build tag. One relevant PR in this context: https://github.com/tinygo-org/tinygo/pull/1964.
If this is possible in GopherJS, I can recommend this strategy. It keeps the build process much simpler and (perhaps more importantly) means that supporting a new Go version in GopherJS will be a lot simpler: TinyGo currently needs to do very little to support new Go versions and in fact supports multiple at the same time - this is at least in part due to the policy of only replacing standard library packages as a whole (so that we reduce our reliance on implementation details). For GopherJS, using GOOS=js GOARCH=wasm might be a good strategy. For users which want to differentiate between "true" js/wasm and the GopherJS version, you can simply add a `gopherjs` build tag (just like there are `gc` and `gccgo` build tags to differentiate other compilers).
That said, I do recognize that this would be a really big change which could very well break things as they are.

> Also worth mentioning that initially `go/build` did not support modules and was not meant to, but at some point (not sure when, but probably recently) to did get some level of it by doing that same `go list` under the hood. So maybe TinyGo implementation could be simplified a bit at this stage :slightly_smiling_face:

We currently call `go list` directly and this works well in practice. I don't think we'll want to change this. See https://github.com/golang/go/issues/38953 for one reason why we're using raw `go list`.

paralin:
On the topic of alternative package implementations, I was wondering if @aykevl  and @nevkontakte had any comment on the following idea: https://github.com/paralin/goscript

Recently there are great tools like https://esbuild.github.io/ which are native-go applications to parse and transform javascript, typescript, jsx, ... 

Go's syntax seems like it can be directly translated into TypeScript syntax, with some package replacements (as you're describing have been done in tinygo) for optimization / compat purposes

aykevl:
I think GoScript is getting very much off topic. I'm happy to discuss this in a different place.

nevkontakte:
@aykevl thanks for your detailed comment, all of what you said makes sense. In general I have high hopes that a lot of GopherJS standard library overlays could be simplified if we try to reuse wasm implementations from the upstream. That said, there's a significant challenge that is unique to GopherJS: we don't have access to the raw memory. So we have to patch any library that does pointer arithmetics, which means it's considerably more packages than `runtime`, `os`, `net` and friends. One of my biggest objectives is to make GopherJS less divergent from upstream, but that's a lot of work and I have only so many hours in a day :)

@paralin I would agree with @aykevl that this discussion would be straying quite far from this PR's subject, so how about we continue it in [#gopherjs channel](https://app.slack.com/client/T029RQSE6/C039C0R2T) in the Gopher Slack? I like the idea of goscript, but only after having worked on GopherJS compiler for a few months I began to appreciate how many subtle challenges there are if you start considering features like goto, goroutines and runtime packages.

paralin:
@nevkontakte Getting the following when importing syscall/js - that package is supported for compat w/ wasm code, right?

`could not import syscall/js (invalid package name: "")`

Thanks for the work on this

nevkontakte:
@paralin I can't reproduce the error with this minimal example: https://gist.github.com/nevkontakte/56c18ec181a10881ddd84975bd8bc782

Commands I run:

```
gopherjs build .
node run *.js
```

paralin:
@nevkontakte repro here: https://github.com/paralin/gopherjs-repro-1042 just run "go run ./" in the root dir.

nevkontakte:
@paralin I've fixed the bug, this was indeed a regression introduced by this PR. I also fixed the bug which caused CI to skip `syscall/js` package in testing.

@flimzy I think this PR is now complete. Could you take another look at it?

flimzy (approved):

paralin:
@nevkontakte thanks for looking into it & doing the work to implement modules with gopherjs!

PR #1048, opened 2021-08-08
Support go1.17 slice to array pointer conversion.

nevkontakte:
Conversion is fully supported for slices of numeric types, which is
probably the most common use case for this feature judging by the
discussion in the proposal.

However, it is only partially supported for slices of complex types
(e.g. strings), since they are backed by the JavaScript's built-in Array
type, which lacks ability to share backing memory among subarrays.
Conversion works in cases when the slice is converted into array that
exactly matches the slice's underlying array, but panics if we are
trying to convert a subslice. I feel like an explicit failure is better
than a chance of introducing subtle bugs.

It also seems that GopherJS internally doesn't really distinguish
between array types and pointer to array types, which makes the whole
implementation somewhat messy when it comes to nil vs non-nil pointers
to arrays.

Last but not least, I've moved pointer cache for numeric arrays into the
backing ArrayBuffer, which ensures that pointers are comparable between
different subarrays.


nevkontakte:
I suspect that a matching change will have to be implemented in the reflect package, but I haven't had time to look into that yet.

nevkontakte:
I've added support for converting slices to arrays via `reflect` package. Unfortunately, `TestConvert` is still failing for a number of cases. There is some odd behavior around named array pointer types (e.g. `type myArray *[0]byte`), which I can't quite make sense of.

I feel like we really ought to rethink type system implementation and try to bring it closer to Go's. A lot of our struggles come from the fact that GopherJS doesn't really define consistent semantics for pointers and interfaces, and largely pretends they don't exist at runtime.

flimzy (approved):

nevkontakte:
Even though this PR is strictly speaking incomplete (a bunch of relevant `reflect` test cases are still failing), I'm going to merge it into the `wip-go1.17` branch to reduce chances for merge conflicts and such. I'm going to look further into the issues later this week.

PR #1050, opened 2021-08-18
prelude: fix $internalize for Function for Chrome extension

benma:
Loading a GopherJS compiled Javascript file in a Chrome extension
resulted in this error:

`Uncaught Error: interface conversion: interface is
map[string]interface {}, not func(...interface {}) *js.Object`

The reason was that these two functions from natives/src/reflect and
natives/src/internal/reflectlite were internalized incorrectly:

```
var callHelper = js.Global.Get("$call").Interface().(func(...interface{}) *js.Object)
```

and

```
var selectHelper = js.Global.Get("$select").Interface().(func(...interface{}) *js.Object)
```

This compiles to:

```
callHelper = $assertType($internalize($call, $emptyInterface), funcType$1);
```

Since `$call` is a Function, `$internalize()` should have entered the
`case Function` switch case. It does so in all Browsers, but in a
Chrome extension, it fell through to the default case, leading to the
result being a map instead of a function.

I could not find any information about why a function constructor is
different in Chrome extensions, but this patch just compares to
another function's constructor, making it work in Browsers as well as
in Chrome extensions.


nevkontakte:
Thanks for the fix, and appreciate a detailed explanation.

Could you please also update minified prelude? `go generate ./compiler/perlude/...`

benma:
> Thanks for the fix, and appreciate a detailed explanation.
> 
> Could you please also update minified prelude? `go generate ./compiler/perlude/...`

Done.

benma:
Thanks!

A quick new release would be appreciated as this is blocking our progress. :pray:

nevkontakte:
No problem, there you go: https://github.com/gopherjs/gopherjs/releases/tag/1.16.4%2Bgo1.16.7 :)

PR #1059, opened 2021-09-17
Get `hash/maphash` to work again

flimzy:
This pulls in a few select functions from Go 1.16, which were optimized and thus made unsafe in Go 1.17.

https://github.com/golang/go/issues/47342 should render this change obsolete if/when implemented.


nevkontakte (approved):
Let's also remove the package from `.std_test_pkg_exclusions` to make sure our CI tests it.

PR #1070, opened 2021-10-02
Preserve position information for top-level variable declarations.

nevkontakte:
This allows sourcemap to be generated for top-level variable
declarations with initializer, for example:

```go
var x = initX() // Might panic.
```

This is particularly helpful when debugging panics in such initializers.


nevkontakte:
Eh, apparently auto-merge doesn't await to a review if requested.

PR #1082, opened 2021-11-01
64-bit integer arithmetic performance optimization

nevkontakte:
This PR has two changes aimed on improving math performance under GopherJS:

  - Optimized 64-bit int multiplication algorithm, ~15x faster than the original one. It is still orders of magnitude slower than native 64-bit operations on a modern CPU, but it improves `crypto/ed25519` performance quite a bit.
    
    ```
    name            old time/op  new time/op  delta
    KeyGeneration   29.9ms ± 0%  15.6ms ± 1%  -47.79%  (p=0.008 n=5+5)
    NewKeyFromSeed  30.1ms ± 1%  15.7ms ± 2%  -47.89%  (p=0.008 n=5+5)
    Signing         38.3ms ± 1%  20.2ms ± 2%  -47.31%  (p=0.008 n=5+5)
    Verification     110ms ± 1%    52ms ± 1%  -52.47%  (p=0.008 n=5+5)
    ```
    I think there is some potential for further improvement in adopting JavaScript's `BigInt` for 64-bit math, but browser support for it is far from universal, and it will require significant changes to the compiler

  - Removed use of 64-bit ints in `math/bits` package where possible. This also improved performance of `Add32`, `Mul32`, `Div32`, etc. functions by 10-20x. This had improved performance of asymmetric crypto packages, such as `crypto/rsa`:
    
    ```
    name                  old time/op  new time/op  delta
    RSA2048Decrypt         1.61s ± 0%   0.18s ± 2%  -88.77%  (p=0.008 n=5+5)
    RSA2048Sign            1.68s ± 3%   0.19s ± 2%  -88.56%  (p=0.008 n=5+5)
    3PrimeRSA2048Decrypt   846ms ± 2%   123ms ± 1%  -85.48%  (p=0.008 n=5+5)
    ```

Some more specific benchmarks are available in commit comments. Overall this reduced standard library tests execution time on CI from 14 to 7 minutes, which was my main motivation :)


flimzy (approved):
Very nice!

PR #1090, opened 2021-11-21
Correctly handle built-ins and `js.Object` methods with go keyword.

### LINKED ISSUE #547: `go close(ch)` causes compile error
http://www.gopherjs.org/play/#/uhMTBoBe53

```
angular.js:9937 Error: Unhandled object: *types.Builtin
```

nevkontakte:
Built-ins and `js.Object` methods are represented by JS expressions that may not be straightforward function calls, or reference functions defined in the prelude. In such cases, the callable expression must be wrapped in a lambda.

#998 has fixed a similar issue for the `defer` keyword, and this change generalizes the same fix for the `go` keyword.

This change also contains a small improvement to the compiler panic message, which makes it easier to read at-a-glance.

Fixes #547.

### DIFF
--- a/tests/goroutine_test.go
+++ b/tests/goroutine_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"context"
 	"fmt"
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


tests/goroutine_test.go
```diff
@@ -265,3 +266,18 @@ func TestEventLoopStarvation(t *testing.T) {
 	}()
 	<-ctx.Done()
 }
+
+func TestGoroutineBuiltin(t *testing.T) {
+	go print("")
+}
```
flimzy, line 272: What does this test? I suppose a failure crashes? Might be worth commenting to that effect for clarity?

tests/goroutine_test.go
```diff
@@ -265,3 +266,18 @@ func TestEventLoopStarvation(t *testing.T) {
...
+
+func TestGoroutineJsObject(t *testing.T) {
+	if !(runtime.GOOS == "js" || runtime.GOARCH == "js") {
+		t.Skip("Test requires GopherJS")
+	}
+	o := js.Global.Get("Object").New()
+	go o.Set("x", "y")
+	for o.Get("x").String() != "y" {
+		runtime.Gosched()
+	}
```
flimzy, line 282: Same comment here.

flimzy (approved):

tests/goroutine_test.go
nevkontakte: Added a comment and also improved the test to be more conclusive.

tests/goroutine_test.go
nevkontakte: Done.

PR #1097, opened 2021-12-28
Better stack trace parsing in the browser

flimzy:
This reimplements and expands on #658.

Getting these tests to work was a bit of a challenge, and requires an ugly hack.  We don't run the standard tests against the `runtime` package, since the majority of that functionality in upstream is irrelevant to us. So adding tests to that package, where this new functionality lives, is meaningless by default.

Getting those tests to work is no small task, due to many unavailable symbols, etc. And making those symbols available is no small task, because the GopherJS compiler has several special cases surrounding the `runtime` package that prevent our overlays from being used, etc.

To work around this, we run the new tests with `gopherjs test ./compiler/natives/src/runtime`, which bypasses the checks for the `runtime` package.  But it does mean we need to add one more special case for `runtime`, and stop checking its imports list against upstream.

In the grand scheme of things, adding one more exception to something that's already a huge special case didn't seem like that big of a deal, but maybe others disagree.

On to the functionality... This is more robust than #658 (one great reason to uses tests, to make that easy!)  My browser (Chrome) actually outputs multiple formats of stack trace lines, so this new implementation tries to parse all of them.

Some of the formats omit the function name and/or file name, which I've indicated with `<none>` and `<anonymous>` in the response from `runtime.Caller()`.  I'm not sure if that's actually a good idea, but it seemed more descriptive than simply returning the empty string, which is the only alternative that occurred to me.

With this PR merged, it should be trivial to add new stack traces (i.e. from Safari, Mozilla, wherever) and extend the parsing functionality to be more robust.


compiler/natives/src/runtime/runtime_test.go
```diff
@@ -0,0 +1,80 @@
...
+	at HTMLInputElement.c (https://ajax.googleapis.com/ajax/libs/angularjs/1.2.18/angular.min.js:31:207)`,
+			want: `foo https://gopherjs.github.io/playground/playground.js 102
+main https://gopherjs.github.io/playground/playground.js 102
+$init https://gopherjs.github.io/playground/playground.js 102
+$goroutine https://gopherjs.github.io/playground/playground.js 102
+$runScheduled https://gopherjs.github.io/playground/playground.js 102
+$go https://gopherjs.github.io/playground/playground.js 102
+eval https://gopherjs.github.io/playground/playground.js 102
+eval https://gopherjs.github.io/playground/playground.js 102
+eval  0
```
nevkontakte, line 51: These lines in the stack trace look strange to me. I would expect it would print <unknown> or <anonymous> instead of just blanks.

compiler/natives/src/runtime/runtime_test.go
```diff
@@ -0,0 +1,80 @@
...
+$b https://gopherjs.github.io/playground/playground.js 102
+k.e.$externalizeWrapper.e.$externalizeWrapper [as run] https://gopherjs.github.io/playground/playground.js 5
+ at   0
```
nevkontakte, line 54: I would expect "https://ajax.googleapis.com/ajax/libs/angularjs/1.2.18/angular.min.js 175" to be present in this stack trace instead of being just "at 0".

circle.yml
```diff
@@ -137,6 +137,10 @@ jobs:
         command: |
           set +e
           ulimit -s 10000
+          # RUNTIME_TEST_HACK: This ugly hack works around all the special cases
```
nevkontakte, line 140: I think I can live with this hack, at least until we can make `runtime` tests pass :)

However, I can propose an alternative hack: make `parseCallstack` exported and move the test to gopherjs/tests package. Since it will be hidden in the overlay sources, I doubt that anyone will ever discover it or try to use it for anything and it would be a bit less of a special case. A stern warning not to use it may also help the situation.

There is probably also a way to test it by getting the current call stack and examining `runtime.CallersFrames()`, but in that case it's harder to know what the expected correct values would be, so I'm not really proposing that.

compiler/natives/src/runtime/runtime_test.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/gopherjs/gopherjs/js"
+)
+
+func Test_parseCallstack(t *testing.T) {
+	tests := []struct {
```
nevkontakte, line 15: It would be useful to add another couple of test cases:
  - a stack trace captured with sourcemaps enabled
  - a stack trace generated by nodejs.

compiler/natives/src/runtime/runtime_test.go
```diff
@@ -0,0 +1,80 @@
...
+		name  string
+		input string
+		want  string
+	}{
+		{
+			name: "Chrome 96.0.4664.110 on Linux",
+			input: `at foo (eval at $b (https://gopherjs.github.io/playground/playground.js:102:11836), <anonymous>:25887:60)
```
nevkontakte, line 22: nit: this test case can probably be trimmed down in size by removing similar lines, which would help readability.

nevkontakte (approved):
Note that `runtime.Stack()` and, by extension, `runtime/debug.Stack()` do not currently do stack trace parsing/formatting. This means that panic stack traces printed to the console will remain as-is.

This is not a bad thing (I personally prefer to have the original stack trace with more breadcrumbs), just making sure it matches your expectations.

compiler/natives/src/runtime/runtime_test.go
flimzy: Agreed. I'll take a closer look at this.

compiler/natives/src/runtime/runtime_test.go
flimzy: Agreed. I can easily add Mozilla, too, I think, while I'm at it.

compiler/natives/src/runtime/runtime_test.go
flimzy: Good point. I might do each line as a separate test, too.  I also think I can tidy up `parseCallstack` a bit more.

flimzy:
> Note that runtime.Stack() and, by extension, runtime/debug.Stack() do not currently do stack trace parsing/formatting. This means that panic stack traces printed to the console will remain as-is.

> This is not a bad thing (I personally prefer to have the original stack trace with more breadcrumbs), just making sure it matches your expectations.

I agree. In fact, if it were possible to encapsulate the original stack trace in the return from `Caller()` and `Callers()`, that would probably be ideal, but given the limited data type those return, that's not really an option.

circle.yml
flimzy: Nice suggestion.  It worked pretty smoothly.

flimzy:
@nevkontakte Care to take another look at the updates before I merge?

nevkontakte:
LGTM, thanks for making the changes!

PR #742, opened 2018-01-27
Orders source files before compilation to ensure reproducible output

dave:

PR #1100, opened 2022-01-03
Orders source files before compilation to ensure reproducible output

flimzy:
Rebase of #742

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -225,7 +225,10 @@ func parseAndAugment(xctx XContext, pkg *PackageData, isTest bool, fileSet *toke
 			if err != nil {
 				panic(err)
 			}
-			file, err := parser.ParseFile(fileSet, fullPath, r, parser.ParseComments)
+			// Files should be uniquely named and in the original package directory in order to be
+			// ordered correctly
+			newPath := path.Join(pkg.Dir, "__"+name)
+			file, err := parser.ParseFile(fileSet, newPath, r, parser.ParseComments)
 			if err != nil {
 				panic(err)
 			}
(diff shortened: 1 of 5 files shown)


build/build.go
```diff
@@ -225,7 +225,10 @@ func parseAndAugment(xctx XContext, pkg *PackageData, isTest bool, fileSet *toke
 			if err != nil {
 				panic(err)
 			}
-			file, err := parser.ParseFile(fileSet, fullPath, r, parser.ParseComments)
+			// Files should be uniquely named and in the original package directory in order to be
+			// ordered correctly
+			newPath := path.Join(pkg.Dir, "__"+name)
```
nevkontakte, line 230: Let's use `gopherjs__` prefix to make sure there is absolutely no chance of a name clash with upstream files. 

Also IIRC Go build tooling treats files with names starting with `_` as special (for example, such directories will not be considered as packages), so I would avoid using such prefixes just in case it breaks some deeply baked-in assumptions in one if the `go/...` libraries.

compiler/compiler_test.go
```diff
@@ -0,0 +1,150 @@
...
+	for i, j := 0, len(sourceFiles)-1; i < j; i, j = i+1, j-1 {
+		sourceFiles[i], sourceFiles[j] = sourceFiles[j], sourceFiles[i]
+	}
+
+	outputReversed, err := compile(path, sourceFiles, minify)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if string(outputNormal) != string(outputReversed) {
```
nevkontakte, line 67: We are already using https://github.com/google/go-cmp in other tests, let's stick to it. https://github.com/gopherjs/gopherjs/blob/be292294a6971bc1df34f570fc9d174322ab12bc/build/context_test.go#L51-L53 has an idiomatic example of using that package.

nevkontakte (approved):

Issue #620, opened 2017-04-07
panic: nil package for unexported qualified name


fiatjaf:
I'm seeing this error when building a package that imports another package. It uses `--tags`, don't know if it is related somehow.

The error: https://pastebin.com/raw/5RSvSQF9
My versions:

```
fiatjaf@cantillon ~/c/g/s/g/f/summuladb> gopherjs version
GopherJS 1.8-1
fiatjaf@cantillon ~/c/g/s/g/f/summuladb> go version
go version go1.8 linux/amd64
```

To (try to) reproduce, download the code from https://github.com/fiatjaf/summuladb/tree/ca927b4878876cb184118de8227382cfb16d31dc and run `gopherjs build --tags=levelupjs` on it.

But it is much stranger than that: basically the error vanishes and the package compiles perfectly as soon as I edit [`qualifiedName`](https://github.com/gopherjs/gopherjs/blob/master/third_party/importer/export.go#L318-L327) to include a `log.Print()` statement, which I did to start debugging. After a `go get` the error stopped happening and the logged things were being logged to the terminal, as expected. I then removed the `log.Print()` line and ran `go get` again. Still no errors.

An hour later the error appeared again in the meantime I had run `gopherjs build` multiple times in my package, without a problem, and didn't edit or ran anything else. Again, the error vanished after I did the `log.Print()` trick. This comeback keep happening every once in a while.

See more in the discussion starting at https://gophers.slack.com/archives/C039C0R2T/p1491492286272286.

flimzy:
This may not be of any help, but when I find myself chasing phantom errors like this in GopherJS, I've usually discovered that it's the result of a caching bug (See #559). If you clear the GopherJS cache before each test, do you still get errors?  Do they become consistent?

    rm -rf ${GOPATH}/pkg/*_js

fiatjaf:
Solved it!

shurcooL:
It's good to hear that clearing the cache fixed it.

It'd still be nice to resolve this, so the problem cannot occur.

I can't immediately tell if it's identical to #559, but it's at least related.

The panic happens in [`github.com/gopherjs/gopherjs/third_party/importer`](https://godoc.org/github.com/gopherjs/gopherjs/third_party/importer) package, which is code we don't control directly. It's a copy of an older package from x/tools, that was deprecated during various changes done there. It might be worth it to check if equivalent functionality is available elsewhere in x/tools by now that GopherJS can be updated to use instead.

[...]

Issue #598, opened 2017-02-26
Building tags not work as expected for `gopherjs build`


oskca:
In my VueJS binding project [Vuejs](github.com/oskca/gopherjs-vue/), I tried
to include different version of `Vue Javascript Code` using go building tags,
but the generated js code does not change with or without building tags.

I hava tow files to switch tags: inc_debug.go
```go
//+build debug

package vue

//using vue-2.1.10.inc.js 219K
import _ "github.com/oskca/gopherjs-vue/jscode/debug" 
```
and inc_minified.go
```go
//+build !debug

package vue

//using vue-2.1.10.min.inc.js 71K
import _ "github.com/oskca/gopherjs-vue/jscode/debug" 
```

## Steps to show the problem

0. RM to clean the build
```
$ rm -rf $GOPATH/pkg/linux_js/github.com/oskca
```
1. A refresh build works with expected package building sequence
```
$ gopherjs build -v --tags debug $GOPATH/src/github.com/oskca/gopherjs-vue/examples/basic/basic.go
github.com/oskca/gopherjs-dom
github.com/oskca/gopherjs-json
github.com/oskca/gopherjs-vue/jscode/debug
github.com/oskca/gopherjs-vue
main
```
2. And outputs a debug version
```
$ lh $GOPATH/src/github.com/oskca/gopherjs-vue/examples/basic/basic.js
-rw-rw-r-- 1 oskca oskca 340K 2月  26 22:56 basic.js ## bigger file size
```

3. The second build without the building flags
```
$ gopherjs build -v $GOPATH/src/github.com/oskca/gopherjs-vue/examples/basic/basic.go
github.com/oskca/gopherjs-vue/jscode/minified ## expected
main
```
4. Outputs not change, still the debug version
```
$ lh $GOPATH/src/github.com/oskca/gopherjs-vue/examples/basic/basic.js
-rw-rw-r-- 1 oskca oskca 340K 2月  26 22:57 basic.js ## size not change
```

## If you swap setp 3 and step 2

the generated js code would only include the `vue-2.1.10.min.inc.js`
```
$ lh $GOPATH/src/github.com/oskca/gopherjs-vue/examples/basic/basic.js
-rw-rw-r-- 1 oskca oskca 192K 2月  26 23:10 basic.js ## smaller one
```

## I also tried building tags in *.inc.js, but it not work as #468

flimzy:
[...]

PR #1105, opened 2022-02-16
Improved build cache invalidation

### LINKED ISSUE #440: Wrongly reusing cached precompiled code when changing build tags

nevkontakte:
Previously GopherJS used to store its build cache under
$GOPATH/pkg/$GOOS_$GOARCH, and after Go Modules introduction some parts
of it were stored under os.UserCacheDir().

Starting from this commit, *all* build cache is located under
os.UserCacheDir(), in a manner similar to the modern Go tool. The cache
is keyed by a set of build options (such as minification, compiler
version, etc.) to ensure that incompatible archives aren't picked up
(see gopherjs#440 for example).

This change doesn't solve *all* possible cache-related issues (for
example, it still relies on timestamps to invalidate the cache, see
gopherjs#805), but should eliminate a large class of confusing failure
modes.

Fixes #440, #598, #620


build/build.go
```diff
@@ -654,40 +655,25 @@ func (s *Session) BuildPackage(pkg *PackageData) (*compiler.Archive, error) {
...
-
-	objFile, err := os.Create(pkgObj)
-	if err != nil {
-		return err
+// ImportResolverFor returns a function which returns a compiled package archive
+// given an import path. Import paths are resolved relative to the |pkg|.
+func (s *Session) ImportResolverFor(pkg *PackageData) func(string) (*compiler.Archive, error) {
+	return func(path string) (*compiler.Archive, error) {
+		if archive, ok := s.UpToDateArchives[path]; ok {
+			return archive, nil
```
flimzy, line 669: I'm probably just not following the logic here, but this appears to check a global cache for `path`, while the GoDoc suggests that `path` should be relative to pkg.

build/cache/cache.go
```diff
@@ -0,0 +1,142 @@
...
+
+// StoreArchive compiled archive in the cache. Any error inside this method
+// will cause the cache not to be persisted.
+func (bc *BuildCache) StoreArchive(a *compiler.Archive) {
+	if bc == nil {
+		return // Caching is disabled.
+	}
+	path := cachedPath(bc.archiveKey(a.ImportPath))
+	if err := os.MkdirAll(filepath.Dir(path), 0750); err != nil {
+		return
```
flimzy, line 98: Should these errors be surfaced anywhere? Even if only via a debug flag or similar?

build/cache/cache.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	f, err := os.Create(path)
```
flimzy, line 100: nit:  Should we open this file under a different/temporary name, and do an atomic* rename at the end?

I'm concerned that otherwise concurrent compilations (which may be quite common in certain IDE environments with compile/test-on-save type features) could unintentionally read a half-written file.

*where supported by the filesystem, anyway.

flimzy (approved):
A few comments, any of which can be done in a future PR, if decided. Looks good to me.

build/build.go
nevkontakte: Oops, that's a stale comment from an earlier iteration. Fixed.

build/cache/cache.go
nevkontakte: Done.

build/cache/cache.go
nevkontakte: Good point, we could use some debug logging in more than one place :) I'll do it in a separate PR.

PR #1114, opened 2022-04-17
Add flags to collect CPU and allocation profiles for GopherJS

nevkontakte:
This will be a helpful tool for performance issue investigations.

In order to implement this I also had to refactor use of `os.Exit()` in the main package, since it was preventing Cobra's post-run functions from executing.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -100,13 +100,13 @@ func main() {
 	cmdBuild.Flags().AddFlagSet(flagQuiet)
 	cmdBuild.Flags().AddFlagSet(compilerFlags)
 	cmdBuild.Flags().AddFlagSet(flagWatch)
-	cmdBuild.Run = func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


tool.go
```diff
@@ -169,10 +170,9 @@ func main() {
 				}
 				return nil
 			}()
-			exitCode := handleError(err, options, nil)
 
 			if s.Watcher == nil {
-				os.Exit(exitCode)
+				return err
```
flimzy, line 175: This is may be correct, but it looks odd.  Checking for `s.Watcher == nil` instead of `err != nil`?

flimzy (approved):

tool.go
nevkontakte: I think the idea is that if the watch mode is activated, we don't want to exit upon compiler error, but simply wait for another iteration to happen. However, your comment highlighted another flaw in my code, which I'll fix shortly.

Issue #1092, opened 2021-12-15
Add Go 1.18 support


flimzy:
[Go 1.18 beta 1 is out](https://go.dev/blog/go1.18beta1). That means it's time to start working toward adding support to GopherJS.

 - [x] Compile and execute fuzz targets like normal tests. https://github.com/gopherjs/gopherjs/pull/1132
 - [x] ~Re-enable `encoding/xml` tests once generics are supported.~ Moved to https://github.com/gopherjs/gopherjs/issues/1013.
 - [x] Generate explicit errors when encountering typeparams.
 - [x] Cleanup overlays that are not necessary in 1.18.
 - [x] `syscall/js`: Remove `Wrapper` interface.

nevkontakte:
Was there no RC this time? In any case, it's time indeed. I propose to separate generics support from the rest of 1.18, since this is a large feature that will require quite a bit of design and implementation effort.

Skimming through the [draft release notes](https://tip.golang.org/doc/go1.18), I note the following:

 - Fuzz support would be fancy, but realistically it is quite far out of GopherJS scope. I don't think we need to implement it.
 - There is a number of new `reflect` methods, which we will have to support.
 - `Wrapper` interface was removed from `syscall/js`.

nevkontakte:
I've created a new [go1.18-wip](https://github.com/gopherjs/gopherjs/tree/go1.18-wip) branch to do the work in.

Here's an easy recipe to start working on 1.18 support:

```
$ go install golang.org/dl/go1.18beta1@latest
$ go1.18beta1 download
$ go1.18beta1 install -tags gopherjsdev github.com/gopherjs/gopherjs &&  gopherjs test -v --short ./tests -p=1
```

nevkontakte:
My [go1.18-wip](https://github.com/nevkontakte/gopherjs/tree/go1.18-wip) branch now has complete Go 1.18 compatibility with the exception of generics, which brings us to the question of our approach to the release.

I plan to start working on generics at some point soon, but this is a large and complicated feature and will take a while to get right. The big question is whether to block GopherJS 1.18 release until that work is completed or not. 

The main argument in favor of delaying the release is that generics is arguably the most important feature in Go 1.18 and it would be highly confusing to release GopherJS 1.18 without it; also generics seem like too big a feature to release in a patch-release of 1.18.x.

At the same time there are multiple arguments against postponing the release:

[...]

PR #1120, opened 2022-06-07
Various standard library and compiler fixes to support Go 1.18

nevkontakte:
This set of changes resolves most (although not all) standard library test failures found by the CI. The list of changes include:

 - reflect: Provide stubs for string-optimized map access methods used by reflection.
 - prelude: correctly handle slice to array pointer conversion when the array is shorter than the original slice.
 - net/netip: do not use `internal/intern` optimization, which relies on unsupported unsafe features.
 - strings: support `Clone()` method.
 - math: use JavaScript `NaN` literal as a `math.NaN()` return value to address errors in the `internal/fuzz` package tests.
 - net/http: update a reference to a renamed variable.

Additionally this PR improves compiler panic handling that helped me debug some of the issues in this PR and temporarily excludes `encoding/xml` package from the test set, since it uses generics in tests. The package itself, however, works correctly and we will bring the tests back once generics support is implemented.

Updates #1092.


paralin:
Seems to build & run fine for me, thanks for your work on this.

PR #1122, opened 2022-06-08
Improve nodejs stack size limit heuristic.

### LINKED ISSUE #470: "Maximum call stack size exceeded" errors get swallowed

nevkontakte:
Setting stack size exactly to the system limit will cause node to
segfault before it has a chance to notice that it ran out of space. In
my experiemnts values closer than 64 KiB may still cause a segfault from
time to time.

I also changed the multiplier to 1024, since that's what v8 actually
uses for the flag value.

Fixes #470.

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -891,9 +891,11 @@ func runNode(script string, args []string, dir string, quiet bool, out io.Writer
 		// -	OS process limit
 		// -	Node.js (V8) limit
 		//
-		// GopherJS fetches the current OS process limit, and sets the
-		// Node.js limit to the same value. So both limits are kept in sync
-		// and can be controlled by setting OS process limit. E.g.:
+		// GopherJS fetches the current OS process limit, and sets the Node.js limit
+		// to a value slightly below it (otherwise nodejs is likely to segfault).
+		// The backoff size has been determined experimentally on a linux machine,
+		// so it may not be 100% reliable. So both limits are kept in sync and can
+		// be controlled by setting OS process limit. E.g.:
 		//
 		// 	ulimit -s 10000 && gopherjs test
 		//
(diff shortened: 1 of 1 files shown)


tool.go
```diff
@@ -891,17 +891,23 @@ func runNode(script string, args []string, dir string, quiet bool, out io.Writer
...
 		// 	ulimit -s 10000 && gopherjs test
 		//
 		cur, err := sysutil.RlimitStack()
 		if err != nil {
 			return fmt.Errorf("failed to get stack size limit: %v", err)
 		}
-		allArgs = append(allArgs, fmt.Sprintf("--stack_size=%v", cur/1000)) // Convert from bytes to KB.
+		cur = cur / 1024 // Convert bytes to KiB.
+		if backoff := uint64(64); cur > backoff {
+			cur = cur - backoff
```
flimzy, line 908: This logic looks a bit funny, but I don't think it really matters.

If cur = 64, the result is 64.
If cur - 65, the result is 1.  (which will probably crash immediately)

I doubt anyone will ever run with such a small ulimit, but maybe a slightly better heuristic would be

```
if backoff := uint(64); cur > backoff*2 {
```

?

Not  a big deal either way, since as I said, I doubt anyone will be running under such tight limits.  We can always wait to see if we get a bug report before fixing it.

flimzy (approved):

tool.go
nevkontakte: TBH, I'm pretty sure this is a moot point, since such small stacks wouldn't allow for any useful programs anyway. However, a marginally more sensible solution could be to only apply the backoff when the value is above the default stack size limit (984K). WDYT?

PR #1132, opened 2022-07-31
Detect and execute fuzz targets as tests

nevkontakte:
This change adds support for discovering fuzz targets and executing them in unit test mode. The tests are executed with seed inputs only, since we don't have the `libfuzzer` infrastructure to actually perform fuzzing.

While at it, I also factored out test function discovery into its own internal package, which is a step towards decluttering `tool.go` and testing logic in it.

Updates #1092.


internal/srctesting/srctesting.go
```diff
@@ -63,3 +65,16 @@ func ParseFuncDecl(t *testing.T, src string) *ast.FuncDecl {
 	}
 	return fdecl
 }
+
+// Format AST node into a string.
+//
+// The node type must be *ast.File, *printer.CommentedNode, []ast.Decl,
+// []ast.Stmt, or assignment-compatible to ast.Expr, ast.Decl, ast.Spec, or
+// ast.Stmt.
+func Format(t *testing.T, fset *token.FileSet, node any) string {
```
flimzy, line 74: I would expect a function like this to call `t.Helper()`, but maybe upstream doesn't for some reason?

internal/testmain/testmain.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+func (tm *TestMain) scanPkg(fset *token.FileSet, files []string, loc FuncLocation) error {
+	for _, name := range files {
+		srcPath := path.Join(tm.Package.Dir, name)
+		f, err := buildutil.OpenFile(tm.Package.InternalBuildContext(), srcPath)
+		if err != nil {
+			return fmt.Errorf("failed to open source file %q: %w", srcPath, err)
+		}
+		defer f.Close()
```
flimzy, line 95: Nit: Since deferred functions don't execute until the entire function returns, we could end up with a large number of open files here.  Probably not likely to be an actual problem, unless some package has thousands of files in it. Do we care? :shrug:

internal/testmain/testmain.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	if sel, ok := ptr.X.(*ast.SelectorExpr); ok && sel.Sel.Name == "M" {
+		return true
+	}
+	return false
+}
+
+// isTest tells whether name looks like a test (or benchmark, according to prefix).
+// It is a Test (say) if there is a character after Test that is not a lower-case letter.
+// We don't want TesticularCancer.
```
flimzy, line 242: Understatement of the year. :laughing:

internal/testmain/testmain_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/gopherjs/gopherjs/build"
+	"github.com/gopherjs/gopherjs/internal/srctesting"
+	. "github.com/gopherjs/gopherjs/internal/testmain"
+)
+
+const inPackage = ``
+const external = ``
```
flimzy, line 16: I don't see these constants referenced anywhere. Am I loverlooking something?

internal/srctesting/srctesting.go
nevkontakte: Nope, I just forgot to do it.

internal/testmain/testmain.go
nevkontakte: I don't think we do until proven otherwise. That would have to be a single package with *hundreds* of test source files (not just any source files), which I don't think I've seen. But if we get a bug report like that, I promise to fix this :)

internal/testmain/testmain_test.go
nevkontakte: Oops,  that's a leftover from an early iteration. Removed.

internal/testmain/testmain.go
nevkontakte: Can't say I am a fan of this comment, but I decide to preserve it as a historical artifact ¯\_(ツ)_/¯

flimzy (approved):

Issue #1135, opened 2022-08-04
len and range do not scale well for map types.


tomconnell-wf:
I'm using go 1.17.3 and gopherjs 1.17.1+go1.17.3

I have noticed that calling `len()` or `range` on a map is implemented in JS to call length on `$keys`.  

`var $keys = function(m) { return m ? Object.keys(m) : []; };`

This appears to scale linearly as a map grows in size.

Consider the following benchmarks:

```go
import "testing"

func makeMap(size int) map[int]string {
	myMap := make(map[int]string, size)
	for i := 0; i < size; i++ {
		myMap[i] = `data`
	}

	return myMap
}

func makeInt(size int) []int {
	slice := make([]int, size)
	for i := 0; i < size; i++ {
		slice[i] = i
	}

	return slice
}

func BenchmarkSliceLen(b *testing.B) {
	slice := makeInt(1000000)

	for i := 0; i < b.N; i++ {
		if len(slice) > 0 {
		}
	}
}

func BenchmarkMapLen(b *testing.B) {
	myMap := makeMap(1000000)

	for i := 0; i < b.N; i++ {
		if len(myMap) > 0 {
		}
	}
}

func BenchmarkMapNilCheck(b *testing.B) {
	myMap := makeMap(1000000)

	for i := 0; i < b.N; i++ {
		if myMap != nil {
		}
	}
}

func BenchmarkMapNilElementCheck(b *testing.B) {
	myMap := makeMap(1000000)

	for i := 0; i < b.N; i++ {
		if myMap[0] != `` {
		}
	}
}
```

In Go:
```
✗ go test ./... -benchmem -bench "^Benchmark.*"
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkSliceLen-12              	1000000000	         0.2519 ns/op	       0 B/op	       0 allocs/op
BenchmarkMapLen-12                	1000000000	         0.5852 ns/op	       0 B/op	       0 allocs/op
BenchmarkMapNilCheck-12           	1000000000	         0.5832 ns/op	       0 B/op	       0 allocs/op
BenchmarkMapNilElementCheck-12    	162799234	         6.268 ns/op	       0 B/op	       0 allocs/op
PASS
```

In Gopherjs:
```
✗ GOOS=linux gopherjs test ./... --bench="Benchmark.*"
goos: linux
goarch: js
[...]

PR #1136, opened 2022-08-05
Use a Javascript Map for Go Maps.  Improves performance of len() calls by orders of magnitude. 🗺️ 

tomconnell-wf:
# Overview
https://github.com/gopherjs/gopherjs/issues/1135

The performance of len() on maps brought me here, because it would call js `Object.keys(myMap).Length`.  It was many orders of magnitude slower than len() in Go.  It was a pitfall that those writing idiomatic go would fall into.

This PR switches the backing implementation of Maps in GopherJS to [ECMAScript 2015 Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).  These maps provide accounting so that Map.size can be called to make len() fast.  I was hopeful that iterating maps would be faster, also, because the len of the map is used, but unfortunately that is not the case.  However, it should be trivial to change the loop implementation to use `[Map.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries) when GopherJS adopts ECMAScript 2017.  

One thing to note:  ES 2015 Maps do not return in random order.  There is a pitfall, the I believe is acceptable, that someone could write GopherJS code that relies on the order of elements in a map, that does not work when run in the Go runtime.

Using the benchmarks committed as part of this PR, with a size of 10000 elements:
**Master**
```
✗ GOOS=linux gopherjs test ./... --bench="Benchmark.*"
Using GOOS=linux and GOARCH=ecmascript in GopherJS is deprecated and will be removed in future. Use GOOS=js GOARCH=ecmascript instead.
goos: js
goarch: ecmascript
BenchmarkSliceLen           	1000000000	         0.2600 ns/op
BenchmarkMapLen             	    9886	    115719 ns/op
BenchmarkMapNilCheck        	1000000000	         0.5070 ns/op
BenchmarkMapNilElementCheck 	144578312	         8.328 ns/op
BenchmarkSliceRange         	  226414	      5070 ns/op
BenchmarkMapRange           	    6315	    183215 ns/op
PASS
ok  	benchmark/tom	7.682s
```

**This Branch**
Map len is ~6 orders of magnitude faster.  Unfortunately, it appears element access and ranging has gotten ~20% worse, and ranging ~3% worse
```
✗ GOOS=linux gopherjs test ./... --bench="Benchmark.*"
Using GOOS=linux and GOARCH=ecmascript in GopherJS is deprecated and will be removed in future. Use GOOS=js GOARCH=ecmascript instead.
goos: js
goarch: ecmascript
BenchmarkSliceLen           	1000000000	         0.2820 ns/op
BenchmarkMapLen             	1000000000	         0.5100 ns/op
BenchmarkMapNilCheck        	1000000000	         0.2610 ns/op
BenchmarkMapNilElementCheck 	104008665	        10.50 ns/op
BenchmarkSliceRange         	  222222	      5094 ns/op
BenchmarkMapRange           	    6082	    188096 ns/op
PASS
ok benchmark/tom	7.001s
```

# Testing
It seems like the tests cover many more cases than manual testing ever could.  I may have tunnel vision as the author of the PR, but I believe that passing CI tests should be considered good enough.


tomconnell-wf:
At this point, maps cannot be accessed by index, but their size can be acquired, and they can be iterated.

Using the same benchmark I used to report the slowness (with a million entries):
```
goos: linux
goarch: js
BenchmarkSliceLen           	1000000000	         0.2530 ns/op
BenchmarkMapLen             	1000000000	         0.5040 ns/op
BenchmarkMapNilCheck        	1000000000	         0.5050 ns/op
BenchmarkMapNilElementCheck 	144578312	         8.300 ns/op
PASS
```

The len check is as quick as I would hope.  (The setup of the map is very slow, and could possibly be improved (Because I'm making a slice of maps that then get converted in $makeMap, but I don't think that is any different than master.)

tomconnell-wf:
FWIW, this is with 1 million elements
```
goos: linux
goarch: js
BenchmarkSliceLen           	1000000000	         0.2560 ns/op
BenchmarkMapLen             	1000000000	         0.5920 ns/op
BenchmarkMapNilCheck        	1000000000	         0.6890 ns/op
BenchmarkMapNilElementCheck 	98797087	        10.13 ns/op
BenchmarkSliceRange         	    2221	    510581 ns/op
```

compiler/prelude/types.go
```diff
@@ -617,14 +617,15 @@ var $mapType = function(key, elem) {
...
-  }
-  return m;
+ //console.log("$makeMap input", keyForFunc, entries) 
+ var m = new Map();
+ for (var i = 0; i < entries.length; i++) {
+   var e = entries[i];
+   m.set(keyForFunc(e.k), e.v);
+ }
+ //console.log("$makeMap output", m)
+ return m;
```
flimzy, line 627: Nit: Indentation seems to have been changed here, which makes discerning the meaningful changes a bit harder.

flimzy (commented):
Thanks for working on this! Looks quite promising. I look forward to seeing the rest of the implementation.

compiler/expressions.go
```diff
@@ -474,10 +474,23 @@ func (fc *funcContext) translateExpr(expr ast.Expr) *expression {
...
-			}
-			return fc.formatExpr(`(%1s = %2e[%3s], %1s !== undefined ? %1s.v : %4e)`, fc.newVariable("_entry"), e.X, key, fc.zeroValue(t.Elem()))
+				return fc.formatExpr(
+					`(%1s = typeof %2e.get === "function" ? %2e.get(%3s) : undefined, %1s !== undefined ? [%1s, true] : [%4e, false])`,
+					entryVar,
+					e.X,
+					key,
+					fc.zeroValue(t.Elem()),
+				)
+			}
```
tomconnell-wf, line 486: Sorry if this goes against preferred formatting.  My brain was unable to parse the long line and multiple arguments.  I can put it back to single line, if preferred.

go.mod
```diff
@@ -12,15 +12,19 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.7.0
```
tomconnell-wf, line 15: If the maintainers are against a new dependency, I will update the new tests to manually assert conditions.  I am a big fan of testify, though.  Would recommend!  😁

tests/map_test.go
```diff
@@ -0,0 +1,80 @@
+package tests
```
tomconnell-wf, line 1: I didn't see anything that existed like this.  This should give an okay exercise of the map api and built-in functions that operate on maps.  If I understand CI right, these should get run by `gopherjs test`.

go.mod
flimzy: I would shy away from this dependency, not per se because it's a new dependency, but because it goes against established Go guidelines to use an assertion library, and I think it's valuable to try to follow Go guidelines, since we're kind of a cousin project. 

https://github.com/golang/go/wiki/TestComments#assert-libraries

But I'm interested to hear if @nevkontakte feels differently.

tomconnell-wf: Fair enough, I won't make waves.

go.mod
```diff
@@ -20,7 +20,6 @@ require (
 
 require (
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
-	github.com/shurcooL/vfsgen v0.0.0-20200824052919-0d455de96546 // indirect
```
tomconnell-wf, line 23: `go mod` is about 90% of a mystery to me.  I ran `go mod tidy`, and here we are.

tomconnell-wf:
Well, I think I am almost there.  It looks like I have a Sprintf problem to fix, and then done?

go.mod
flimzy: This one bites me all the time, too.  I think running `go generate ./...` re-adds this.

We should probably explicitly add this dependency somewhere in the main code, to prevent this from happening. :thinking:

tests/map_test.go
flimzy: I think these tests are good.  While most/all of these changes may also be exercised by other (stdlib) tests, I think it's good to have explicit tests, to help pinpoint if a failure ever occurs.

I'm not sure of the best place for them to live, though.  Here is probably as good a place as any, unless @nevkontakte  has another idea.  It's always easy to move tests if we want to in the future.

compiler/expressions.go
flimzy: My 2¢: This seems fine, and more readable than before.

tomconnell-wf:
As you can see in CI, there are some fmt stdlib tests that are failing.  It seems like it is because I didn't wire something up to reflect/reflectlite properly.  

Here's one of the tests that are failing.
https://cs.opensource.google/go/go/+/master:src/fmt/example_test.go;l=295-300?q=dach&ss=go%2Fgo

The expected output is
```
map[dachshund:false peanut:true] map[string]bool{"dachshund":false, "peanut":true}
```

The incorrect output is 
```
map[] map[string]bool{}
```

I'm a bit out of my depth here, but it seems like this is because the keys and values aren't reflecting properly.  My gut says it has something to do with the change from e to e.v in https://github.com/gopherjs/gopherjs/pull/1136/files#diff-70e5ca5fb5a75cb3a234c5a2caee3036426e4ffbb46a9e6b2ee556cb244eccbcR623.  I'll keep taking cracks at it, but if the maintainers have a hunch, I could use the guidance.

tomconnell-wf:
Or maybe it has something to do with `$internalize` and `$externalize`...?

tomconnell-wf:
I'm going with the theory that by doing [this](https://github.com/gopherjs/gopherjs/pull/1136/files#diff-70e5ca5fb5a75cb3a234c5a2caee3036426e4ffbb46a9e6b2ee556cb244eccbcR623) I clobbered the information needed for reflection to work properly.  I'll massage it so that we can use the Map as a POJSO, with the reflection type info...I think.

compiler/expressions.go
nevkontakte: I have no objections either :)

compiler/statements.go
```diff
@@ -211,19 +211,23 @@ func (fc *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
...
-			fc.translateLoopingStmt(func() string { return iVar + " < " + keysVar + ".length" }, s.Body, func() {
+			fc.Printf("%s = %s ? %s.keys() : [];", keysVar, refVar, refVar)
+
+			fc.translateLoopingStmt(func() string { return iVar + " < " + refVar + ".size" }, s.Body, func() {
+				keyVar := fc.newVariable("_key")
 				entryVar := fc.newVariable("_entry")
-				fc.Printf("%s = %s[%s[%s]];", entryVar, refVar, keysVar, iVar)
+				fc.Printf("%s = %s.next().value;", keyVar, keysVar)
+				fc.Printf("%s = %s.get(%s);", entryVar, refVar, keyVar)
+				//fc.Printf("console.log('loop', %s, %s, %s, %s, %s)", refVar, keyVar, entryVar, keysVar, iVar)
```
nevkontakte, line 221: Nit: looks like a debugging statement left here by accident?

go.mod
nevkontakte: `go generate ./...` will fail and complain about attempting to use an unknown module. https://github.com/gopherjs/gopherjs/blob/master/compiler/natives/doc.go is the place which uses this package.

I think the official recommendation it to create a source `tools.go` somewhere in the source tree with something like this:

```go
//go:build ignore
package tools

import (
  _ "github.com/shurcooL/vfsgen"
)
```

The build constraint makes sure  it doesn't actually get build, but it's enough for `go mod tidy` to recognize the dependency as legit and keep it. I've been meaning to do this for a while, but haven't gotten around to it. Feel free to do this :)

tests/map_test.go
nevkontakte: Thanks @tomconnell-wf for adding these tests, historically gopherjs test coverage has been somewhat spotty and relied a lot on standard library tests (like @flimzy said). Standard library tests indeed cover most of language use cases, but if something breaks, it makes for a complicated debugging session. So having specific tests is very valuable and welcome.

Placing tests here is okay too. The `tests` package has gotten somewhat out of hand though, and we should reorganize it into several subpackages, but that's out of scope for this PR.

tests/map_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"strings"
+	"testing"
+)
+
+// These tests exercise the api of maps and built-in functions that operate on maps
+func Test_MapLiteral(t *testing.T) {
```
nevkontakte, line 9: For completeness you could also add a test case with this:

```go
	myMap := map[string]int{}
	myMap[`test`] = 0
	myMap[`key`] = 1
	myMap[`charm`] = 2
```

nevkontakte:
@tomconnell-wf I've looked through the diff and there is nothing obviously incorrect in it I could pin the test failure on. My best guess is that somewhere in the `reflect` package we read the `$keys().length` value, which is now 0, and so iteration is not happening properly. I am not sure I will have time for a deep dive until the weekend, but if you don't find the bug until then, I'll give it a shot.

As a side note, I recommend you rebasing on top of the latest master, which includes Go 1.18 support. There have been some additions to the `reflect` map-related APIs in 1.18 which may be impacted by your change or will cause merge conflicts when you are ready.

flimzy:
> I think the official recommendation it to create a source tools.go somewhere

We can go one step further, and put it in a subdir so that its in its own package that is never even parsed during compilation (but is parsed by `go mod`)

tomconnell-wf:
This may be considered phoning it in, but I can't figure out the reflection.  As best I can understand, it's built around the keys and the elements being available on the javascript object.  Rather than having a special case for Map types, and since these are all references anyway, I am thinking of leaving the key and element as they were before, but also putting them into the javascript Map api, for faster size calls.  I will push commits for this later today, and some javascript heap snapshots to make sure memory usage isn't significantly higher.

nevkontakte:
I would strongly prefer not to do this, not for performance, but for maintenance reasons. A lot of our work on gopherjs over the past year has been dedicated to simplifying and documenting the codebase, and I would very much prefer to help you figure out the reflection side of the problem over adding more cruft that will be confusing people in future.

tomconnell-wf:
Side note - when it is decided that ECMA Script 2017 is supported, we can use `.entries()` instead of `.keys()` on the map range statement, and it'll be about 2x faster.  
https://github.com/gopherjs/gopherjs/commit/cf9a8fa87bb80f90e95df2c21e13a50e85d96d49

10000 elements:

**Before**
✗ GOOS=linux gopherjs test ./... --bench="Benchmark.*"
Using GOOS=linux and GOARCH=ecmascript in GopherJS is deprecated and will be removed in future. Use GOOS=js GOARCH=ecmascript instead.
goos: js
goarch: ecmascript
BenchmarkSliceLen           	1000000000	         0.2520 ns/op
BenchmarkMapLen             	1000000000	         0.2490 ns/op
BenchmarkMapNilCheck        	1000000000	         0.5010 ns/op
BenchmarkMapNilElementCheck 	363636363	         3.311 ns/op
BenchmarkSliceRange         	  222222	      4991 ns/op
BenchmarkMapRange           	    5454	    186285 ns/op
PASS

**After**
✗ GOOS=linux gopherjs test ./... --bench="Benchmark.*"
Using GOOS=linux and GOARCH=ecmascript in GopherJS is deprecated and will be removed in future. Use GOOS=js GOARCH=ecmascript instead.
goos: js
goarch: ecmascript
BenchmarkSliceLen           	1000000000	         0.2440 ns/op
BenchmarkMapLen             	1000000000	         0.2410 ns/op
BenchmarkMapNilCheck        	1000000000	         0.4830 ns/op
BenchmarkMapNilElementCheck 	377358489	         3.246 ns/op
BenchmarkSliceRange         	  222222	      4847 ns/op
BenchmarkMapRange           	   13656	     83553 ns/op
PASS

tomconnell-wf:
> I would strongly prefer not to do this, not for performance, but for maintenance reasons
I spent most of the day yesterday trying to figure it out.  I'll give it another shot.  Maybe having slept  on it will unlock something.

tomconnell-wf:
Current state of performance.  Range is a bit worse?!?  :(

10000 elements

Master:
```
✗ GOOS=linux gopherjs test ./... --bench="Benchmark.*"
Using GOOS=linux and GOARCH=ecmascript in GopherJS is deprecated and will be removed in future. Use GOOS=js GOARCH=ecmascript instead.
goos: js
goarch: ecmascript
BenchmarkSliceLen           	1000000000	         0.2500 ns/op
BenchmarkMapLen             	   10774	    107203 ns/op
BenchmarkMapNilCheck        	1000000000	         0.4900 ns/op
BenchmarkMapNilElementCheck 	150753768	         8.026 ns/op
BenchmarkSliceRange         	  230768	      4853 ns/op
BenchmarkMapRange           	    6315	    164054 ns/op
PASS
ok  	github.com/Workiva/doc_plat_client/tool/internal/gopherjsbenchmarks/benchmark/tom	7.427s
```


This commit:
```
✗ GOOS=linux gopherjs test ./... --bench="Benchmark.*"
Using GOOS=linux and GOARCH=ecmascript in GopherJS is deprecated and will be removed in future. Use GOOS=js GOARCH=ecmascript instead.
goos: js
goarch: ecmascript
BenchmarkSliceLen           	1000000000	         0.2440 ns/op
BenchmarkMapLen             	1000000000	         0.2430 ns/op
BenchmarkMapNilCheck        	1000000000	         0.4840 ns/op
BenchmarkMapNilElementCheck 	357142856	         3.214 ns/op
BenchmarkSliceRange         	  218181	      4968 ns/op
BenchmarkMapRange           	    6458	    183958 ns/op
PASS
ok  	github.com/Workiva/doc_plat_client/tool/internal/gopherjsbenchmarks/benchmark/tom	6.549s
```

tomconnell-wf:
Those numbers make me think I could change tack and just account for the size of the map during creating, assign and delete.  That seems semantically less correct, through. 🤔

compiler/natives/doc.go
```diff
@@ -5,5 +5,9 @@
 // in src subfolder.
 package natives
 
+import (
+	_ "github.com/shurcooL/vfsgen" // Force go.mod to require this package
+)
+
```
tomconnell-wf, line 11: I tried it with the build tag and an unused directory, and go.mod didn't seem to pick it up.  This seemed like a good spot.

nevkontakte:
I haven't tested it, but I'm pretty sure this is the like you need to modify to make `reflect` work: https://github.com/gopherjs/gopherjs/blob/fa8544dfc7cb6b7b2e18dc7a67976e77e47b5efc/compiler/natives/src/reflect/reflect.go#L711

it probably should be something like `keys: js.InternalObject(m).Call("keys"),`.

https://github.com/gopherjs/gopherjs/blob/fa8544dfc7cb6b7b2e18dc7a67976e77e47b5efc/compiler/natives/src/reflect/reflect.go#L757 and https://github.com/gopherjs/gopherjs/blob/fa8544dfc7cb6b7b2e18dc7a67976e77e47b5efc/compiler/natives/src/reflect/reflect.go#L1382 are two other places.

tomconnell-wf:
Is there something special I need to do to include reflect changes?  If I change reflect.go or reflectlite.go, then `go install` and then build/serve with gopherjs, the javascript output doesn't include my changes.  This may be what I have been struggling with today and yesterday.

flimzy:
> Is there something special I need to do to include reflect changes?

I believe this is probably what you're looking for: https://github.com/gopherjs/gopherjs/wiki/Developer-Guidelines#code-generation

tomconnell-wf:
Yeah, I just discovered that vfsgendev wasn't working right.  🤦🤦🤦

tomconnell-wf:
That'll make things easier.

tomconnell-wf:
I don't really understand the whole `keysFor` construct, but I figure that's been because I've been dealing with strings and ints; probably needed for reliable keys for structs and such?

Anyway, now I can see the failing fmt test mapiter shape, I think I can fix it up.
Master
image

This branch
image

compiler/statements.go
```diff
@@ -703,19 +703,15 @@ func (fc *funcContext) translateAssign(lhs, rhs ast.Expr, define bool) string {
...
-				fc.translateExpr(l.X),
+				`%s = %s; (%s || $throwRuntimeError("assignment to entry in nil map")).set(%s.keyFor(%s), { k: %s, v: %s })`,
+				keyVar,
+				fc.translateImplicitConversionWithCloning(l.Index, t.Key()),
 				fc.translateExpr(l.X),
 				fc.typeName(t.Key()),
 				keyVar,
 				keyVar,
-				valueVar,
+				fc.translateImplicitConversionWithCloning(rhs, t.Elem()),
```
tomconnell-wf, line 714: This should fix the Issue 21687 test failure.  

Consider 
```
func maps() (int, int) {
	one := map[int]int{0: 1}
	two := map[int]int{0: 2}

	x := one
	x[0] += func() int {
		x = two
		return 0
	}()

	return one[0], two[0]
}
```

With the way that I had changed it, one was getting set to two.  Yikes.

tomconnell-wf:
I hope CI doesn't cost much.  I think it may actually pass now.

```
➜  gopherjs git:(javascript_map_impl) ✗ gopherjs test  --short fmt log os ./tests

PASS
ok  	fmt	1.170s
PASS
ok  	log	0.444s
PASS
ok  	os	1.447s
PASS
ok  	github.com/gopherjs/gopherjs/tests	6.368s
```

tomconnell-wf:
(no need to run CI) I'm grappling with some reflect.DeepEquals problems right now.  I think I saw somewhere in master that empty maps evaluate to false, but in this branch, they do not.  I'm out of time for the week, pretty much, so I just wanted to communicate that.

nevkontakte:
@tomconnell-wf no worries, and thanks for your work. This isn't an easy change, but it will be a valuable one.

I will probably have some spare time this weekend, so I'll take a look and see if I can help you with debugging.

tomconnell-wf:
There may be more problems, but the current problem with archive/tar tests, for example, is default values and len().  

```
println(paxHdrs, v.paxHdrs)
println(len(paxHdrs)==0, len(v.paxHdrs)==0);
```

```
MASTER
{} false
true true
PASS
ok  	archive/tar	1.012s
```
```
BRANCH
        got  map[]
        want map[]
Map(0) {} false
0 true
```

tomconnell-wf:
It's some sort of inference problem.  

paxHeaders is a map[string]string (not that it matters)

`len(paxHdrs) == 0` equals 0

```
i2 := len(paxHdrs)
i2 == 0 // equals true
```

tomconnell-wf:
This is the test case (I'm going to have to go shortly, so I'm getting chatty about where I am at)

```
func Test_MapLenBool(t *testing.T) {
	m := make(map[string]string)

	if len(m) != 0 { 
                // this case is hit
		t.Error(`inline len should have been 0`)
	}

	i := len(m)

	if i != 0 { 
                // this case is NOT hit
		t.Error(`assigned len should have been 0`)
	}
}
```

nevkontakte:
@tomconnell-wf thanks for posting the test case, it made it very easy to reproduce and find the issue. I pushed the fix into your branch.

The Go expression `len(m) == 0` compiles into `!((m ? m.size : 0 === 0))`. Apparently, `===` operator has higher priority than `?:`, so it was actually evaluated as `!((m ? m.size : (0 === 0)))` → `!((true ? 0 : true))` → `!0` → `true`. Adding brackets around the expression fixed the issue.

compiler/natives/doc.go
nevkontakte: SGTM.

compiler/natives/src/reflect/reflect.go
```diff
@@ -705,10 +706,20 @@ func (iter *hiter) skipUntilValidKey() {
 }
 
 func mapiterinit(t *rtype, m unsafe.Pointer, it *hiter) {
+	mapObj := js.InternalObject(m)
+	//var keys *js.Object
```
nevkontakte, line 710: I assume this was left by accident?

nevkontakte:
I fixed a couple more bugs and all `reflect` tests are passing now. Hopefully the remaining few won't be too difficult to weed out.

nevkontakte:
I fixed a couple more bugs and all `reflect` tests are passing now. Hopefully the remaining few won't be too difficult to weed out.

tomconnell-wf:
I'll take a look today

compiler/natives/src/reflect/reflect.go
```diff
@@ -623,12 +623,12 @@ func makemap(t *rtype, cap int) (m unsafe.Pointer) {
...
 }
 
-func keyFor(t *rtype, key unsafe.Pointer) (*js.Object, string) {
+func keyFor(t *rtype, key unsafe.Pointer) (*js.Object, *js.Object) {
 	kv := js.InternalObject(key)
 	if kv.Get("$get") != js.Undefined {
 		kv = kv.Call("$get")
 	}
-	k := jsType(t.Key()).Call("keyFor", kv).String()
+	k := jsType(t.Key()).Call("keyFor", kv)
```
tomconnell-wf, line 631: Wow, good call.  I feel like being able to remove things like this is a nice side-benefit of this PR, as hard as it was to find them.

compiler/expressions.go
```diff
@@ -952,7 +952,7 @@ func (fc *funcContext) translateBuiltin(name string, sig *types.Signature, args
 		case *types.Pointer:
 			return fc.formatExpr("(%e, %d)", args[0], argType.Elem().(*types.Array).Len())
 		case *types.Map:
-			return fc.formatExpr("%e ? %e.size : 0", args[0], args[0])
+			return fc.formatExpr("(%e ? %e.size : 0)", args[0], args[0])
```
tomconnell-wf, line 955: 🙇

compiler/natives/src/reflect/reflect.go
tomconnell-wf: Oops, yep

tomconnell-wf:
Alright, tests are all passing now.  I looked over the full changeset, and it looks not half bad to me.  Ready for review.

nevkontakte:
@tomconnell-wf thanks for doing this work! I spotted one potential issue and I should have time to do a careful review in the next day or two. Because maps are such a fundamental part of the language, I'd like to exercise some extra caution in merging this change to make sure we don't break anyone.

Another request would be to rebase on top of the latest master (if you haven't already) and squash the commits into one or two with a detailed commit message, explaining the semantics of the change and reasons behind it. This will both help future contributors in understanding the history and ensure the change is easy to revert if we discover some kind of a critical bug.

compiler/expressions.go
```diff
@@ -969,7 +981,12 @@ func (fc *funcContext) translateBuiltin(name string, sig *types.Signature, args
 	case "delete":
 		args = fc.expandTupleArgs(args)
 		keyType := fc.pkgCtx.TypeOf(args[0]).Underlying().(*types.Map).Key()
-		return fc.formatExpr(`delete %e[%s.keyFor(%s)]`, args[0], fc.typeName(keyType), fc.translateImplicitConversion(args[1], keyType))
+		return fc.formatExpr(
+			`(typeof %1e.delete === "function" && %1e.delete(%2s.keyFor(%3s)))`,
```
flimzy, line 985: When would delete not be a function? And what should we do in such a case?

tomconnell-wf:
I'd be happy to rebase.

compiler/expressions.go
tomconnell-wf: I don't know of a more graceful way of handling this, but it is possible to call `delete(nil)` in go and the javascript would wind up calling delete() on undefined.  I'm hand waving a bit based on what I saw.  I could make an explicit test case, if that would help?

flimzy: Ah. If that's the only edge case, then this is probably correct. Although perhaps checking for null/js.Undefined would be more explicit, if that's not too cumbersome.

tomconnell-wf: I was worried that there might be some situations where a JS object had been set, but it wouldn't acutally be a Map type, and an undefined check would still cause an error.  Perhaps that is an unfounded worry.

flimzy: That's the kind of corner case I was wondering about. If that were to happen, what would it actually mean?

tomconnell-wf:
My curiosity is afire wanting to know what the one potential issue might be. 🐱

compiler/expressions.go
tomconnell-wf: As long as the places that gopherjs compiler makes Maps are all using `new Map()`, I would expect things would be generally good.  AFAIK, I got all those.  The edge cases would operate outside of that object flow.  I'm not sure what sort of ways that could happen.  

Passing in a wrong type is a RTE.  Maybe there is a way to coerce a type so that delete() is satisfied, but the actual object is incorrect?  Maybe using `js.Global` to make an object and coerce that into a map to pass to delete?

nevkontakte:
One other potential source of corner cases we need to verify is map internalization/externalization (that is, conversions between `js.Object` and Go native types). I would start by checking that mappings at the top of https://pkg.go.dev/github.com/gopherjs/gopherjs/js still hold. Then there is also the matter of embedding `js.Object` into structs:

```go
type S structs {
  *js.Object
  Prop map[string]int `js:prop`
}
```

tomconnell-wf:
Good call

tomconnell-wf:
> One other potential source of corner cases we need to verify is map internalization/externalization 

What is the expectation on wrapping?  It seems to me that we would still want the new map implementation to wrap to a js.Object with map keys as properties, otherwise existing code will break.  In other words:

```
type myStruct struct {
  map[string]string myMap
}

ms := &myStruct{map[string]string{`key`: `value`}}

wrapper := js.MakeFullWrapper(ms)

// Old Behavior - I think we should keep
wrapper.Get(`key`) // value

// Use Map API, breaking change
wrapper.Call(`get`, `key`) // value
```

nevkontakte:
Yes, I think we should make sure this doesn't break existing code and that Go maps are correctly translated into objects and back.

tomconnell-wf:
How do those tests look to demonstrate correctness of wrapping/internalize/externalize?

tomconnell-wf:
Oh, node doesn't like that.  I'm not much of a Node guy.  I wonder how best to do that namespacing for both contexts.

nevkontakte:
@tomconnell-wf see my suggestion [here](https://github.com/gopherjs/gopherjs/commit/156e55f1a88f547055befb93e93ac738f45b8a48#r81462862) :)

tomconnell-wf:
Fwiw, I transpiled all of our rendering code with this branch, and it seemed to operate and output normally.

tomconnell-wf:
How are you all feeling about this?  Is there anything additional I can do to increase comfort levels?

tomconnell-wf:
I can squash again when this is close to merge

tests/map_test.go
```diff
@@ -100,6 +101,96 @@ func Test_MapRange(t *testing.T) {
...
+	}
+	if swmUnwrapped.DummyMap[`key`].Msg != swm.DummyMap[`key`].Msg {
+		t.Error(`Unwrapped DummyMap did not match`)
+	}
+	if swmUnwrapped.MapFunc(mapFuncArg)[`key2`] != swm.MapFunc(mapFuncArg)[`key2`] {
+		t.Error(`Unwrapped MapFunc did not match`)
+	}
+}
+
+func Test_MapStructObjectWrapper(t *testing.T) {
```
nevkontakte: When I was talking about embedding `*js.Object` I was referring to something like this:

```go
o := js.Global.Get("JSON").Call("parse", `{"props": {"one": 1, "two": 2}}`)

type data struct {
  *js.Object
  Props map[string]int `js:"props"`
}

d := data{Object: o}
d.props["one"] // Expect: 1
d.props["two"] // Expect: 2
```

nevkontakte:
@tomconnell-wf thanks, I left a few more comments. Please give me a day or two to rack my brain for more corner cases, but I agree that it is dangerously close to being merge-ready 😉

compiler/statements.go
```diff
@@ -213,7 +213,9 @@ func (fc *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 			keysVar := fc.newVariable("_keys")
 			fc.Printf("%s = %s ? %s.keys() : [];", keysVar, refVar, refVar)
```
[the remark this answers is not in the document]
tomconnell-wf: Good call.

tomconnell-wf:
You're cooking up some good things to investigate.  I'm not particularly in a rush.  Now that the 1.18.5 tag is out, I will be busy getting upgraded for a while.

compiler/expressions.go
tomconnell-wf: For the test I just committed, without the typeof protection, delete called on a nil map would result in this:
```
--- FAIL: Test_MapDelete (0.00s)
TypeError: nilMap.delete is not a function
    at Test_MapDelete (/github.com/gopherjs/gopherjs/tests/map_test.go:140:3)
    at tRunner (/testing/testing.go:1439:3)
    at $goroutine (/Users/tomconnell/go/src/github.com/gopherjs/gopherjs/tests/tests_test.3034384146.js:1632:19)
    at $runScheduled (/Users/tomconnell/go/src/github.com/gopherjs/gopherjs/tests/tests_test.3034384146.js:1678:7)
    at $schedule (/Users/tomconnell/go/src/github.com/gopherjs/gopherjs/tests/tests_test.3034384146.js:1702:5)
    at queueEntry (/Users/tomconnell/go/src/github.com/gopherjs/gopherjs/tests/tests_test.3034384146.js:1775:5)
    at $send (/Users/tomconnell/go/src/github.com/gopherjs/gopherjs/tests/tests_test.3034384146.js:1734:5)
    at $b (/Users/tomconnell/go/src/github.com/gopherjs/gopherjs/tests/tests_test.3034384146.js:7687:9)
    at $b (/syscall/js/gopherjs__js.go:83:5)
    at /Users/tomconnell/go/src/github.com/gopherjs/gopherjs/tests/tests_test.3034384146.js:68:72
    at FSReqCallback.wrapper [as oncomplete] (node:fs:798:5)
FAIL	github.com/gopherjs/gopherjs/tests	0.719s
```

In other words, I believe the code is correct.

tests/map_test.go
tomconnell-wf: Oh, okay, I haven't done much with that before.  Hopefully the test I cribbed from your example is sufficient to show correctness?

compiler/statements.go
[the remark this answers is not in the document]
tomconnell-wf: Turns out, size would have been undefined and the loop would have `break`'d immediately.  But, I updated the code a bit to be, I think, a little less suspicious.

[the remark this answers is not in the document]
tomconnell-wf: Just for reference, something like that compiled to this:
```
		m = false;
		_ref = m;
		_i = 0;
		_keys = _ref ? _ref.keys() : [];
		_size = _ref.size;
		while (true) {
			if (!(_i < _size)) { break; }
			_key = _keys.next().value;
			_entry = _ref.get(_key);
			if (_entry === undefined) {
				_i++;
				continue;
			}
			k = _entry.k;
			v = _entry.v;
			$unused(k);
			$unused(v);
			_i++;
		}
	};
```

tomconnell-wf:
Any brainstorms over the weekend about edge case problems?

compiler/expressions.go
```diff
@@ -475,9 +475,21 @@ func (fc *funcContext) translateExpr(expr ast.Expr) *expression {
 			}
 			key := fmt.Sprintf("%s.keyFor(%s)", fc.typeName(t.Key()), fc.translateImplicitConversion(e.Index, t.Key()))
 			if _, isTuple := exprType.(*types.Tuple); isTuple {
-				return fc.formatExpr(`(%1s = %2e[%3s], %1s !== undefined ? [%1s.v, true] : [%4e, false])`, fc.newVariable("_entry"), e.X, key, fc.zeroValue(t.Elem()))
-			}
-			return fc.formatExpr(`(%1s = %2e[%3s], %1s !== undefined ? %1s.v : %4e)`, fc.newVariable("_entry"), e.X, key, fc.zeroValue(t.Elem()))
+				return fc.formatExpr(
+					`(%1s = typeof %2e.get === "function" ? %2e.get(%3s) : undefined, %1s !== undefined ? [%1s.v, true] : [%4e, false])`,
```
nevkontakte, line 479: Here and below: let's make the generated code a bit more compact by adding a helper function in prelude:

```js
$mapIndex = function(m, key) {
  return typeof m.get === "function" ? m.get(key) : undefined;
};
```

```
(%1s = $mapIndex(%2e, %3s), %1s !== undefined ? [%1s.v, true] : [%4e, false])
```

This will also clarify operator precedence in this expression.

compiler/expressions.go
nevkontakte, line 985: Similar to above, introducing a `$mapDelete()` helper should make the generated code more compact and compressible.

compiler/statements.go
```diff
@@ -211,10 +211,15 @@ func (fc *funcContext) translateStmt(stmt ast.Stmt, label *types.Label) {
 			iVar := fc.newVariable("_i")
 			fc.Printf("%s = 0;", iVar)
 			keysVar := fc.newVariable("_keys")
-			fc.Printf("%s = $keys(%s);", keysVar, refVar)
```
nevkontakte, line 214: That reminds me... Is `$keys` function used anywhere else in the codebase after this change? If not, we should remove it from prelude.

tests/map_js_test.go
```diff
@@ -0,0 +1,117 @@
...
+	swmWrapper := js.MakeFullWrapper(swm)
+	swmUnwrapped := swmWrapper.Interface().(*StructWithMap)
+	mapFuncArg := map[string]string{`key2`: `value2`}
+
+	if swmWrapper.Get(`StringMap`).Get(`key`).String() != swm.StringMap[`key`] {
+		t.Error(`StringMap did not match`)
+	}
+	if swmWrapper.Get(`IntMap`).Get(`1`).Int() != swm.IntMap[1] {
+		t.Error(`IntMap did not match`)
+
```
nevkontakte, line 43: nit: Here and below: redundant new line.

tests/map_js_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	if swmUnwrapped.DummyMap[`key`].Msg != swm.DummyMap[`key`].Msg {
+		t.Error(`Unwrapped DummyMap did not match`)
+	}
+	if swmUnwrapped.MapFunc(mapFuncArg)[`key2`] != swm.MapFunc(mapFuncArg)[`key2`] {
+		t.Error(`Unwrapped MapFunc did not match`)
+	}
+}
+
+func Test_MapStructObjectWrapper(t *testing.T) {
```
nevkontakte, line 68: I think this test case is redundant with the one above? Or am I missing something?

tests/map_js_test.go
```diff
@@ -0,0 +1,117 @@
+//go:build js && !wasm
```
nevkontakte, line 1: General test-related nitpick: unlike error messages, test errors are full sentences, starting with a capital letter and ending with punctuation.

Also, this is my personal preference, but I find that phrasing test errors in "Got: $some_unexpected_result. Want: $expected_behavior" pattern makes it easier to debug failures of the test you may not be familiar with. The human-friendly message in the "want" part gives the opportunity to convey finer semantics than, for example, "message doesn't contain a particular substring". (See my comment below for an example).

Since this is my personal preference rather than a formally adopted style requirement, you don't have to consider it merge-blocking, but hopefully the reasoning makes sense to you :)

tests/map_test.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+func Test_MapType(t *testing.T) {
+	defer func() {
+		if err := recover(); err == nil {
+			t.Error(`assignment on nil map should panic`)
+		} else {
+			str := err.(error).Error()
+			if !strings.Contains(str, `assignment to entry in nil map`) {
+				t.Error(`assigning to a nil map should panic`)
```
nevkontakte, line 56: To illustrate my meta-comment above, here the error message could be:

```go
t.Error("Got: unexpected panic: %v. Want: a panic caused by assignment to a nil map.")
```

tests/map_js_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	type StructWithMap struct {
+		StringMap map[string]string
+		IntMap    map[int]int
+		DummyMap  map[string]*Dummy
+		MapFunc   func(map[string]string) map[string]string
+	}
+
+	dummyMap := map[string]*Dummy{`key`: {Msg: `value`}}
```
nevkontakte, line 25: nit: I notice you are using backticks for all string literals. For consistency with the rest of the codebase it would be better to use double-quotes for single-line literals.

nevkontakte (changes requested):
Sorry for the delay, my day job has kept my attention yesterday...

I couldn't think of any additional corner cases, so hopefully we didn't miss anything too bad :) I left what is hopefully the last round of comments, most of which are style-related. Other than that we need to:

  - Rebase the PR on top of 1.18 version of master (if you haven't already).
  - Fix CI failures (currently I can see that the [gopherjs_tests workflow is failing](https://app.circleci.com/pipelines/github/gopherjs/gopherjs/333/workflows/e5be0ca4-2c69-4a58-a2cf-73c43bb789ec/jobs/2484)).
  - Squash commits into one with an informative commit description.

And that should be it! 🎉 Thanks again for your hard work and patience with my nitpicks 😅

compiler/statements.go
tomconnell-wf: I think it is, but I will check

tomconnell-wf:
I agree with all of your suggestions.  I'll get on that today.

tests/map_js_test.go
tomconnell-wf: Maybe it isn't testing anything novel.  It tests that a doubly-wrapped map is usable.  I can remove it if that isn't of particular value.

tomconnell-wf:
Okay, I think that's everything.  I'll be happy to assist if any bugs come up with this in the near future.

tests/map_js_test.go
nevkontakte: Okay, let's keep it for now. My intuition says that it is likely redundant, but I and not sure and having redundant tests is better than not having any.

compiler/statements.go
nevkontakte: No worries, if it's still used we can keep it. Just wanted to make sure it wasn't completely dead code.

nevkontakte:
All looks good to me. I've edited the commit message to exclude irrelevant bits, but I think as soon as CI completes, we are good to merge \o/

nevkontakte (approved):

Issue #1013, opened 2021-04-05
GopherJS generics support


nevkontakte:
It appears that generic (a.k.a. type parameters support) is well underway in the upstream Go ([dev.typeparams](https://github.com/golang/go/tree/dev.typeparams) branch) and is likely to be included in 1.17. Which means GopherJS also needs to support this, ideally at the same time as Go 1.17, which I expect is a lot of work. Does anyone have even approximate sense of what it would take to support generics? @flimzy @dmitshur @neelance.

For now, I'm filing this issue to look for a volunteer (or several) to contribute this support. By myself I almost certainly won't be able to implement this, considering that we have several other important features from earlier releases to take care of, such as modules and `embed`.

inliquid:
I think it's planned for `go1.18`.

nevkontakte:
I've been doing some research on potential ways to support generics in gopherjs and I think it would be good for me to share my thoughts. This isn't a proper design document, but I hope it would help us start a discussion and perhaps make it easier for people to contribute towards this :)

I'll start by sharing links to a few important resources I used:

 * [Go language specification](https://go.dev/ref/spec#Type_parameter_declarations), specifically its parts related to type parameters and instantiation. Reading these parts helped me understand semantics of generics in Go, what they can or can't do. Thankfully a lot of type-checking heavy lifting is done for us in the `go/types` package, so all we need to do is the code generation/runtime support.
 * [Type parameters proposal](https://go.googlesource.com/proposal/+/refs/heads/master/design/43651-type-parameters.md). This is a loooong read and not strictly necessary, but it has a lot of interesting context of why generics came to be that way.
 * And a trio of implementation designs:
   * [Generics implementation - Stenciling](https://go.googlesource.com/proposal/+/refs/heads/master/design/generics-implementation-stenciling.md)
   * [Generics implementation - Dictionaries](https://go.googlesource.com/proposal/+/refs/heads/master/design/generics-implementation-dictionaries.md)
   * [Generics implementation - GC Shape Stenciling](https://github.com/golang/proposal/blob/master/design/generics-implementation-gcshape.md)

These design docs deserve the most attention here, as they describe options the Go team considered for their own implementation. The first two docs present two ends of the spectrum: generate a specialized copy of executable code for each instantiation and generate one copy of universal code, delegating type-specific logic to helper functions passed via the dictionary. Ultimately, the Go team went with the hybrid approach as a trade off between binary size, compile time and complexity.

[...]

PR #1137, opened 2022-08-07
Use ES 2015 syntax to generate more compact code for blocking functions.

nevkontakte:
The idea is that we can avoid referencing each variable identifier multiple times in blocking functions intro and outro using ES2015 syntax sugar for object literals and descructuring assignments.

Consider following example:

```go
package main

var c = make(chan string, 1)

func producer() {
	message := "hi"
	c <- message
}

func consumer() string {
	return <-c
}

func main() {
	go producer()
	println(consumer())
}
```

Generated code of the `producer()` function before this change:

```js
	producer = function() {
		var message, $s, $r;
		/* */ $s = 0; var $f, $c = false; if (this !== undefined && this.$blk !== undefined) { $f = this; $c = true; message = $f.message; $s = $f.$s; $r = $f.$r; } s: while (true) { switch ($s) { case 0:
		message = "hi";
		$r = $send(c, message); /* */ $s = 1; case 1: if($c) { $c = false; $r = $r.$blk(); } if ($r && $r.$blk !== undefined) { break s; }
		$s = -1; return;
		/* */ } return; } if ($f === undefined) { $f = { $blk: producer }; } $f.message = message; $f.$s = $s; $f.$r = $r; return $f;
	};
```

We can see that the local variable `message` is referenced 3 times in lines 1-2 and twice on the last line.

Generated code after this change:

```js
	producer = function() {
		var {message, $s, $r, $c, $f} = $restore(this, {});
		/* */ $s = $s || 0; s: while (true) { switch ($s) { case 0:
		message = "hi";
		$r = $send(c, message); /* */ $s = 1; case 1: if($c) { $c = false; $r = $r.$blk(); } if ($r && $r.$blk !== undefined) { break s; }
		$s = -1; return;
		/* */ } return; } if ($f === undefined) { $f = { $blk: producer }; } $f = {...$f, $r, message, $s};return $f;
	};
```

Local variable `message` is referenced only once on the line 1 and once more on the last line, plus overall the function intro code is a lot more compact.

Reference app: [jQuery TodoMVC](https://github.com/gopherjs/todomvc) (`acf500a6c32a83d8c4582d967b09a65febf0e120`)

|          BRANCH          |            ORIGINAL             |            MINIFIED             |       COMPRESSED (GZIP)        |
|--------------------------|---------------------------------|---------------------------------|--------------------------------|
| Pull request (es6-loads) | 2,904,199 bytes                 | 1,913,886 bytes                 | 381,037 bytes                  |
| Target branch (master)   | 8.83% decrease (-281,161 bytes) | 7.86% decrease (-163,201 bytes) | 3.23% decrease (-12,722 bytes) |

Updates #136.


flimzy (approved):
Impressive size results!

Any idea if/how this affects performance of the executed code?

nevkontakte:
Good question, I assumed it should be more or less neutral, but the GoroutineSwitching benchmark shows a significant degradation:

```
$ benchstat /tmp/old.txt /tmp/new.txt
name                old time/op  new time/op  delta
GoroutineSwitching   356ns ± 3%  1218ns ± 2%  +241.96%  (p=0.000 n=10+8)
```

In real life it probably isn't so bad — our CI run time did not change much, but I'll look into that and see if we can improve it.

At some future point I was also considering reimplementing goroutines using ES 2015 generator functions, which should be even more compact and hopefully performant, but that would be a much bigger change.

nevkontakte:
After some investigation I was able to make this change more or less performance-neutral, and even push size reduction a bit further.

Here are some benchmarking results:

```
Env: Node.js 12 / WSL
name                old time/op  new time/op  delta
GoroutineSwitching   350ns ± 3%   343ns ± 3%  -1.84%  (p=0.004 n=10+9)
Env: Chrome 104 / Win10
name                old time/op  new time/op  delta
GoroutineSwitching   266ns ± 2%   253ns ± 4%  -4.85%  (p=0.000 n=9+10)
Env: Firefox 103 / Win10
name                old time/op  new time/op  delta
GoroutineSwitching   684ns ±14%   735ns ±29%   ~     (p=0.156 n=9+10)
Env: Edge 104 / Win 10
name                old time/op  new time/op  delta
GoroutineSwitching   251ns ± 2%   268ns ± 1%  +6.77%  (p=0.000 n=9+10)
```

A couple of caveats for the benchmark numbers:

 - Performance deltas for Node, Chrome and Edge are likely explained by variable background load on my PC. I didn't bother setting up clean benchmarking environment, so I would consider this neutral overall.
 - Firefox seems to show some decrease in performance, but it also shows very high variability in the performance numbers, to the point that benchstat calls the difference statistically insignificant. Unfortunately, its profiling tools lack the ability of per-line profile, so I was unable to pinpoint the difference.

|          BRANCH          |            ORIGINAL             |            MINIFIED             |       COMPRESSED (GZIP)        |
|--------------------------|---------------------------------|---------------------------------|--------------------------------|
| Pull request (es6-loads) | 2,877,935 bytes                 | 1,895,863 bytes                 | 380,638 bytes                  |
| Target branch (master)   | 9.65% decrease (-307,425 bytes) | 8.72% decrease (-181,224 bytes) | 3.33% decrease (-13,121 bytes) |

nevkontakte:
Overall it's  worth noting that the like with every synthetic benchmark, the real-world difference is likely to be much, much smaller. `GoroutineSwitching` benchmark was designed to cause goroutines block on each other as aggressively as possible, which is unlikely to happen in real-life code.

nevkontakte:
As a curious side note, gopherjs on Node 12 performs in this benchmark only two times slower than native go:

```
name                old time/op  new time/op  delta
GoroutineSwitching   157ns ± 0%   343ns ± 3%  +118.22%  (p=0.000 n=9+9)
```

Hat tip to V8 devs for that.

flimzy (approved):

paralin:
@nevkontakte thanks for your work on this and gopherjs in general. I can try to test this PR on a fairly complex project and see if it works. Unfortunately some packages I'm importing started using generics... Will try to see if I can help implement that perhaps.

nevkontakte:
Thanks @paralin, any additional testing is always welcome, although I'm fairly confident in this PR and was going to merge it as soon as I address a couple of minor comments.

If you'd like to help adding support for generics, it will be much appreciated. I've been doing some research on how it could be done in GopherJS context and will try to write that down https://github.com/gopherjs/gopherjs/issues/1013 today after work.

tomconnell-wf:
I'm a fan of this.  Our app compiles to 5.6MB minified-and-gzipped, so even a 3% decrease in size is going to be at least 100-some kb smaller, which is significant.

Issue #847, opened 2018-07-24
Semantic version tag to GopherJS itself


hajimehoshi:
I realized that I need to write `v0.0.0-(datetime)` version to represent dependency on GopherJS. I'd be happy if GopherJS has semantic version tags to `go.mod` cleaner. I think v0.x.y would be fine. Any thoughts?

flimzy:
GopherJS already follows the equivalent Go version numbers. i.e. `1.10-4`.  Are you suggesting adjusting this to follow semver, i.e. `1.10.4`?  Or create an entirely new versioning scheme?

hajimehoshi:
Sorry if I am wrong, but does GopherJS already version tags?

My suggestion is just to add tags. I don't have a strong opinion what version numbers should be used for GopherJS. If GopherJS's API is stable enough, version 1.x.y would be fine, but I don't think it is necessary that GopherJS version follows Go version.

flimzy:
It doesn't have [tags](https://github.com/gopherjs/gopherjs/releases), but it does have a [version number](https://github.com/gopherjs/gopherjs/blob/8dffc02ea1cb8398bb73f30424697c60fcf8d4c5/compiler/version_check.go). Currently it is set to 1.10-4.

myitcv:
> I don't see any reason for a separate versioning system--especially when the GopherJS API follows the Go API as closely as possible.

I think I agree with @hajimehoshi in that GopherJS should follow semver, and hence add releases/tags accordingly.

GopherJS as a tool, and the non-internal packages `github.com/gopherjs/gopherjs/...`, can all be dependencies of other projects. 

With go modules, use of semver-compatible versions (tags) for dependencies (modules) is a requirement:

https://tip.golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning

I'm no expert on semver (https://semver.org/); clearly we would need to give some thought as to how to follow the Go version numbers whilst allowing for intervening releases of GopherJS, for example to fix a bug between `1.11.1` and `1.11.2`. But hopefully this is possible, else we might need to reconsider our approach.

hajimehoshi:
So the import path would be `github.com/gopherjs/gopherjs/v10/js` or `github.com/gopherjs/gopherjs/v11/js`?

myitcv:
@hajimehoshi I was just working through that _exact_ point 👍 

[...]

PR #1146, opened 2022-09-11
README: suggest go install for installing

dmitshur:
GopherJS gained support for building code in module mode in 2021 (see PR #1042), and it itself didn't have problems being built in module mode since much earlier.
Recently, it started getting semver tags that are recognized in module mode like `v1.17.2` and `v1.18.0-beta1` (see issue #847). Start suggesting the `go install` command to install its latest release in the README.

(It remains possible to get the latest development version with something like `go install github.com/gopherjs/gopherjs@HEAD`, or one of the pre-release versions from `go list -m -versions` such as `go install github.com/gopherjs/gopherjs@v1.18.0-beta1`.)

While here, also update to the new shorter Go website domain and latest Go 1.18 point release.


nevkontakte:
Thanks!

nevkontakte (approved):

dmitshur:
Thanks for reviewing!

PR #1152, opened 2022-09-18
compiler: go:linkname support method

visualfc:
**syntax**
go:linkname pkg.type.method
go:linkname pkg.(*type).method

**type convert**

```
$linknames["pkg.(*type).method"] = function(s) { return s.method(...[...arguments].slice(1))};
```
```
// type convert pkg.(*type).method
	var r = v;
	var ptrType = $ptrType(T);
	if (v.constructor != ptrType) {
		switch (T.kind) {
		case $kindStruct:
			r = $pointerOfStructConversion(v, ptrType);
			break;
		case $kindArray:
			r = new ptrType(v);
			break;
		default:
			r = new ptrType(v.$get,v.$set,v.$target);
		}
	}
// type convert pkg.type.method
	var r = v;
	if (v.constructor != $ptrType(T)) {
		switch (T.kind) {
		case $kindStruct:
			r = $clone(v, T);
			break;
		case $kindSlice:
			r = $convertSliceType(v, T);
			break;
		case $kindComplex64:
		case $kindComplex128:
			r = new T(v.$real, v.$imag);
			break;
		default:
			r = new T(v);
		}
	}
```

**demo**
```
package main

import (
	_ "bytes"
	"fmt"
	_ "unsafe"
)

/*
type bytes.Buffer struct {
	buf      []byte // contents are the bytes buf[off : len(buf)]
	off      int    // read at &buf[off], write at &buf[len(buf)]
	lastRead readOp // last read operation, so that Unread* can work correctly.
}
*/
type buffer struct {
	buf      []byte
	off      int
	lastRead int8
}

//go:linkname _WriteString bytes.(*Buffer).WriteString
func _WriteString(buf *buffer, s string) (n int, err error)

func main() {
	var buf buffer
	_WriteString(&buf, "hello world")
	fmt.Println(buf, string(buf.buf))
}

```


nevkontakte:
Can you give an example when this would be necessary? In general, it seem to me that this would encourage [monkey patching](https://en.wikipedia.org/wiki/Monkey_patch), which is often fragile and is a source of difficult to debug issues.

visualfc:
> Can you give an example when this would be necessary? In general, it seem to me that this would encourage [monkey patching](https://en.wikipedia.org/wiki/Monkey_patch), which is often fragile and is a source of difficult to debug issues.

Yes, it is difficult to implement complete type conversion. 

example

https://github.com/goplusjs/reflectx/blob/main/name_js.go#L17
https://github.com/goplusjs/reflectx/blob/main/rtype_js.go#L80

this package provide https://github.com/goplus/igop Interpreter  Go/Go+ code for web.
the playground https://github.com/goplusjs/play

but this project build for github.com/goplusjs/gopherjs only run on go1.16.

nevkontakte:
Actually, I think I may have misunderstood the intent of this PR.

It allows you to make an unexported method from a third-party package available to your code, right? It does _not_ attempt to replace method implementation in the third-party package with your own? If so, my concern about monkey-patching is not valid.

Another question: does the upstream Go compiler support this style of `go:linkname` directives?

visualfc:
> Actually, I think I may have misunderstood the intent of this PR.
> 
> It allows you to make an unexported method from a third-party package available to your code, right? It does _not_ attempt to replace method implementation in the third-party package with your own? If so, my concern about monkey-patching is not valid.

it can call other package unexport method

> 
> Another question: does the upstream Go compiler support this style of `go:linkname` directives?

gc support go:linkname func symbol ( func and mehtod  pkg.func/pkg.type.method/pkg.(*type).method  )

nevkontakte:
That makes sense, thanks. I see you are still updating this PR. Please let me know when it is ready for a detailed review, and I'll try to do it without too much delay (hopefully) 😀

visualfc:
@nevkontakte add tests and reset commit for this PR.

tests/linkname_test.go
```diff
@@ -24,3 +25,12 @@ func TestLinknames(t *testing.T) {
...
 }
+
+func TestLinknameMethods(t *testing.T) {
+	defer func() {
+		if err := recover(); err != nil {
+			t.Fatalf("method.TestLinkname() paniced: %s", err)
+		}
+	}()
+	method.TestLinkname()
+}
```
nevkontakte, line 36: Very nice to see a comprehensive set of tests, thank you!

One suggestion though: the way this test is currently written, it would be nearly impossible to determine where the panic came from, especially considering that the panic message is  just the bare "got" value with no context. I would suggest reorganizing the tests like so:

```go
// In .../method/method.go:

func TestString(t *testing.T) {
	var i _string
	string_Set(&i, "hello world")
	got := string_Get(i)
	want := "hello world"
	if  got != want{
		t.Fatalf("Got: string_Get(i) = %q. Want: %q.",  got, want)
	}
}

// In tests/linkname_test.go:
func TestLinknameMethods(t *testing.T) {
	t.Run("struct", method.TestStruct)
	// and so on...
	t.Run("string", method.TestString)
}
```

In this example we make use of sub-tests, which will allow the Go test framework to tell us which case exactly broke, and more detailed error message would help diagnosing the problem right away. Does that make sense?

nevkontakte (changes requested):
Apologies for the delay, here's the first round of comments. I think overall this looks good, but there are a couple of places that (I think) could be simplified.

compiler/compiler.go
```diff
@@ -314,11 +316,72 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, gls goLinknameS
...
+		case $kindComplex128:
+			r = new T(v.$real, v.$imag);
+			break;
+		default:
+			r = new T(v);
+		}
+	}
+	return r.%v(...[...arguments].slice(1));
+};
+`, d.LinkingName.String(), recv, method)
```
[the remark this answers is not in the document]
visualfc: $methodToFunction

compiler/package.go
```diff
@@ -568,6 +567,15 @@ func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, impor
...
 				}
 			})
+			if named, ok := o.Type().(*types.Named); ok {
+				if t, ok := named.Underlying().(*types.Interface); ok {
+					pkgPath := named.Obj().Pkg().Path()
+					for i := 0; i < t.NumMethods(); i++ {
+						sym := SymName{PkgPath: pkgPath, Name: named.Obj().Name() + "." + t.Method(i).Name()}
+						d.IMethodLinkingNames = append(d.IMethodLinkingNames, sym)
+					}
+				}
```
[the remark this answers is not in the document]
visualfc: remove interface method support. just for gc compatibility but it seems worthless.

compiler/compiler.go
```diff
@@ -314,11 +316,72 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, gls goLinknameS
...
+		case $kindArray:
+			r = new ptrType(v);
+			break;
+		default:
+			r = new ptrType(v.$get,v.$set,v.$target);
+		}
+	}
+	return r.%v(...[...arguments].slice(1));
+};
+`, d.LinkingName.String(), recv[1:], method)
```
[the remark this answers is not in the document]
nevkontakte: Why not use the existing `$methodExpr` helper?

I may be mistaken, but semantically these two programs should be exactly equivalent:

Example 1:
```go
package main

type X struct{}

func (x X) Foo(i int) { println(i) }

func main() {
	x := X{}
	X.Foo(x, 1)  // Note the special method calling syntax here.
}
```

Example 2:
```go
package main

import (
	_ "unsafe"
)

type X struct{}

func (x X) Foo(i int) { println(i) }

//go:linkname X_Foo main.X.Foo
func X_Foo(x X, i int)

func main() {
	x := X{}
	X_Foo(x, 1)
}
```

We already support example 1 syntax using the `$methodExpr()` helper, so I think it should be reusable for the example 2 with the edits I suggested above.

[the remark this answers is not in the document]
visualfc: $methodToFunction must support type convert for compatible go compiler

eample  the main.name must convert to unexport reflect.name by call name_isExported
```
import (
    _ "unsafe"
    _ "reflect"
)
// compatible to reflect.name
type name struct {
	bytes *byte
}
//go:linkname name_isExported reflect.name.isExported
func name_isExported(name) bool
```

tests/linkname_test.go
visualfc: use testing.T
but not use t.Run for catch panic error

compiler/compiler.go
```diff
@@ -318,7 +320,16 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, gls goLinknameS
...
 			// callers via $linkname object (declared in prelude). We are not using
 			// $pkg to avoid clashes with exported symbols.
-			code := fmt.Sprintf("\t$linknames[%q] = %s;\n", d.LinkingName.String(), d.Vars[0])
+			var code string
+			if recv, method, ok := d.LinkingName.IsMethod(); ok {
+				if strings.HasPrefix(recv, "*") {
+					code = fmt.Sprintf("\t$linknames[%q] = $unsafeMethodToFunction(%v,%q,true);\n", d.LinkingName.String(), d.NamedRecvType, method)
+				} else {
+					code = fmt.Sprintf("\t$linknames[%q] = $unsafeMethodToFunction(%v,%q,false);\n", d.LinkingName.String(), d.NamedRecvType, method)
+				}
```
nevkontakte, line 329: I think this can be further simplified to this:
```suggestion
				code = fmt.Sprintf("\t$linknames[%q] = $unsafeMethodToFunction(%v,%q,%t);\n", d.LinkingName.String(), d.NamedRecvType, method, strings.HasPrefix(recv, "*"))
```

tests/testdata/linkname/method/method.go
```diff
@@ -0,0 +1,470 @@
...
+	return int(m)
+}
+
+type basic uintptr
+
+//go:linkname basic_Set github.com/gopherjs/gopherjs/tests/testdata/linkname/method.(*Basic).Set
+func basic_Set(*_int, int) int
+
+//go:linkname basic_Get github.com/gopherjs/gopherjs/tests/testdata/linkname/method.Basic.Get
+func basic_Get(_int) int
```
nevkontakte, line 236: It seems like these aren't used in any test case?

compiler/prelude/prelude.go
```diff
@@ -522,4 +522,48 @@ var $interfaceIsEqual = function(a, b) {
...
+           break;
+         case $kindArray:
+           r = new ptrType(v);
+           break;
+         default:
+           r = new ptrType(v.$get,v.$set,v.$target);
+         }
+      }
+      return r[name](...[...arguments].slice(1));
+    }
```
nevkontakte, line 544: Since we can use ES 2015 syntax now, using rest parameters is [faster](https://jsbench.me/h5l8nnamor/1). Also I think we don't need separate variables for `r` and `v`.

```suggestion
    return function(r, ...args) {
      var ptrType = $ptrType(typ);
      if (r.constructor != ptrType) {
         switch (typ.kind) {
         case $kindStruct:
           r = $pointerOfStructConversion(r, ptrType);
           break;
         case $kindArray:
           r = new ptrType(r);
           break;
         default:
           r = new ptrType(r.$get, r.$set, r.$target);
         }
      }
      return r[name](...args);
    }
```

Same idea applied to the non-pointer branch.

compiler/compiler.go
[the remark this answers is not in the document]
nevkontakte: I see, I missed that. I guess this is necessary to reattach different methods to the object, in case the linked method calls other methods on that type. In that sense GopherJS diverges from native Go, where the appropriate method table is determined and fixed at compile time, rather than by being pointed at by the object itself.

Then having a different helper makes sense. Although I would note that the same problem remains with arguments of the method. If we are implicitly changing types for those, their method calls may fail as well. Not sure how big of a problem that is.

compiler/linkname.go
```diff
@@ -55,17 +62,32 @@ func newSymName(o types.Object) SymName {
...
+	return
+}
+
 // parseGoLinknames processed comments in a source file and extracts //go:linkname
 // compiler directive from the comments.
 //
 // The following directive format is supported:
 // //go:linkname <localname> <importpath>.<name>
+// //go:linkname <localname> <importpath>.<type>.<name>
+// //go:linkname <localname> <importpath>.<(*type)>.<name>
```
nevkontakte, line 84: Could you also update documentation in doc/pargma.md? Specifically:

 - That methods are now supported too.
 - That receiver type will be converted if necessary, but not the argument types.

nevkontakte (approved):
This change looks good to me!

nevkontakte:
And thank you again for yet another contribution to GopherJS! ☺

PR #1154, opened 2022-09-23
compiler/prelude: fix array type size

visualfc:
fix https://github.com/gopherjs/gopherjs/issues/1155


nevkontakte (approved):
Thanks for the fix!

Any idea why CircleCI is not running for your PRs? This seems to be unique to your PRs only, and I can't figure out what may be wrong here 😕

nevkontakte:
Thank you for the fix!

Issue #1156, opened 2022-09-24
reflect value.Interface not copy for struct


visualfc:
demo
```
package main

import (
	"fmt"
	"reflect"
)

type student struct {
	name string
}

func main() {
	a := []student{student{"A"}, student{"B"}, student{"C"}}
	v := reflect.ValueOf(a)
	i := v.Index(0).Interface()
	a[0] = student{"X"}
	fmt.Println(a, i)
}
```

go run main.go
```
[{X} {B} {C}] {A}
```

gopherjs run main.go
```
[{X} {B} {C}] {X}
```

PR #1157, opened 2022-09-24
compiler/natives/src/reflect: fix valueIntrface check struct copy

visualfc:
fix https://github.com/gopherjs/gopherjs/issues/1156

### DIFF
--- a/compiler/natives/src/reflect/reflect.go
+++ b/compiler/natives/src/reflect/reflect.go
@@ -913,6 +913,11 @@ func valueInterface(v Value, safe bool) interface{} {
 	}
 
 	if isWrapped(v.typ) {
+		if v.flag&flagIndir != 0 && v.Kind() == Struct {
+			cv := jsType(v.typ).Call("zero")
+			copyStruct(cv, v.object(), v.typ)
+			return interface{}(unsafe.Pointer(jsType(v.typ).New(cv).Unsafe()))
+		}
 		return interface{}(unsafe.Pointer(jsType(v.typ).New(v.object()).Unsafe()))
 	}
 	return interface{}(unsafe.Pointer(v.object().Unsafe()))


nevkontakte:
A couple of questions:

 - Should the copy happen when `Value.Index()` or `Value.Interface()` method is called? I think in the equivalent non-reflect code the copy would happen on the index operation.
 - Does `reflectlite` have the same bug? If so, maybe fix it as well?

visualfc:
> A couple of questions:
> 
> * Should the copy happen when `Value.Index()` or `Value.Interface()` method is called? I think in the equivalent non-reflect code the copy would happen on the index operation.

copy struct on reflect.Value to interface time.
  v.Index(0) == a[0]
```
	a := []student{student{"A"}, student{"B"}, student{"C"}}
	v := reflect.ValueOf(a)
	i := v.Index(0)
        j := i.Interface()
	a[0] = student{"X"}
	fmt.Println(a, i, j)
        // output [{X} {B} {C}] {X} {A}
```

> * Does `reflectlite` have the same bug? If so, maybe fix it as well?

sync with reflect

tests/copy_test.go
```diff
@@ -131,3 +132,16 @@ func TestExplicitConversion(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestCopyStructByReflect(t *testing.T) {
+	type Info struct {
```
nevkontakte, line 137:
```suggestion
func TestCopyStructByReflect(t *testing.T) {
	// https://github.com/gopherjs/gopherjs/issues/1156
	type Info struct {
```

For such corner cases I usually add a reference to the original issue in the test case, this helps with code archeology a year or two later :)

nevkontakte (approved):

nevkontakte:
Thank you for your contribution!

PR #1159, opened 2022-10-02
Comments and readability improvements for the functionContext type.

nevkontakte:
A couple of minor refactorings to make generics-related work a bit easier:

 - Add comments for all `functionContext` fields. Some of these could be improved, but I am not familiar enough to do that.
 - Make use of the utility `signatureTypes` type in the `functionContext`, which enables more self-descriptive conditions in the code. This type will be further expanded with generics-specific utilities in later commits.
 - Minor improvements to the indentation-related methods.

### DIFF
--- a/compiler/package.go
+++ b/compiler/package.go
@@ -70,22 +70,53 @@ func (sel *fakeSelection) Index() []int              { return sel.index }
 func (sel *fakeSelection) Obj() types.Object         { return sel.obj }
 func (sel *fakeSelection) Type() types.Type          { return sel.typ }
 
-// funcContext maintains compiler context for a specific function (lexical scope?).
+// funcContext maintains compiler context for a specific function.
+//
+// An instance of this type roughly corresponds to a lexical scope for generated
+// JavaScript code (as defined for `var` declarations).
 type funcContext struct {
 	*analysis.FuncInfo
-	pkgCtx        *pkgContext
-	parent        *funcContext
-	sig           *types.Signature
-	allVars       map[string]int
-	localVars     []string
-	resultNames   []ast.Expr
-	flowDatas     map[*types.Label]*flowData
-	caseCounter   int
-	labelCases    map[*types.Label]int
-	output        []byte
+	// Surrounding package context.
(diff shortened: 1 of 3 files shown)


compiler/package.go
```diff
@@ -70,22 +70,53 @@ func (sel *fakeSelection) Index() []int              { return sel.index }
...
+	// Function context, surrounding this function definition. For package-level
+	// functions or methods it is the package-level function context (even though
+	// it technically doesn't correspond to a function). nil for the package-level
+	// function context.
+	parent *funcContext
+	// Information about function signature types. nil for the package-level
+	// function context.
+	sigTypes *signatureTypes
+	// All variable names available in the current function scope. The key is a Go
+	// variable names and the value is the number of synonymous variable names
```
flimzy, line 90:
```suggestion
	// variable name and the value is the number of synonymous variable names
```

compiler/utils.go
```diff
@@ -769,6 +778,17 @@ func (st signatureTypes) Param(i int, ellipsis bool) types.Type {
 	return st.VariadicType().(*types.Slice).Elem()
 }
 
+// HasResults returns true if the function signature returns something.
+func (st signatureTypes) HasResults() bool {
+	return st.Sig.Results().Len() > 0
+}
+
+// HasNamesResults returns true if the function signature returns something and
```
flimzy, line 786: Would `HasNamedResults` make more sense?

```suggestion
// HasNamesResults returns true if the function signature returns something and
```

flimzy (approved):
LGTM. A few possible spelling errors noticed in comments.

paralin (approved):
LGTM, but check the typos @flimzy found, thanks for the hard work @nevkontakte !

compiler/utils.go
nevkontakte: Yes. Strange that the linter didn't call out mismatch with the actual method name.

compiler/package.go
nevkontakte: Done.

PR #1160, opened 2022-10-07
compiler/natives/src/reflect: compatible go reflect

visualfc:
compiler/natives/src/reflect for compatible go reflect
- newName => resolveReflectName 
- uncommonType can be set

### DIFF
--- a/compiler/natives/src/reflect/reflect.go
+++ b/compiler/natives/src/reflect/reflect.go
@@ -63,7 +63,7 @@ func reflectType(typ *js.Object) *rtype {
 		rt := &rtype{
 			size: uintptr(typ.Get("size").Int()),
 			kind: uint8(typ.Get("kind").Int()),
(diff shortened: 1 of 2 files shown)


nevkontakte:
This change looks reasonable to me, but could you explain what problem does it address? In particular, the change to the uncommonType handling.

visualfc:
> This change looks reasonable to me, but could you explain what problem does it address? In particular, the change to the uncommonType handling.

- resolveReflectName:  compatible go upstream 
- uncommonType : this change for add new type with method by reflect.
in go upstream , uncommType embed rtype and can pre allocate uncommType space. but in gopherjs uncommType allocate cannot pre allocate only mocking from $methodSet, so use `js.InternalObject(rt).Set("uncommonType", js.InternalObject(ut))` to dynamic set enable.

see https://github.com/goplusjs/reflectx provide dynamic add new type with method by reflect.

nevkontakte (approved):

PR #1162, opened 2022-10-17
Replacing eval() calls in syscall/js

matthewbelisle-wf:
Minor tweak to `syscall/js` to follow a security best practice as advised by MDN:

[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!)

> Additionally, modern JavaScript interpreters convert JavaScript to machine code. This means that any concept of variable naming gets obliterated. Thus, any use of eval() will force the browser to do long expensive variable name lookups to figure out where the variable exists in the machine code and set its value. Additionally, new things can be introduced to that variable through eval() such as changing the type of that variable, forcing the browser to re-evaluate all of the generated machine code to compensate.
> 
> Fortunately, there's a very good alternative to eval(): using the [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) constructor. Bad code with eval():

In addition to the security benefits, this means that gopherjs can now be used in environments with a CSP that doesn't allow `scritp-src: 'unsafe-eval'`.

### DIFF
--- a/compiler/natives/src/syscall/js/js.go
+++ b/compiler/natives/src/syscall/js/js.go
@@ -117,9 +117,9 @@ var (
 
 func init() {
 	if js.Global != nil {
-		id = js.Global.Call("eval", "(function(x) { return x; })")
-		instanceOf = js.Global.Call("eval", "(function(x, y) { return x instanceof y; })")
-		getValueType = js.Global.Call("eval", `(function(x) {
+		id = js.Global.Call("Function", "x", "return x")
+		instanceOf = js.Global.Call("Function", "x", "y", "return x instanceof y")
+		getValueType = js.Global.Call("Function", "x", `
   if (typeof(x) === "undefined") {
     return 0; // TypeUndefined
   }
@@ -142,7 +142,7 @@ func init() {
     return 7; // TypeFunction
   }
   return 6; // TypeObject
-})`)
+`)
 	}
 }
 


nevkontakte (approved):

nevkontakte:

compiler/natives/src/syscall/js/js.go
```diff
@@ -117,9 +117,9 @@ var (
 
 func init() {
 	if js.Global != nil {
-		id = js.Global.Call("eval", "(function(x) { return x; })")
-		instanceOf = js.Global.Call("eval", "(function(x, y) { return x instanceof y; })")
-		getValueType = js.Global.Call("eval", `(function(x) {
+		id = js.Global.Call("Function", "x", "return x")
+		instanceOf = js.Global.Call("Function", "x", "y", "return x instanceof y")
+		getValueType = js.Global.Call("Function", "x", `
```
matthewbelisle-wf, line 122: @hajimehoshi I'm wondering if you have any thoughts on this one? It seems to me like this should be okay but would definitely like a second opinion. The [Function constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) is the preferred alternative to `eval()` apparently.

[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!)

> Fortunately, there's a very good alternative to eval(): using the [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) constructor.

何卒よろしくお願い申し上げます🙇‍♂️

matthewbelisle-wf:
You bet @nevkontakte thanks for the quick merge!

matthewbelisle-wf:
For posterity, even though this did make the JS follow MDN best practices, it still cannot run in a CSP context without `script-src 'unsafe-eval'` because both `eval()` and `Function()` require the `unsafe-eval` directive.

nevkontakte:
That makes sense, since `Function()` does still evaluate an arbitrary string as code. If you want to make this work with `unsafe-eval`, you could introduce a `$typeof()` helper function in prelude and call it from the `syscall/js` package, instead of defining it dinamically.

matthewbelisle-wf:
Thanks for the info @nevkontakte , is this what you're describing?

https://github.com/gopherjs/gopherjs/pull/1168

Issue #1163, opened 2022-10-18
reflect.Value.Set error for embed struct


visualfc:
- main.go
```
package main

import (
	"fmt"
	"reflect"
)

func main() {
	c := &A{}
	c.value = true
	c.point = Point{100, 200}
	in := reflect.ValueOf(c).Elem()
	v := reflect.New(in.Type())
	e := v.Elem()
	f0 := e.Field(0)
	e.Set(in)
	fmt.Println(e.Field(0), f0, e.Field(0) == f0)
}

type Point struct {
	x int
	y int
}

type Embed struct {
	value bool
	point Point
}

type A struct {
	Embed
}
```
- go run main.go
```
{true {100 200}} {true {100 200}} true
```
- gopherjs run main.go
```
{true {100 200}} {false {0 0}} false
```
reflect.Value.Set error no sync with embed field.

PR #1164, opened 2022-10-18
compiler/natives/src/reflect: fix reflect.Value.Set kind Struct by flagIndir

visualfc:
fix https://github.com/gopherjs/gopherjs/issues/1163 reflect.Value.Set for struct by flagIndir


nevkontakte (approved):
Thank you for the bug report and the fix!

PR #1166, opened 2022-10-23
Make creation of source maps optional with a flag

akolybelnikov:
allow setting the option to false with -s shorthand

Fixed https://github.com/gopherjs/gopherjs/issues/953

### DIFF
--- a/tool.go
+++ b/tool.go
@@ -67,7 +67,7 @@ func init() {
 
 func main() {
 	var (
-		options = &gbuild.Options{CreateMapFile: true}
+		options = &gbuild.Options{}
 		pkgObj  string
 		tags    string
 	)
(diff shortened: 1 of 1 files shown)


tool.go
```diff
@@ -83,6 +83,7 @@ func main() {
 	compilerFlags.StringVar(&tags, "tags", "", "a list of build tags to consider satisfied during the build")
 	compilerFlags.BoolVar(&options.MapToLocalDisk, "localmap", false, "use local paths for sourcemap")
 	compilerFlags.BoolVarP(&options.NoCache, "no_cache", "a", false, "rebuild all packages from scratch")
+	compilerFlags.BoolVarP(&options.CreateMapFile, "source_map", "s", true, "disable generation of source maps")
```
nevkontakte, line 86: If I'm not mistaken, the flag description should be "enable generation of source maps", since passing `-s` will keep them enabled, and you'd need `-s=false` to actually disable them. Alternatively, you could change the flag to be `no_source_map` with the corresponding semantics.

akolybelnikov: You're right. What if we change it to `no_source_map` with `false` value and keep the default as `true` in order not to change the existing behavior?

nevkontakte: I don't think this will behave the way you want it to behave. First of all, you are still binding the flag to the `options.CreateMapFile`, so is you pass `--no_source_map=true`, `options.CreateMapFile` will be also set to `true`, which is the opposite of what it should be.

Personally, I would go back to your original commit and simply change the flag description to match what it actually does: "enable source map generation" :) Then the users will know that if they want to disable source maps, they need to pass `--source_map=false`.

akolybelnikov: That's more logical :)

nevkontakte (approved):
Thank you and congratulations on your first contribution to GopherJS! 🎉

PR #1168, opened 2022-10-24
Making syscall/js CSP compatible

matthewbelisle-wf:
Making the `syscall/js` package compatible with a CSP that does not allow `script-src 'unsafe-eval'`. See [this thread](https://github.com/gopherjs/gopherjs/pull/1162#issuecomment-1287836443) for background.

cc @nevkontakte can you merge this if you think it's good? I tested a CSP manually and it worked, put some snippets below.


compiler/prelude/prelude.go
```diff
@@ -564,4 +564,37 @@ var $unsafeMethodToFunction = function(typ, name, isPtr) {
...
+    return 4; // TypeString
+  }
+  if (typeof(x) === "symbol") {
+    return 5; // TypeSymbol
+  }
+  if (typeof(x) === "function") {
+    return 7; // TypeFunction
+  }
+  return 6; // TypeObject
+};
```
nevkontakte, line 599: Since the mapping between a string type and a number is a `syscall/js`-specific logic, I would only define a `$typeOf` helper in prelude and implement the rest as a Go function in the `syscall/js` package. Then you'd be able able to use Go constants instead of numbers.

matthewbelisle-wf: Okay @nevkontakte I'll do that, thanks for looking.

nevkontakte:
CI tests don't test for CSP compliance (obviously 🙃), but they do run the standard `syscall/js` package tests, so if should inform us if anything is wrong. Overall your change is good, plus or minus the nitpick I left.

matthewbelisle-wf:
Have to log off for the evening and then I'm out of the office tomorrow, but will debug the failures Wednesday morning CDT.

compiler/natives/src/syscall/js/js.go
```diff
@@ -22,27 +22,23 @@ const (
...
+var typeNames = []string{
+	"undefined",
+	"null",
+	"boolean",
+	"number",
+	"string",
+	"symbol",
+	"object",
+	"function",
+}
```
matthewbelisle-wf, line 35: Doing this as a slice because lookups *should* be faster than a switch statement I believe, plus it makes the `getValueType()` implementation easier to match the types.

matthewbelisle-wf:
Tested CSP locally again, looks good.

Screen Shot 2022-10-24 at 10 28 56 PM

compiler/prelude/prelude.go
matthewbelisle-wf: Okay @nevkontakte I finished the rest as a Go function, let me know if there's anything else I should change and I'll tackle it Wednesday morning CDT. Thanks for helping me through this.

nevkontakte (approved):

nevkontakte:
Thank you for your contribution!

PR #1170, opened 2022-10-27
Several small fixes to generics handling

nevkontakte:
- Fix gorepo test runner to handle another variant of the `-gcflags="-G=3"` flag.
 - Correctly return `varFuncLocal` declaration level for JS variables corresponding to local variables.
 - Correctly zero-initialize variables of generic types.

Updates #1013


nevkontakte:
Since these are relatively small changes, I'm going to go ahead and merge them now.

PR #1171, opened 2022-11-03
Fix incorrect continuation callback for a blocking generic function.

nevkontakte:
Previously it was incorrectly returning a reference to the generic factory function. With this change it correctly returns a reference to the generic function instance.

Updates #1013

/cc @paralin


paralin (approved):
LGTM!

PR #1179, opened 2023-01-01
compiler: factor out utility types for processing Go sources and errors.

nevkontakte:
This is the first step in reducing complexity of the compiler.Compile function.

The new `sources` type represents all inputs into the package compilation and simplifies extracting useful information out of them. It is designed to have little business logic of its own and serves as a convenient bridge to other packages like go/types or astrewrite.

The ErrorList type is extended with utility methods that reduce verbosity of the calling code.

/cc @paralin

### DIFF
--- a/compiler/errors.go
+++ b/compiler/errors.go
@@ -1,0 +1,68 @@
+package compiler
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrTooManyErrors is added to the ErrorList by the Trim method.
+var ErrTooManyErrors = errors.New("too many errors")
+
+// ErrorList wraps multiple errors as a single error.
+type ErrorList []error
+
+func (errs ErrorList) Error() string {
+	if len(errs) == 0 {
+		return "<no errors>"
+	}
+	return fmt.Sprintf("%s (and %d more errors)", errs[0].Error(), len(errs[1:]))
+}
+
+// ErrOrNil returns nil if ErrorList is empty, or the error otherwise.
+func (errs ErrorList) ErrOrNil() error {
+	if len(errs) == 0 {
+		return nil
+	}
+	return errs
+}
+
+// Append an error to the list.
+//
+// If err is an instance of ErrorList, the lists are concatenated together,
+// otherwise err is appended at the end of the list. If err is nil, the list is
+// returned unmodified.
+//
+// 	err := DoStuff()
+// 	errList := errList.Append(err)
+func (errs ErrorList) Append(err error) ErrorList {
+	if err == nil {
+		return errs
+	}
+	if err, ok := err.(ErrorList); ok {
+		return append(errs, err...)
+	}
+	return append(errs, err)
+}
+
+// AppendDistinct is similar to Append, but doesn't append the error if it has
+// the same message as the last error on the list.
+func (errs ErrorList) AppendDistinct(err error) ErrorList {
+	if l := len(errs); l > 0 {
+		if prev := errs[l-1]; prev != nil && err.Error() == prev.Error() {
+			return errs // The new error is the same as the last one, skip it.
+		}
+	}
+
+	return errs.Append(err)
+}
+
+// Trim the error list if it has more than limit errors. If the list is trimmed,
+// all extraneous errors are replaced with a single ErrTooManyErrors, making the
+// returned ErrorList length of limit+1.
+func (errs ErrorList) Trim(limit int) ErrorList {
+	if len(errs) <= limit {
+		return errs
+	}
+
+	return append(errs[:limit], ErrTooManyErrors)
+}
--- a/compiler/package.go
+++ b/compiler/package.go
@@ -15,7 +15,6 @@ import (
 	"github.com/gopherjs/gopherjs/compiler/analysis"
 	"github.com/gopherjs/gopherjs/compiler/astutil"
 	"github.com/gopherjs/gopherjs/compiler/typesutil"
-	"github.com/neelance/astrewrite"
 	"golang.org/x/tools/go/gcexportdata"
 )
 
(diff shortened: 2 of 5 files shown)


compiler/errors.go
```diff
@@ -0,0 +1,68 @@
...
+// otherwise err is appended at the end of the list. If err is nil, the list is
+// returned unmodified.
+//
+// 	err := DoStuff()
+// 	errList := errList.Append(err)
+func (errs ErrorList) Append(err error) ErrorList {
+	if err == nil {
+		return errs
+	}
+	if err, ok := err.(ErrorList); ok {
```
flimzy, line 41: Not sure if we'll ever care about the 1.20 `Unwrap() []error` case here, too? Would be easy to add support if we want, and should be safe to add it "early", even before GopherJS is doing anything with 1.20.

On a related note, it would be easy to make `ErrorList` satisfy this interface... and whiel I can't imagine it would be needed, maybe semantically it would be expected? :shrug:

compiler/package.go
```diff
@@ -174,88 +159,28 @@ func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, impor
 			return
 		}
 		// Some other unexpected panic, catch the stack trace and return as an error.
-		err = bailout(fmt.Errorf("unexpected compiler panic while building package %q: %v", importPath, e))
+		err = bailout(fmt.Errorf("unexpected compiler panic while building package %q: %v", importPath2, e))
```
flimzy, line 162: Could be considered a functional change, but while we're here, maybe...

```suggestion
		err = bailout(fmt.Errorf("unexpected compiler panic while building package %q: %w", importPath2, e))
```

compiler/package.go
```diff
@@ -630,8 +555,17 @@ func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, impor
 		return nil, funcCtx.pkgCtx.errList
 	}
 
+	exportData := new(bytes.Buffer)
+	if err := gcexportdata.Write(exportData, nil, typesPkg); err != nil {
+		return nil, fmt.Errorf("failed to write export data: %v", err)
```
flimzy, line 560:
```suggestion
		return nil, fmt.Errorf("failed to write export data: %w", err)
```

compiler/sources.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"go/ast"
+	"go/token"
+	"go/types"
+	"sort"
+
+	"github.com/neelance/astrewrite"
+)
+
+// sources is a slice of parsed Go sources.
```
flimzy, line 12: I guess these are all source files in a specific package, given a single ImportPath? Is it worth mentioning that in the comment?

compiler/sources.go
```diff
@@ -0,0 +1,115 @@
...
+	if importer.Errors.ErrOrNil() != nil {
+		return nil, nil, importer.Errors.Trim(errLimit).ErrOrNil()
+	}
+	// Return any other type errors.
+	if typeErrs.ErrOrNil() != nil {
+		return nil, nil, typeErrs.Trim(errLimit).ErrOrNil()
+	}
+	// Any general errors that may have occurred during type checking.
+	if err != nil {
+		return nil, nil, ErrorList{err}
```
flimzy, line 80: Is it necessary/useful to return a `ErrorList` here, rather than just `err`?

compiler/sources.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return typesInfo, typesPkg, nil
+}
+
+// ParseGoLinknames extracts all //go:linkname compiler directive from the sources.
+func (s sources) ParseGoLinknames() ([]GoLinkname, error) {
+	goLinknames := []GoLinkname{}
```
flimzy, line 87: Looks like this could be preallocated, for a micro-optimization

compiler/sources.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (pi *packageImporter) Import(path string) (*types.Package, error) {
+	if path == "unsafe" {
+		return types.Unsafe, nil
+	}
+
+	a, err := pi.ImportContext.Import(path)
+	if err != nil {
+		pi.Errors = pi.Errors.AppendDistinct(err)
+		return nil, err
```
flimzy, line 111: Cataloging the error in `pi.Errors`, then also returning it, is a bit odd.  I usually expect every error to be handled once... but maybe this makes sense.  I'm not sure where/how `pi.Errors` is used.

flimzy (approved):
Looks reasonable, with a few optional comments inline.  The main "smell" I see is where we're returning an error, and also assigning to `pi.Errors`.  I haven't dug deeper to see if that really makes sense.

compiler/errors.go
nevkontakte: I think supporting `Unwrap() []error` interface if fine, although I'm not sure if it would break `error.Is()` and `error.As()` in some weird way if compiled before 1.20?

Though even if we supported that interface, it don't think it would be semantically correct for us to unwrap multiple errors here, even if the incoming error theoretically implemented that interface. In general case the wrapping error may carry some useful information (like an error message), unwrapping would discard it here. `ErrorList` carries no semantics of its own other than "these errors happened", so flattening it is okay. Does that make sense?

compiler/package.go
```diff
@@ -134,35 +133,21 @@ type flowData struct {
...
-
-	return pi.importContext.Packages[a.ImportPath], nil
-}
-
-func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, importContext *ImportContext, minify bool) (_ *Archive, err error) {
+// Compile the provided Go sources as a single package.
+//
+// Import path must be the absolute import path for a package. Provided sources
+// are always sorted by name to ensure reproducible JavaScript output.
+func Compile(importPath2 string, files2 []*ast.File, fileSet2 *token.FileSet, importContext *ImportContext, minify bool) (_ *Archive, err error) {
```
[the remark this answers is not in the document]
nevkontakte: Oops, that was a temporary rename to make sure there were no stray references to those variables, and I forgot to put that back.

compiler/sources.go
nevkontakte: Yeah, it is a bit odd, but https://github.com/gopherjs/gopherjs/issues/119 explains why that is necessary.

compiler/sources.go
nevkontakte: Not in this case. Although we know how many source files we have, it isn't known in advance how many compiler directives each one of them would have.

Also, personally I find this specific type of micro-optimization a potential source of fragility if the underlying assumption about slice sizes being equal stops being true. So I tend to avoid it unless I know this is a performance bottleneck.

compiler/sources.go
nevkontakte: Good catch, not anymore.

compiler/sources.go
nevkontakte: This is a bit tricky... It is true that all sources should belong to the same package, they don't have to represent a real package that exists on the file system (e.g. playground) or may be only a part of a package (e.g. `gopherjs build foo.go bar.go`). I'll try to add this to the comment.

compiler/package.go
nevkontakte: Had to undo this one because `e` is `interface{}`, not and `error` and `go vet` is upset about it.

flimzy (approved):

Issue #1169, opened 2022-10-26
readme version mismatch, easy fix


nfstein:
The [readme instructions](https://github.com/gopherjs/gopherjs#installation-and-usage) state to install gopherjs@latest which is for the 1.17 release but the go installations below are for the 1.18 release. When I followed the instructions locally, gopherjs was broken until I switched my gopherjs installation to v1.18.0-beta1+go1.18.5.

flimzy:
The documentation still refers to 1.17 intentionally, because 1.18 is still in beta.

nevkontakte:
I think this has confused several people already, so we should just update the README instructions for 1.18-beta1, since this is what most people want anyway.

PR #1180, opened 2023-01-03
README: document the current GOOS and GOARCH values

dmitshur:
GopherJS 1.18 has standardized on a single value for GOOS and GOARCH in PR #1111. Update the README accordingly. (There has been agreement in issue #1169 that it's okay to be updating the top-level README for GopherJS 1.18-specific changes by now.)

Also reword advice in compatibility.md to be more general. The upstream Go WebAssembly port (i.e., GOOS=js GOARCH=wasm, and notably not wasm64) is limited to 4 GiB of linear memory (at least [for now](https://github.com/WebAssembly/memory64/blob/d0fbbe7b1d3c2cabf8409a48f70c94956ebfbe7f/proposals/memory64/Overview.md)), it just [happens to](https://github.com/golang/go/issues/18892#issuecomment-341889653) at this time use 64-bit variables for int/uint/uintptr types. So maybe it's not a great example to refer to. But a reminder that some types vary in size based on the architecture or implementation is probably worth including anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,9 @@ For more details see [Jason Stone's blog post](http://legacytotheedge.blogspot.d
 
 #### General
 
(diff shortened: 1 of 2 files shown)


nevkontakte (approved):
Thanks!

PR #1181, opened 2023-01-03
compiler: break functions related to Decls out of Compile().

nevkontakte:
For each package-level entity we emit one or more Decl struct, which contains all JS code fragments and metadata required to produce the final executable script.

For each decl type (imports, vars, functions and types) I creates a separate function that contains the logic responsible for its creation (and some auxiliary functions). The main objective is to keep the Compile() function very high-level and clearly reflecting various compilation stages we go through.

I tried to add comments to make the code more accessible for future contributors (and future self...), although there are still some aspects I don't fully grasp.

Ideally, we would have tests for all these new functions, but that's way more work than I'm able to take on right now.

/cc @paralin


paralin:
Lgtm

nevkontakte:
@paralin I'm curious, from your perspective how much easier it is to understand what's going on with this part of the code now? :)

compiler/decls.go
```diff
@@ -0,0 +1,541 @@
...
+	// symbol depends on for dead code elimination purposes.
+	DceDeps []string
+	// Set to true if a function performs a blocking operation (I/O or
+	// synchronization). The compiler will have to generate function code such
+	// that it can be resumed after a blocking operation completes without
+	// blocking the main thread in the meantime.
+	Blocking bool
+}
+
+// minify returns a copy of Decl with unnecessary whites paces removed from the
```
flimzy, line 62:
```suggestion
// minify returns a copy of Decl with unnecessary whitespace removed from the
```

compiler/decls.go
```diff
@@ -0,0 +1,541 @@
...
+		panic(bailout(fmt.Errorf("functionContext.discoverObjects() must be only called on the package-level context")))
+	}
+
+	for _, file := range srcs.Files {
+		for _, decl := range file.Decls {
+			switch d := decl.(type) {
+			case *ast.FuncDecl:
+				sig := fc.pkgCtx.Defs[d.Name].(*types.Func).Type().(*types.Signature)
+				if sig.Recv() == nil {
+					fc.objectName(fc.pkgCtx.Defs[d.Name]) // register toplevel name
```
flimzy, line 85: I see here (and elsewhere) that this method modifies `fc`. Is it wise to describe this in the godoc comment as well?  As currently documented, this looks like a pure function.

Edit: I see the same happening in a number of other methods, too.  It may be unnecessarily tedious to document every instance of a method that modifies the receiver here.

compiler/decls.go
```diff
@@ -0,0 +1,541 @@
...
+		panic(bailout(fmt.Errorf("functionContext.funcDecls() must be only called on the package-level context")))
+	}
+
+	var mainFunc *types.Func
+	funcDecls := []*Decl{}
+	for _, fun := range functions {
+		funcDecls = append(funcDecls, fc.newFuncDecl(fun))
+
+		if o := fc.pkgCtx.Defs[fun.Name].(*types.Func); o.Name() == "main" {
+			mainFunc = o // main() function candidate.
```
flimzy, line 276: "candidate"?  Can there be more than one?

flimzy: Random side note: Is there any compile-time check that `main()` takes no arguments, and has no return values?

compiler/package.go
```diff
@@ -127,6 +137,35 @@ type funcContext struct {
...
+			minify:       minify,
+			fileSet:      srcs.FileSet,
+		},
+		allVars:     make(map[string]int),
+		flowDatas:   map[*types.Label]*flowData{nil: {}},
+		caseCounter: 1,
+		labelCases:  make(map[*types.Label]int),
+	}
+	for name := range reservedKeywords {
+		rootCtx.allVars[name] = 1
```
flimzy, line 164: What is the significance of `1` here?

compiler/decls.go
nevkontakte: Done.

compiler/decls.go
nevkontakte: Yeah, pretty much every `functionContext` method may mutate the receiver. I tired to keep functions that can be pure separate.

I'm not too happy with the state of `functionContext` type in general. Side-effect mutations are one thing, but it's also a mostly untestable god object, it directly mutates fields of a bunch of data structures instead of using a semantic API and its name is misleading — it's more like a "JS scope context", rather than "a particular function context". But so far I've got no idea how to fix all of that, so we are going to have to live with it until someone has a 💡 moment.

compiler/decls.go
nevkontakte: Both are good questions...

I *think* there can't be two main in a single package functions in a type-checked code. There *can* be multiple candidates in the whole program though, e.g. if your main package has tests, `gopherjs test` would synthesize another main function. I am actually unsure what happens then, the current implementation looks like we'd invoke both main functions, which is clearly wrong.

compiler/package.go
nevkontakte: It's documented in the allVars godoc: the number of visible Go objects with that exact name potentially accessible from the current scope (e.g. due to variable shadowing). By setting this value to `1` we won't assign the reserved keyword to aby JS object if we encounter a synonymous Go object, but rather `keyword$1`.

paralin:
@nevkontakte yes it definitely makes it easier to understand what's going on!

compiler/decls.go
flimzy: It seems the check is for make's arguments is being done, just not by us.

https://gopherjs.github.io/playground/#/2smtCKvpi_

I haven't tested the test scenario you describe, but I expect it would have bitten me in the past if it were a problem.  When I have some time, I'll try to verify, and file a bug if it's an issue.

flimzy (approved):

PR #1184, opened 2023-01-27
Support methods on generic types.

nevkontakte:
Generic type methods are compiled largely in the same way as standalone generic functions, into a generic function factory. However, unlike a standalone function, the method factory is registered in the receiver type factory and is invoked when the type is being instantiated, with type arguments passed to the generic type. The instantiated method is then attached to the type instance.

At this point we have we are down to 52 failing test cases, and looking through them reveals a few bugs in this PR's implementation, however I will address those separately.

Updates #1013.

/cc @paralin


paralin:
lgtm, great work as always!

compiler/astutil/astutil.go
```diff
@@ -69,6 +76,12 @@ func FuncKey(d *ast.FuncDecl) string {
 	if star, ok := recv.(*ast.StarExpr); ok {
 		recv = star.X
 	}
+	if index, ok := recv.(*ast.IndexExpr); ok {
+		recv = index.X
+	}
+	if index, ok := recv.(*ast.IndexListExpr); ok {
+		recv = index.X
+	}
```
flimzy, line 84: Could a type switch be used here for brevity and clarity?

compiler/package.go
```diff
@@ -530,7 +566,7 @@ func translateFunction(typ *ast.FuncType, recv *ast.Ident, body *ast.BlockStmt,
 		// If a blocking function is being resumed, initialize local variables from the saved context.
 		localVarDefs = fmt.Sprintf("var {%s, $c} = $restore(this, {%s});\n", strings.Join(localVars, ", "), strings.Join(params, ", "))
 		// If the function gets blocked, save local variables for future.
-		saveContext := fmt.Sprintf("var $f = {$blk: "+funcRef+", $c: true, $r, %s};", strings.Join(c.localVars, ", "))
+		saveContext := fmt.Sprintf("var $f = {$blk: "+instanceVar+", $c: true, $r, %s};", strings.Join(c.localVars, ", "))
```
flimzy, line 569: nit: although this predates this PR, building a format string with string concatenation seems a bit odd, given that `instanceVar` doesn't contain formatting verbs.  :shrug:

flimzy (approved):
a couple optional comments inline. Otherwise looks good to me!

compiler/astutil/astutil.go
nevkontakte: Done.

compiler/package.go
nevkontakte: Good catch, done.

PR #1186, opened 2023-02-08
Replace type-switch with a sequence of ifs.

nevkontakte:
Turns out, for a method like `func (v *value[T]) get() T {...}` first `*ast.StarExpr` is unwrapped and then `*ast.IndexExpr`. Using a type-switch allows only one of them to be unwrapped.

I also enabled branch protection on the generics branch to catch this type of issue before merge.


flimzy (approved):
LGTM. It might be nice to add a comment explaining why a switch is avoided.

PR #1189, opened 2023-04-07
Assign identity to all function literals and use them as funcRefs.

nevkontakte:
The main change is that we assign explicit names to all function objects that correspond to Go functions (named and literals). Function name is declared as `var f = function nameHere() { ... }` and is visible inside the function scope only. Doing so serves two purposes:

 - It is an identifier which we can use when saving state of a blocked function to know which function to call upon resumption.
 - It shows up in the stack trace, which helps distinguish similarly-named functions. For methods, we include the receiver type in the identifier to make A.String and B.String easily distinguishable.

The main trick is that we synthesize names for the function literals, which are anonymous as far as go/types is concerned. The upstream Go compiler does something very similar.

Lest but not least, with this change the generic-related code path in the funcContext.translateFunctionBody becomes more self-contained, which will become handy in the upcoming commits.

Updates #1013 

/cc @paralin

### DIFF
--- a/compiler/functions.go
+++ b/compiler/functions.go
@@ -19,20 +19,21 @@ import (
 
 // newFunctionContext creates a new nested context for a function corresponding
 // to the provided info.
-func (fc *funcContext) nestedFunctionContext(info *analysis.FuncInfo, sig *types.Signature) *funcContext {
+func (fc *funcContext) nestedFunctionContext(info *analysis.FuncInfo, o *types.Func) *funcContext {
 	if info == nil {
 		panic(errors.New("missing *analysis.FuncInfo"))
 	}
-	if sig == nil {
-		panic(errors.New("missing *types.Signature"))
+	if o == nil {
+		panic(errors.New("missing *types.Func"))
 	}
 
 	c := &funcContext{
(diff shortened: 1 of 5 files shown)


compiler/functions.go
```diff
@@ -44,6 +45,44 @@ func (fc *funcContext) nestedFunctionContext(info *analysis.FuncInfo, sig *types
...
 	}
 
+	// Synthesize an identifier by which the function may reference itself. Since
+	// it appears in the stack trace, it's useful to include the receiver type in
+	// it. Note that we are using a middle dot unicode character to delimit
+	// between receiver and the function name, since it's a valid identifier
+	// character in JS.
+	funcRef := o.Name()
+	if typeName := c.sigTypes.RecvTypeName(); typeName != "" {
+		funcRef = typeName + "·" + funcRef
```
flimzy, line 55: I wonder if it might be a good idea to define a constant for `"·"`, to both make it more visually distinct (this just looks like a period, without additional textual context or very close visual inspection), and to make it easier to reproduce.

flimzy (approved):

PR #1190, opened 2023-04-08
Map method's type params to their canonical receiver type type params.

nevkontakte:
In Go it is allowed to give different type parameter names in the method declaration compared to the receiver type, for example:

    type A[T any] struct{}
    func (a A[T1]) Method(x T1) {}

To support that go/types creates separate instances of the `*types.TypeParam` object in the type and method declarations for what is essentially the same entity. This was creating an issue when we were generating method reflection metadata for a generic type if the following conditions are true:

  - Method gives different names to the type parameters compared to its type.
  - Any of the types in the method signature depend on the type parameter.

Because method metadata is generated in the type's generic factory, the type parameters would be declared in the JS code according to their names in the type declaration. However, the method metadata is generated from the method's *types.Func object, which uses type parameters from the method declaration. This mismatch would result in the undefined variable errors.

(Strictly speaking this may also happen when they have the same name, but due to nuances of our JS variable name allocation they may or may not end up being given the same JS variable name and work by accident).

We solve this by establishing a mapping between method's type params and their canonical counter parts from the receiver type declaration. When determining a variable name for a type param, we would prefer using the canonical one if it exists, avoiding the mismatch.

Updates #1013.


compiler/typesutil/signature.go
```diff
@@ -0,0 +1,101 @@
...
+
+// IsGeneric returns true if the signature represents a generic function or a
+// method of a generic type.
+func (st Signature) IsGeneric() bool {
+	return st.Sig.TypeParams().Len() > 0 || st.Sig.RecvTypeParams().Len() > 0
+}
+
+// RecvTypeName returns receiver type for a method signature. For pointer
+// receivers the named type is unwrapped from the pointer type. For non-methods
+// an empty string is returned.
```
flimzy, line 77:
```suggestion
// nil is returned.
```

flimzy (approved):

compiler/typesutil/signature.go
nevkontakte: Done.

paralin:
@nevkontakte thanks for the continued work on this, I will try to start testing the generics branch this week.

PR #1197, opened 2023-06-10
Replace vfsgen with go:embed

flimzy:
This should make the development process slightly easier, and reduce diff noise a bit.

This may be possible to optimize further, by removing the `http.FS` wrapper, but that's a bit of a larger change. Perhaps worth looking into separately.


nevkontakte (approved):
I am a fan of this! :)

PR #1200, opened 2023-06-11
Use go:embed for prelude JS snippets

flimzy:
The main advantage to this specific PR is that it allows for easier JS editor/IDE support when editing JS. As a follow-up, I'll work on eliminating the go:generate step for minification as well.


flimzy:
> 0.39% increase (11,351 bytes)

This appears to be the result of different indentation as enforced by my editor when saving *.js files. We may want to alter this formatting, and add some eslint/prettier rules over time.

nevkontakte:
Yeah, I'm not worried about size increase in the non-minified version, I doubt anyone uses it in production. I think adding a common prettier config would be a good thing to make sure that all contributors follow the same format.

nevkontakte (approved):

PR #1201, opened 2023-06-11
Minify Preldue on the fly without uglifyjs

flimzy:


flimzy:
We decided to go with esbuild in #1204 instead, so closing this PR.

PR #1203, opened 2023-06-11
Diagnose some of the typeparam test failures.

nevkontakte:
Also pick up more advanced run args parsing from the upstream to better deal with -gcflags flag.

### DIFF
--- a/tests/gorepo/run.go
+++ b/tests/gorepo/run.go
@@ -111,9 +111,6 @@ var knownFails = map[string]failReason{
 	"fixedbugs/issue23188.go":  {desc: "incorrect order of evaluation of index operations"},
 	"fixedbugs/issue24547.go":  {desc: "incorrect computing method sets with shadowed methods"},
 
-	// These are new tests in Go 1.11.5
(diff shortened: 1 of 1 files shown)


tests/gorepo/run.go
```diff
@@ -1315,3 +1310,65 @@ func getenv(key, def string) string {
...
+// For example, the following string:
+//
+//	a b:"c d" 'e''f'  "g\""
+//
+// Would be parsed as:
+//
+//	[]string{"a", "b:c d", "ef", `g"`}
+//
+// [copied from src/go/build/build.go]
+func splitQuoted(s string) (r []string, err error) {
```
flimzy, line 1331: this looks like a good candidate for some tests, but up to you

flimzy (approved):

tests/gorepo/run.go
nevkontakte: It's a direct lift from the upstream: https://github.com/golang/go/blob/go1.18.10/test/run.go#L2204-L2264. Our run.go is actually based on the old version of that upstream script and could use some updates. But specifically for this function, I don't feel too motivated to write our own tests for it...

PR #1204, opened 2023-06-11
Use esbuild to minify prelude

flimzy:
Possible alternative to #1201


compiler/prelude/prelude.go
```diff
@@ -23,3 +24,21 @@ var jsmapping string
...
+		MinifyIdentifiers: true,
+		MinifySyntax:      true,
+		KeepNames:         true,
+		Charset:           api.CharsetUTF8,
+		LegalComments:     api.LegalCommentsEndOfFile,
+	})
+	if len(result.Errors) > 0 {
+		e := result.Errors[0]
+		panic(fmt.Sprintf("%d:%d: %s\n%s\n", e.Location.Line, e.Location.Column, e.Text, e.Location.LineText))
+	}
```
nevkontakte, line 41: Maybe `glog.Errorf()` all the errors and then `glog.Fatalf()` with a general error to abort the program?

nevkontakte (approved):

compiler/prelude/prelude.go
flimzy: I made a similar change now. I also log warnings, although that may not be ideal, as it will appear as noise to users of gopherjs, but only developers will be able to fix it. WDYT?

compiler/prelude/prelude.go
```diff
@@ -2,9 +2,10 @@ package prelude
 
 import (
 	_ "embed"
-)
+	"log"
```
nevkontakte, line 5: Elsewhere we've been using logrus, which supports verbosity levels: https://github.com/gopherjs/gopherjs/blob/d0aab331249583912023f32b44e5965c9ff4ee87/build/cache/cache.go#L14

compiler/prelude/prelude.go
nevkontakte: I think we definitely want to log as many details as possible. The only way the users would see this error is if we ship a badly broken release, in which case getting as much debugging information to us with as few roundtrips as possible would help fix the issue faster.

compiler/prelude/prelude.go
```diff
@@ -23,3 +24,26 @@ var jsmapping string
...
+		LegalComments:     api.LegalCommentsEndOfFile,
+	})
+	for _, w := range result.Warnings {
+		log.Printf("WARNING %d:%d: %s\n%s\n", w.Location.Line, w.Location.Column, w.Text, w.Location.LineText)
+	}
+	if errCount := len(result.Errors); errCount > 0 {
+		for _, e := range result.Errors {
+			log.Printf("ERROR %d:%d: %s\n%s\n", e.Location.Line, e.Location.Column, e.Text, e.Location.LineText)
+		}
+		log.Fatalf("JS minification failed with %d errors", errCount)
```
nevkontakte, line 45: nit: "Prelude minification failed..."

PR #1206, opened 2023-06-19
Add linter job to GitHub Actions

flimzy:


.golangci.toml
nevkontakte, line 1: That seems like a very extensive config. Is it an auto-generated default, or is it based on some other project you have been maintaining?

nevkontakte (approved):

.golangci.toml
flimzy: It is copied from another project.  I do intend to clean it up before merging. Let me mark the PR as a draft for the moment.

nevkontakte: SGTM. I gave it a cursory look, but most linters there are unfamiliar to me, so I don't have a good intuition for which ones would be useful for gopherjs, and which would only get in the way.

Another question is whether it's possible to run the liner in the new-issues-only mode locally? If not, I'm afraid it would be tripping my ocd pretty badly 😅

flimzy: Yes, that is possible, and in fact it's what https://github.com/gopherjs/gopherjs/pull/1206/files#diff-9e47c4a624b5f266347841b90d904727e63e37ec36291784ab0f3d9c688c65bfR119 accomplishes

flimzy:
We have enough odd code in our tests that trip up a bunch of linters, that this needs to be handled very carefully, and probably gradually.  I'll close this PR, and may begin introducing additional linters over time, as I have a chance to validate which ones make the most sense for this project.

PR #1207, opened 2023-06-19
gofumpt the repo

flimzy:
This runs `gofumpt` on the entire repo.  This is the default formatter I tend to use, and it was reformatting every file as I changed it.  It seems cleaner to do it all in one fell swoop, assuming we agree to use this format.

https://github.com/mvdan/gofumpt

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,25 @@
+name: golangci-lint
+on:
+  pull_request:
+permissions:
+  contents: read
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v3
+        with:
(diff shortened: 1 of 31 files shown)


flimzy:
It's completely open to discussion. I'm not fixated on gofumpt, although I do quite like it.  And it will be enforced by the linter added in the other PR, if we choose to go this route.

nevkontakte:
The only argument against gofumpt I can think of is that it increases barrier to entry for new contributors. Was that a problem in your other projects?

flimzy:
It's never been a serious barrier, however I've also not used it on an active open source project before, either.

nevkontakte:
Then I am open to giving it a shot, as long as we document it in https://github.com/gopherjs/gopherjs/wiki/Developer-Guidelines. We can always revert back to gofmt if we see people struggling.

flimzy:
Merged, and wiki entry updated.

PR #1215, opened 2023-06-26
Perform a few automated ES6 transformations to our prelude

flimzy:
This is done with the help of [lebab](https://github.com/lebab/lebab).  More transformations could be done. In particular, by using `let` and `const`, instead of `var`, we can safely do a number of other conversions, however, this tool is not able to safely do the let/const conversion automatically, so we should probably do (or at least validate) that manually.


nevkontakte (approved):
This is neat!

PR #1216, opened 2023-06-26
Ability to call overridden functions from the std library

flimzy:
Rebase of #798, and applied change suggested in https://github.com/gopherjs/gopherjs/pull/798#issuecomment-955238429, and a couple cleanups.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -117,7 +117,7 @@ func ImportDir(dir string, mode build.ImportMode, installSuffix string, buildTag
 	return pkg, nil
 }
 
-// Test if we find the '//gopherjs:keep_overridden' comment
+// Test if we find the '//gopherjs:keep-original' comment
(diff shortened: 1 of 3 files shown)


flimzy:
One last bikeshed on this one... Now that we've settled on `gopherjs:keep-original`, would it make sense to rename the original function to match?  i.e.  in place of:

```golang
_gopherjs_overridden_TestOnePassCutoff(t)
```

Do we want this?

```golang
_gopherjs_original_TestOnePassCutoff(t)
```

nevkontakte:
> Do we want this?
> 
> ```go
> _gopherjs_original_TestOnePassCutoff(t)
> ```

I don't feel too strongly, but seems like a good idea.

nevkontakte (approved):

flimzy:
Thanks @r-l-x for creating this original PR. I'm glad to see it finally merged!

PR #1221, opened 2023-07-14
Use native JS API for ParseInt where possible

### LINKED ISSUE #1218: Investigate native JS implementation for ParseInt where permitted
#1101 gained us some performance by using the native JS API for some string-to-int conversions. Now that #798 has been merged, we ought to experiment with using the native JS API for compatible calls to `ParseInt`, then fall back to the Go implementation where necessary.

See https://github.com/gopherjs/gopherjs/pull/1101#issuecomment-1004186289

flimzy:
WIP PR to fix #1218

### DIFF
--- a/compiler/natives/src/strconv/atoi.go
+++ b/compiler/natives/src/strconv/atoi.go
@@ -14,6 +14,10 @@ const (
 
 //gopherjs:keep-original
 func ParseInt(s string, base, bitSize int) (i int64, err error) {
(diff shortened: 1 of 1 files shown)


nevkontakte:
I'm curious why did you decide to close this? Did it not yield a meaningful improvement?

flimzy:
There are many corner cases that make it not as easy as I first expected. So while I intend to get back to it, it didn't seem right to lave this virtually empty PR open in the meantime.

PR #1227, opened 2023-08-28
natives: duplicate aliasing.go in x/crypto revisted

benma:
Same as fcd349a8b55efd1bb45c715624591e75dd237138.

x/crypto internally moved the AnyOverlap from
x/crypto/internal/subtle/ to /x/crypto/internal/alias/.


flimzy:
I believe #1230 will resolve the CI failures, that are unrelated to the changes in this PR.

nevkontakte (approved):
We really ought to figure out a better strategy for handing this than cargo-culting this hack. One of those days...

PR #1229, opened 2023-08-29
support internalising structs

### LINKED ISSUE #975: Transpiling junegunns fzf "Cannot internalize utils.Char"
Hello,

I am trying to transpile junegunn/fzf, a really really good fuzzy-finder written in Golang.

My main-Function looks like this:

```
package main

import (
	"github.com/junegunn/fzf/src/algo"
	"github.com/junegunn/fzf/src/util"
)

func main() {
	js.Global.Set("fzf", map[string]interface{}{
		"FuzzyMatchV2": algo.FuzzyMatchV2,
	})
}
```

Calling `fzf.FuzzyMatchV2(true, false, false, "Hello World", "aW", true, null)` leads to an error:

> ..js:1419 Uncaught Error: runtime error: cannot internalize util.Chars

The util.Chars-Struct looks like this:

```
type Chars struct {
	slice           []byte // or []rune
	inBytes         bool
	trimLengthKnown bool
	trimLength      uint16

	// XXX Piggybacking item index here is a horrible idea. But I'm trying to
	// minimize the memory footprint by not wasting padded spaces.
	Index int32
}
```

If you want to try it, use my fork of fzf, branch "gopherjs". There is a small change, which takes out some unix syscall dependencies, to make the build happen.

> https://github.com/dithmer/fzf/tree/gopherjs

I cannot find out what is leading to this error. Any help is appreciated. 

Thanks in advance.

naorzr:
This PR adds code that maps JavaScript object fields to Go struct fields.
Updated the $kindStruct case with a loop to dynamically map JavaScript object fields to corresponding Go struct fields.

Stuff I've tested,

- [x] simple object to struct conversion
- [x] an object that is nested within an object
- [x] exported/none exported fields

possibly fixes #975

### DIFF
--- a/tests/js_test.go
+++ b/tests/js_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/gopherjs/gopherjs/js"
 )
 
@@ -308,11 +309,12 @@ func TestInternalizeStruct(t *testing.T) {
 		Name string
 		Age  int
 	}
-	var a Person
+	var a, expected Person
+	expected = Person{Name: "foo", Age: 952}
 
 	js.Global.Set("return_person", func(p *Person) *Person {
 		if p == nil {
-			t.Fail() // Or t.Error("Received nil pointer")
+			t.Fail()
 			return nil
 		}
 		a = *p
@@ -320,19 +322,17 @@ func TestInternalizeStruct(t *testing.T) {
 	})
 
 	js.Global.Call("eval", "return_person({Name: 'foo', Age: 952})")
-
-	if a.Name != "foo" || a.Age != 952 {
-		t.Fail()
+	if diff := cmp.Diff(a, expected); diff != "" {
+		t.Errorf("Mismatch (-want +got):\n%s", diff)
 	}
 }
-
 func TestInternalizeStructUnexportedFields(t *testing.T) {
 	type Person struct {
 		Name string
 		age  int
 	}
-	var a Person
-
+	var a, expected Person
+	expected = Person{Name: "foo", age: 0}
 	js.Global.Set("return_person", func(p *Person) *Person {
 		a = *p
 		return p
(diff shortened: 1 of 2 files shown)


flimzy:
I've introduced #1230 which I hope will resolve the build failures, which appear unrelated to these changes.

tests/js_test.go
```diff
@@ -303,6 +303,95 @@ func TestInternalizeDate(t *testing.T) {
...
+		}
+		a = *p
+		return p
+	})
+
+	js.Global.Call("eval", "return_person({Name: 'foo', Age: 952})")
+
+	if a.Name != "foo" || a.Age != 952 {
+		t.Fail()
+	}
```
nevkontakte, line 326: Here and below: please use https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff to do the comparison.

nevkontakte:
The change looks good to me, save for one minor suggestion.

nevkontakte:
> there's a different issue though, unrelated to this PR, it seems pointer fields, such as *string for example, are causing the function to throw, is this well known?

That description doesn't ring a bell. Would you mind separate issue and providing a reproduction example?

tests/js_test.go
naorzr: done

naorzr:
> 

sure, opened here #1231

tests/js_test.go
```diff
@@ -303,6 +304,99 @@ func TestInternalizeDate(t *testing.T) {
...
+		a = *p
+		return p
+	})
+
+	js.Global.Call("eval", "return_person({Name: 'foo', age: 952})")
+
+	// Manually check unexported fields
+	if a.age != expected.age {
+		t.Errorf("Mismatch in age: got %v, want %v", a.age, expected.age)
+	}
```
nevkontakte, line 346: You can use https://pkg.go.dev/github.com/google/go-cmp/cmp#AllowUnexported make diff compare both exported and unexported fields.

flimzy (approved):

tests/js_test.go
naorzr: yea, so I've already tried it, but I got 
`Error: no support for forcibly accessing unexported fields` 

full func
```go

func TestInternalizeStructUnexportedFields(t *testing.T) {
	type Person struct {
		Name string
		age  int
	}
	var a, expected Person
	expected = Person{Name: "foo", age: 0}

	js.Global.Set("return_person", func(p *Person) *Person {
		a = *p
		return p
	})

	js.Global.Call("eval", "return_person({Name: 'foo', age: 952})")

	// Using cmp.Diff with AllowUnexported to compare both exported and unexported fields
	if diff := cmp.Diff(a, expected, cmp.AllowUnexported(Person{})); diff != "" {
		t.Errorf("Mismatch (-want +got):\n%s", diff)
	}
}
```

nevkontakte: That's odd, I'm pretty sure I was able to use it with private fields in another project. In any case, this is too much of a minor issue to block this PR on.

PR #1235, opened 2023-09-10
Type conversion support between a type parameter and basic types.

nevkontakte:
Since gopherjs doesn't generate unique code per generic type instantiation, we won't know the real type of a typeparam until run time. This means that type conversion between a typeparam types and any other type also has to shift into runtime. We do this by adding a static method `convertFrom()` to each type constructor. That way, when the exact type is now known at compile time, the compiler can emit code `T.convertFrom(expr)` regardless of what type `T` is.

This PR adds some basic infrastructure for type conversion and implements support for converting to most basic types: strings, ints, floats, booleans, complex. Later PRs will add support for complex types, which are, well, more complicated to handle.

Updates #1013.


nevkontakte:
The test failures seem to be due to a bug in minified variable allocation logic, which is not directly related to this PR, but somehow triggered by it. I'll look into it and send a fix.

flimzy (approved):

nevkontakte:
Also added support for type conversion to interfaces, since without it many test cases were regressing. They used to work just by the virtue that most of the times conversion to an interface at runtime is actually a no-op, but the original version of this PR explicitly threw an error for unsupported conversions.

https://github.com/gopherjs/gopherjs/pull/1237 should've addressed another class of errors that CI detected previously.

nevkontakte:
Eh, apparently something isn't quite right with reflection. I'll look into it tomorrow.

nevkontakte:
@flimzy okay, all seems to be fixed now. I made some significant changes since your approval though, would you mind giving another look?

compiler/prelude/types.js
```diff
@@ -792,3 +863,234 @@ var $assertType = (value, type, returnTuple) => {
...
+    if (srcType === dstType) {
+        return src.$val;
+    }
+
+    switch (srcType.kind) {
+        case $kindInt64:
+        case $kindUint64:
+            return new dstType(src.$val.$high, src.$val.$low);
+        case $kindUintptr:
+            // GopherJS quirk: a uintprt value may be an object converted to
```
flimzy, line 943: typo? Or is this a term I'm unfamiliar with?

```suggestion
            // GopherJS quirk: a uintptr value may be an object converted to
```

tests/typeparams/conversion_test.go
```diff
@@ -0,0 +1,250 @@
...
+// The expectation is that got and want have the same underlying Go type, and
+// they contain equal values. Src is the original value before type conversion,
+// provided for error message purposes.
+//
+// Note that this function is suitable for checking most conversion results
+// except conversion to interfaces. This is because use of reflect APIs requires
+// conversion to `any` interface, which must be assumed correct for this test to
+// be meaningful.
+func checkConversion(t *testing.T, src, got, want any) {
+	if reflect.TypeOf(got) != reflect.TypeOf(want) {
```
flimzy, line 24: should this have a `t.Helper()` call?

tests/typeparams/conversion_test.go
```diff
@@ -0,0 +1,250 @@
...
+	want dstType
+}
+
+func (tc complexConversion[srcType, dstType]) Run(t *testing.T) {
+	checkConversion(t, tc.src, dstType(tc.src), tc.want)
+}
+
+type stringLike interface {
+	// Ideally, we would test conversions from all integer types. unfortunately,
+	// that tips up the stringintconv check in `go vet` that is ran by `go test`
```
flimzy, line 74:
```suggestion
	// that trips up the stringintconv check in `go vet` that is ran by `go test`
```

tests/typeparams/conversion_test.go
```diff
@@ -0,0 +1,250 @@
...
+	got1 = tc.src
+	var got2 any = tc.src
+	var got3 any = any(tc.src)
+
+	for i, got := range []any{got1, got2, got3} {
+		t.Run(fmt.Sprint(i), func(t *testing.T) {
+			checkConversion(t, tc.src, got, tc.src)
+
+			concrete := got.(srcType) // Must not panic.
+			if runtime.Compiler == "gopherjs" {
```
flimzy, line 145: In light of the comment, Is there meant to be an `else` case that does use reflect.DeepEqual?

flimzy (approved):

compiler/prelude/types.js
nevkontakte: Done.

tests/typeparams/conversion_test.go
nevkontakte: Done. At first I omitted it because I felt seeing which line in checkConversion() triggered the error would be more useful, but today I feel the other way around ¯\\\_(ツ)\_/¯

tests/typeparams/conversion_test.go
nevkontakte: Done.

PR #1247, opened 2023-11-04
Handle untyped constants with type params.

nevkontakte:
Untyped constants present a challenge that the same literal value (e.g. "0xFFFF") has different runtime representations for different types. With type params we don't know the exact type at run time, so we can't pick the right representation at compile time. Instead, we do this in two steps:

 1. Pick a basic type that can precisely represent the constant and initialize the value for that type.
 2. Generate code to perform runtime type conversion from the known type to the type parameter's concrete type.

Untyped nil is a special, simpler case: since for all types that untyped nil can be converted to also happened to have nil as zero value, whenever we detect such conversion we call for type's zero value, which would give us the appropriate nil representation.

Updates #1013.


compiler/astutil/astutil.go
```diff
@@ -226,3 +233,66 @@ func TakeAddress(info *types.Info, e ast.Expr) *ast.UnaryExpr {
 	SetType(info, ptrType, addrOf)
 	return addrOf
 }
+
+// MakeTypedConstant takes an untyped constant value and makes an AST expression
+// representing it with a concrete type that can represent the constant precisely.
+func MakeTypedConstant(info *types.Info, val constant.Value) ast.Expr {
```
flimzy, line 239: I wonder if "TypedConstant" is the right term here. I think I understand why you chose that name--it's returning a typed constant expression. However, if I understand, it functions more as an untyped constant, or at least, it's not synonymous with the Go spec's concept of a typed constant.

I don't have a concrete suggestion for a better name, though.  Just concerned that this might confuse someone reading it in the future.

flimzy (approved):

compiler/astutil/astutil.go
nevkontakte: I am also, unfortunately, unable to come up with a better name. The reason I chose this name is because this function is used to convert legit untyped constants into a concrete type that can represent them. So in the general sense the name is correct, although it doesn't always (technically) return a constant or a fully valid AST expression...

Renaming functions is easy though, so if someone thinks of a better name at a later date, I'm more than open to it:)

PR #1250, opened 2023-11-14
v1.19 - Moving embedded bit from offset to name in reflect and reflectlite

grantnelson-wf:
I updated the reflect and reflectlite package to set the embedded flag as part of the Go 1.19 upgrade.

Commit [e1e66a0 "move embedded bit from offset to name"](https://go-review.googlesource.com/c/go/+/412138) changes how reflect and reflectlite store the embedded flag. The creation of the `name` and the creation for `structField` is affected by this change for Go 1.19. Both are overwritten by GopherJS.


nevkontakte (approved):
Looks good, thank you!

PR #1254, opened 2023-12-06
Adding new directives to prune generics

grantnelson-wf:
# WIP

### Summary

After a conversation with Nevkontakte ([here](https://gophers.slack.com/archives/C039C0R2T/p1701798056129639)), the current goal is to make go1.19 support without generics possible by dropping generic code. To accomplish this, I'm adding a new directive, `gopherjs:purge`, which will remove parts of the original code without replacing them with something in the overlay/override code. The native code will have the directive for some declaration or specification identifier and both the overlay and original will have that decl/spec removed from it.

This will allow us, in a follow up ticket, to add things into the natives that we want to remove, e.g. `type Pointer[T any] struct {}` in atomic.

### Changes:

- [x] Added `hasDirective` to `astutil.go` which will check comments on anything in the AST which can hold a comment for a given directive. This will work for existing directives and the new one.
- [x] Added unit-tests for the new `astutil` directive methods
- [x] Broke out some of the code from `parseAndAugment` into their own methods to keep `parseAndAugment` from getting to long and hard to read. The new methods are `augmentOverlayFile`, `augmentOriginalImports`, and `augmentOriginalFile`.
- [x] Updated `augmentOverlayFile` and `augmentOriginalFile` to check for the `gopherjs:purge` directive and remove from both the overlay and original files anything that was purged.
- [ ] Added a method, `pruneImports`, to remove unused imports from the overlay and original code after decl/spec have been purged from them.
- [ ] Added unit-tests for the new `parseAndAugment` changes.
- [x] Fixed the imports in the native overlay `net/http` to not use `io/ioutil` since in go1.19 the original no longer uses `io/ioutil` and we were failing a test for extra imports.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -12,7 +12,7 @@ jobs:
 
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.18.10"
+          go-version: "1.19.13"
 
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
--- a/.github/workflows/measure-size.yml
+++ b/.github/workflows/measure-size.yml
@@ -11,7 +11,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-go@v2
         with:
-          go-version: '~1.18.10'
+          go-version: '~1.19.13'
       - uses: gopherjs/output-size-action/measure@main
         with:
           name: jQuery TodoMVC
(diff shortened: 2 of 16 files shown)


grantnelson-wf:
I'm going to break up this work into smaller changes:
- https://github.com/gopherjs/gopherjs/pull/1255
- more soon...

grantnelson-wf:
See above comment. This work has been added to four other PRs

PR #1257, opened 2023-12-18
Extending directive checking

grantnelson-wf:
This ticket extends how the directives are determined by allowing any AST node containing a comment able to be checked. This includes imports, functions, types, fields, and values.

This does not affect `//gopherjs:prune-original` and `//gopherjs:keep-original`, so currently we are still only checking for directives on functions. This change is in preparation for adding the `//gopherjs:purge` directive.

The only functional change should be that the directives are no longer prefix tested, meaning if there was a directive like `//gopherjs:keep-original-and-remove-overlay` we wouldn't accidentally match it to `//gopherjs:keep-original`. This is not currently a problem and will not affect the current code. This is just prevention of a potential problem as the directives are being worked on in future tickets.

### DIFF
--- a/compiler/astutil/astutil.go
+++ b/compiler/astutil/astutil.go
@@ -5,7 +5,7 @@ import (
 	"go/ast"
 	"go/token"
 	"go/types"
-	"strings"
+	"regexp"
 )
 
 func RemoveParens(e ast.Expr) ast.Expr {
@@ -82,15 +82,7 @@ func FuncKey(d *ast.FuncDecl) string {
 // such as code expecting ints to be 64-bit. It should be used with caution
 // since it may create unused imports in the original source file.
 func PruneOriginal(d *ast.FuncDecl) bool {
-	if d.Doc == nil {
-		return false
-	}
-	for _, c := range d.Doc.List {
-		if strings.HasPrefix(c.Text, "//gopherjs:prune-original") {
-			return true
-		}
-	}
-	return false
+	return hasDirective(d, `prune-original`)
 }
 
 // KeepOriginal returns true if gopherjs:keep-original directive is present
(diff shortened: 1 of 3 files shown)


compiler/astutil/astutil.go
```diff
@@ -102,15 +94,61 @@ func PruneOriginal(d *ast.FuncDecl) bool {
...
-	for _, c := range d.Doc.List {
-		if strings.HasPrefix(c.Text, "//gopherjs:keep-original") {
-			return true
+	return hasDirective(d, `keep-original`)
+}
+
+// anyDocLine calls the given predicate on all associated documentation
+// lines and line-comment lines from the given node.
+// If the predicate returns true for any line then true is returned.
+func anyDocLine(node any, predicate func(line string) bool) bool {
```
nevkontakte, line 103: Optional: https://pkg.go.dev/go/ast#Inspect may be a more concise alternative to manual AST traversal.

nevkontakte (approved):

compiler/astutil/astutil.go
grantnelson-wf: Awesome! I hadn't considered/noticed `Inspect`. I'll update to use that instead.

grantnelson-wf: Update. That looks better. I added a few more tests to make sure if we are looking for directives on declarations we don't accidentally dig down into the specifications and get a false positive.

Essentially, I wanted this to work:

```Go
//gopherjs:do-stuff
type (           // <-- has do-stuff
   Bill struct{} // <-- has no directive (inherits do-stuff in augmenter)
   Bob  struct{} // <-- has no directive (inherits do-stuff in augmenter)
)

type (            // <-- has no directive
   //gopherjs:do-stuff
   Sally struct{} // <-- has do-stuff  

   //gopherjs:do-other-thing
   Larry struct{} // <-- has do-other-thing
)

//gopherjs:do-stuff
type Foo struct{} // <-- the declaration has do-stuff, Foo does not but inherits it in augmenter
```

nevkontakte (approved):

PR #1258, opened 2023-12-18
Add purge directive

grantnelson-wf:
### Summary

After a conversation with Nevkontakte ([here](https://gophers.slack.com/archives/C039C0R2T/p1701798056129639)), the current goal is to make go1.19 support without generics possible by dropping generic code. To accomplish this, I'm adding a new directive, `gopherjs:purge`, which will remove parts of the original code without replacing them with something in the overlay/override code. The native code will have the directive for some declaration or specification identifier and both the overlay and original will have that decl/spec removed from it.

This will allow us, in a follow up ticket, to add things into the natives that we want to remove, e.g. `type Pointer[T any] struct {}` in atomic. This will also help with imports for variable type constraints, e.g. `type Foo[T cmp.Ordered](value T)`.

### Dependancies

- https://github.com/gopherjs/gopherjs/pull/1255
- https://github.com/gopherjs/gopherjs/pull/1256
- https://github.com/gopherjs/gopherjs/pull/1257

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -125,9 +125,11 @@ type overrideInfo struct {
 	// If false the original code is removed.
 	keepOriginal bool
 
-	// pruneMethodBody indicates that the body of the methods should be
-	// removed because they contain something that is invalid to GopherJS.
-	pruneMethodBody bool
+	// purgeMethods indicates that this info is for a type and
+	// if a method has this type as a receiver should also be removed.
+	// If the method is defined in the overlays and therefore has its
+	// own overrides, this will be ignored.
+	purgeMethods bool
 }
 
 // parseAndAugment parses and returns all .go files of given pkg.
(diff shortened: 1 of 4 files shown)


build/build.go
```diff
@@ -298,45 +319,176 @@ func augmentOriginalImports(importPath string, file *ast.File) {
...
+func pruneImports(file *ast.File) {
+	unused := make(map[string]int, len(file.Imports))
+	for i, in := range file.Imports {
+		if name := astutil.ImportName(in); len(name) > 0 {
+			unused[name] = i
+		}
+	}
+
+	// Remove "unused import" for any import which is used.
+	ast.Walk(astutil.NewCallbackVisitor(func(n ast.Node) bool {
```
nevkontakte, line 409: Isn't this the same as https://pkg.go.dev/go/ast#Inspect?

build/build_test.go
```diff
@@ -218,22 +212,199 @@ func TestOverlayAugmentation(t *testing.T) {
...
 			fsetSrc := token.NewFileSet()
 			fileSrc := srctesting.Parse(t, fsetSrc, pkgName+test.src)
 
 			overrides := map[string]overrideInfo{}
 			augmentOverlayFile(fileSrc, overrides)
+			pruneImports(fileSrc)
+
+			buf := &bytes.Buffer{}
+			_ = printer.Fprint(buf, fsetSrc, fileSrc)
+			got := buf.String()
```
nevkontakte, line 395: Here and below, I think `srctesting.Format()` can be used instead.

nevkontakte (approved):
I left a couple of minor suggestions, but otherwise looks good. Please ping this PR once the other PRs are merged and it is rebased.

build/build.go
grantnelson-wf: Yep, you're correct. I'll change over too `Inspect`.

I based this code off of [golang.org/x/tools/imports](https://pkg.go.dev/golang.org/x/tools/imports) and it used `ast.Walk` (seen [here](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.16.1:internal/imports/fix.go;l=188)). So that was in my head when I wrote this lightweight version. `Inspect` would definitely be better here. Thanks for catching that.

build/build_test.go
grantnelson-wf: Ooh, yes it can. I didn't see/remember that was available. I'll change to use that. Thanks

build/build.go
grantnelson-wf: Fixed

build/build_test.go
grantnelson-wf: Fixed

nevkontakte (approved):

nevkontakte:
All looks good and I've merged https://github.com/gopherjs/gopherjs/pull/1257. Could you please resolve the conflicts? I'll merge this one then. Thanks!

grantnelson-wf:
Updated and resolved. Thank you!

nevkontakte (approved):
Thank you!

PR #1262, opened 2024-01-16
[go1.19] Updating natives

grantnelson-wf:
This is the updates to the native overrides to remove type parameters from the standard libraries as part of the generic-less go1.19.

There is more to still be done for natives, this just is part of the whole update. More will come later as we work our way through all the tests.

#### Dependencies:

- This code uses the new directive, `gopherjs:override-signature` so won't function until #1260 has been merged into `go1.19`.


nevkontakte (approved):
All looks good, I'll wait until #1260 is merged and merge this one too.

PR #1263, opened 2024-01-30
[go1.19] Known fails and build issues fix

grantnelson-wf:
This PR has three changes:

1. Fixed a problem when removing/keeping imports in the parse and augment phase. The problem comes from when a file imports something without using the blank (`_`) name (e.g. `import "unsafe"` or `import us "unsafe"`) , then augmentation removes all usages of that import except for one or more directives (e.g. `//go:linkname`). In this case leaving the import as is will cause an "unused import" error but removing the import will result in a "directive X needs import Y" error. To solve this I made this specific case change the name of the import to be a blank import (e.g. `import _ "unsafe"`).

2. Chocolatey doesn't have go1.19.13 so the Windows CI fails. I changed the Chocolatey to the nearest version it defines, go1.19.9. This fixes the problems for the Windows' CI.

3. I updated the known fails list for fixed bugs to include new known fails for tests with generics (type parameters) and removed issues which function correctly under go1.19 without generics.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -475,6 +475,8 @@ func pruneImports(file *ast.File) {
 		path, _ := strconv.Unquote(in.Path.Value)
 		directivePrefix, hasPath := directiveImports[path]
 		if hasPath && astutil.HasDirectivePrefix(file, directivePrefix) {
+			// since the import is otherwise unused set the name to blank.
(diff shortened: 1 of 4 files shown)


nevkontakte (approved):
LGTM, thank you!

PR #1264, opened 2024-01-31
[go1.19] Updating more natives for go1.19

grantnelson-wf:
Updating more natives for go1.19 without generics:

1. Updated `crypto`: It had a copy/paste issue and had to remove the cache. The cache hooks into the GC to clear periodically. I wasn't sure if there was a JS equivalent or if we wanted to use a timer, so I just removed the cache, and now it is treated as if it was always a cache miss.
2. Added fix for `debug/pe`: It had some unsafe pointers for casting one struct into another struct's memory footprint.
3. Updated `go/token/position`: Was getting issues with finding most recent file so changed the fix to use `*File` directly.
4. Updated `hash/maphash`: It had parts from go1.16 which could be updated to match closer to go1.19. It had some new methods, `Bytes` and `String`, that were using unsafe pointer.
5. Updated `reflect` and `reflectlite`: It has some minor changes with some notable changes to `Cap` and `Len` to allow for pointers to arrays and to `valueMethodName` to try to get the correct function name.
6. Updated `fastrand` and a usage of a `fastrandu` in `net`: Added the other `fastrand` methods (`fastrandn`, `fastrand64`, and `fastrandu`) to the runtime.
7. Updated `net/netip`: Just some touchups and skipping a test.
8. Updated `sync/atomic`: Skipping some tests.
9. Updated `syscall/js`: Tiny cleanup

After this update go1.19 should be getting very close.

### DIFF
--- a/compiler/natives/src/crypto/internal/boring/bcache/cache.go
+++ b/compiler/natives/src/crypto/internal/boring/bcache/cache.go
@@ -1,0 +1,30 @@
+//go:build js
+// +build js
+
+package bcache
+
+import "unsafe"
+
+// Cache relies on GC to periodically clear the cache.
+// Since GopherJS doesn't have the same GC hooks, it currently can not
+// register this cache with the GC.
+// Without this cache Boring crypto, in particular public and private
+// RSA and ECDSA keys, will be slower because the cache will always miss.
+type Cache struct{}
+
+func (c *Cache) Register()                           {}
+func (c *Cache) Clear()                              {}
+func (c *Cache) Get(k unsafe.Pointer) unsafe.Pointer { return nil }
+func (c *Cache) Put(k, v unsafe.Pointer)             {}
+
+//gopherjs:purge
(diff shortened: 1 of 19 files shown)


nevkontakte (approved):
Some details of the code are difficult for me to validate without spending way too much time, so I will trust into stdlib tests to do that for me. But in general these changes look reasonable to me.

Getting the CI to pass is a huge milestone, thank you!

grantnelson-wf:
@nevkontakte Thank you so much for all your help. Yeah, getting CI passing really helped and will make this work go faster. I'm going to continue working on this effort as needed and help GopherJS continue towards go1.21 (or go1.22) as much as you need and I can.

I got Workiva to run the go1.19 branch on several of their (our since I work for them too) repos to see if the unit-tests for their specific code passes. They will not update to go1.19 until it is in master, this is just a "use case test" of the changes we've been making.

PR #1265, opened 2024-02-12
[go1.19] Limiting when pruneImports is run

grantnelson-wf:
## DO NOT MERGE

This is a test branch used for transparency of the work I'm doing and to get feedback from CI.

---

The part of augmentation that prunes imports based on usage was designed to handle removing imports specific to a type parameter. For example, the import `cmp` might only exist in a file because of the type parameter constraint `cmp.Ordered`. Go doesn't like it if there are unused imports.

However, the code written for pruning imports uses the given name or guesses at the package name via the given path. This does not work for any relative paths (e.g. `./..`) or versioned packages (e.g. `github.com/foo/bar/v2` where the package is `bar`). These two types of paths do not exist in the natives which ware being altered by augmentation, but may exist in customer code.

Since we only really want to prune imports for files which are augmented, I'm adding some checks so that the prune imports code is only run when overlays exist and those overlays have changed the original file.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -486,10 +486,11 @@ func pruneImports(file *ast.File) {
 			// since the import is otherwise unused set the name to blank.
 			in.Name = ast.NewIdent(`_`)
 			delete(unused, name)
+			if len(unused) == 0 {
+				return
+			}
+			break
 		}
-	}
-	if len(unused) == 0 {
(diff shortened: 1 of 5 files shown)


grantnelson-wf:
This work has been broken out into two other PRs

PR #1267, opened 2024-02-13
[go1.19] Limit augmentation to native files which need it

grantnelson-wf:
The part of augmentation that prunes imports based on usage was designed to handle removing imports specific to a type parameter. For example, the import `cmp` might only exist in a file because of the type parameter constraint `cmp.Ordered`. Go doesn't like it if there are unused imports.

However, the code written for pruning imports uses the given name or guesses at the package name via the given path. This does not work for any relative paths (e.g. `./..`), versioned packages (e.g. `github.com/foo/bar/v2` where the package is `bar`), or using the `go.mod` to define a path for one package to an arbitrary path using a `replace`. These problematic paths do not exist in the natives which are being altered by augmentation, but problematic paths may exist in customer code.

Since we only really want to prune imports for files which are augmented, I'm adding some checks so that the prune imports code is only run when overlays exist and those overlays have changed the original file.

I also fixed an issue where a left over `break` was making only the first found directive based import be checked for and none of the following. Added a test that demonstrates the issue and checks the fix.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -180,8 +180,12 @@ func parseAndAugment(xctx XContext, pkg *PackageData, isTest bool, fileSet *toke
 
 	for _, file := range originalFiles {
 		augmentOriginalImports(pkg.ImportPath, file)
-		augmentOriginalFile(file, overrides)
-		pruneImports(file)
+	}
+
+	if len(overrides) > 0 {
+		for _, file := range originalFiles {
+			augmentOriginalFile(file, overrides)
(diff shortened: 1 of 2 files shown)


grantnelson-wf:
To get CI to pass, https://github.com/gopherjs/gopherjs/pull/1266, will have to be merged first since that'll fix the issue with `window_smoke`.

nevkontakte (approved):

Issue #1140, opened 2022-08-17
Add Go 1.19 support


nevkontakte:
[Go 1.19 was released](https://go.dev/blog/go1.19) last week and we should begin work on supporting it.

Skimming through the [release notes](https://tip.golang.org/doc/go1.19), the following changes seem relevant to us:

 - The [sync/atomic](https://tip.golang.org/pkg/sync/atomic/) package defines new atomic types [Bool](https://tip.golang.org/pkg/sync/atomic/#Bool), [Int32](https://tip.golang.org/pkg/sync/atomic/#Int32), [Int64](https://tip.golang.org/pkg/sync/atomic/#Int64), [Uint32](https://tip.golang.org/pkg/sync/atomic/#Uint32), [Uint64](https://tip.golang.org/pkg/sync/atomic/#Uint64), [Uintptr](https://tip.golang.org/pkg/sync/atomic/#Uintptr), and [Pointer](https://tip.golang.org/pkg/sync/atomic/#Pointer).

I hope that supporting it won't be too much work, although I suspect that generics might have started infiltrating the standard library, in which case we'll have to double down on https://github.com/gopherjs/gopherjs/issues/1013 first.

KiddoV:
Any update on this?

nevkontakte:
@KiddoV only that it will happen at some point. I am currently writing a guide for how to add new Go version support in a hope that someone from community would step up and help us out. If no volunteers appear, I will probably work on 1.19 only after generics. Please let me know if you are interested in contributing 🙂

nevkontakte:
In a hope to inspire more contributions I wrote a guide to supporting new Go versions in GopherJS. I would appreciate feedback on what may be unclear (especially to the non-maintainer folks) or can be phrased better (English isn't my native language ¯\_(ツ)_/¯). 

The draft is available here, and should be accessible by anyone (ping me if you are having issues): https://docs.google.com/document/d/1QD_jl9G1jwL2ye_ZeFMtqIEEYZ9cNIHbV3n3R4pXdYA/edit?usp=sharing. I intend to move it into the github wiki, once the feedback is gathered and addressed.

nevkontakte:
Hi @chenrui333, I am acutely aware of this :) Unfortunately, I have only limited amount of time for open source work, and all of it is currently going towards adding generics support. If you have some spare time and the desire to contribute support for  Go 1.19+ - let me know, I'll be happy to provide pointers and advice.

flimzy:
[...]

PR #1269, opened 2024-02-24
Merge go1.19 branch to master

nevkontakte:
With this change GopherJS is now compatible with Go 1.19 standard library and language features (#1140), with the exception of missing generics support (see #1013). Thanks to @grantnelson-wf for doing the majority of work necessary to support the new Go version.


flimzy (approved):
I have not done a thorough review of this PR, as most of it has already been reviewed anyway. But a spot check looks good.

PR #1272, opened 2024-02-25
An alternate approach to generics support

nevkontakte:
This pull request contains a proof-of-concept implementation of generics in GopherJS, which differs from the one I described [in this comment](https://github.com/gopherjs/gopherjs/issues/1013#issuecomment-1217237123) and partially implemented in the [generics](https://github.com/gopherjs/gopherjs/tree/generics) branch.

Conceptually, this implementation falls in the [monomorphization](https://en.wikipedia.org/wiki/Monomorphization) category, where for each instantiation of a generic type of function, a separate, specialized version of the executable code is generated. In gopherjs, we do this in two phases:

 1. Iteratively traverse type-checked AST to discover instantiations of generic objects. 
     1. We start with non-generic code and look for references to generic types and functions.
     2. Then we scan bodies of generic types and functions for each combination of type parameters previously discovered, while substituting references to type parameters with the current set of type arguments.
     3. If new instantiations are discovered, they are enqueued for further scanning.
     4. The process completes once no new combinations of type arguments are getting dicsovered.
  2. During transpilation, each generic object is translated multiple times, once for each unique type argument combination discovered in step 1. During translation references to type parameters are once again replaced with corresponding type arguments, producing specialized code for each combination. Non-generic objects are translated as before, only once.

Most of the generics-related logic is located in the `compiler/internal/typeparams` package. Type parameter substitution logic has been borrowed more or less verbatim from `golang.org/x/tools/go/ssa` and is located in `compiler/internal/typeparams/subst`.

The most obvious disadvantage of this approach is bigger generated code size, since each instantiation of a generic type pretty much generates a whole copy of the thing. However, this drawback is offset by several significant advantages:

 - You only pay the size cost if you use generics. The original approach required moving a lot of logic from compiler to prelude, which made it bigger for everyone.
 - The generated code is equivalent in efficiency to non-generic code. The original approach relied on dynamic dispatch for a lot of low-level operations since exact types were not known at compile time.
 - Changes to the compiler are much, much more limited and easier to maintain. The original approach required duplicating a considerable amount of logic both in compiler (for non-generic code) and in prelude (for generic code).

In future we can also considerably improve code size by moving towards the stenciling approach vanilla Go took, as well as using DCE to remove unused pieces of generic instances independently from each other.

Implementation in this branch currently passes all test cases in the `typeparams` suite in the Go repo (sans a few legitimate divergencies from vanilla), but in not quite ready for production use. Most notably:

 - Missing support for detecting generic instantiations across package boundaries. This should be relatively straightforward to implement with a few modifications to the `build` package.
 - Dead code elimination currently doesn't distinguish different instances of the same function or type: if one of them is used, all of them are considered in alive.
 - Blocking call detection currently conservatively assumes all generic function calls to be blocking. This can be solved by making `compiler/analysis` package instance-aware.
 - Some edge cases around defining generic types inside generic functions are not fully supported. I need to add support for "context" type parameters (in addition to type's own type parameters) in type instances to make that work fully. `typeparam/nested.go` test case represents those edge cases.


paralin:
I like this implementation a lot better. Makes a lot of sense! Thanks so much for your hard work on this @nevkontakte  - I will try to test this with some of my repos soon.

compiler/internal/typeparams/collect.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+		concrete := typesutil.NewSelection(sel.Kind(), recv, index, obj, typ)
+		r.selMemo[sel] = concrete
+		return concrete
+	default:
+		panic(fmt.Errorf("unexpected selection kind %v: %v", sel.Kind(), sel))
+	}
+}
+
+// ToSlice converts TypeParamList into a slice with the sale order of entries.
```
flimzy, line 89:
```suggestion
// ToSlice converts TypeParamList into a slice with the same order of entries.
```

compiler/internal/typeparams/collect.go
```diff
@@ -0,0 +1,266 @@
...
+//
+// Collector will scan non-generic code for any instantiations of generic types
+// or functions and add them to the InstanceSet. Then it will scan generic types
+// and function with discovered sets of type arguments for more instantiations,
+// until no new ones are discovered.
+//
+// InstanceSet may contain unprocessed instances of generic types and functions,
+// which will be also scanned, for example found in depending packages.
+//
+// Note that instanced of generic type methods are automatically added to the
```
flimzy, line 217:
```suggestion
// Note that instances of generic type methods are automatically added to the
```

flimzy (commented):
Will look more closely when not watching kids

compiler/internal/typeparams/collect.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// NewResolver creates a new Resolver with tParams entries mapping to tArgs
+// entries with the same index.
+func NewResolver(tc *types.Context, tParams []*types.TypeParam, tArgs []types.Type) *Resolver {
+	r := &Resolver{
+		subster: subst.New(tc, tParams, tArgs),
+		selMemo: map[typesutil.Selection]typesutil.Selection{},
+	}
+	if len(tParams) != len(tArgs) {
```
flimzy, line 27: Does it make sense to do this check/panic before passing the checked arguments to `subst.New`?

Edit: Now that I look at subst.New, it also seems to panic on the same condition. so maybe we can just eliminate this check/panic entirely?

compiler/internal/typeparams/instance.go
```diff
@@ -0,0 +1,139 @@
...
+
+// ID returns a unique numeric identifier assigned to an instance in the set.
+// The ID is guaranteed to be unique among all instances of the same object
+// within a given program. The ID will be consistent, as long as instances are
+// added to the set in the same order.
+//
+// In order to have an ID assigned, the instance must have been previously added
+// to the set.
+//
+// Note: we these ids are used in the generated code as keys to the specific
```
flimzy, line 97: I think?

```suggestion
// Note: these ids are used in the generated code as keys to the specific
```

compiler/package.go
```diff
@@ -393,69 +387,81 @@ func Compile(importPath string, files []*ast.File, fileSet *token.FileSet, impor
...
-					id := funcCtx.newIdent("", types.NewSignatureType(nil, nil, nil, nil, nil, false))
-					funcCtx.pkgCtx.Uses[id] = o
-					call := &ast.CallExpr{Fun: id}
-					if len(funcCtx.pkgCtx.FuncDeclInfos[o].Blocking) != 0 {
-						funcCtx.Blocking[call] = true
-					}
-					funcCtx.translateStmt(&ast.ExprStmt{X: call}, nil)
+			// Auxiliary decl shared by all instances of the function that defines
+			// package-level variable by which they all are referenced.
+			// TODO(nevkontakte): Set DCE attributes such that it is eliminated of all
```
flimzy, line 402:
```suggestion
			// TODO(nevkontakte): Set DCE attributes such that it is eliminated if all
```

compiler/typesutil/typenames.go
```diff
@@ -0,0 +1,30 @@
...
+
+// TypeNames implements an ordered set of *types.TypeName pointers.
+//
+// The set is ordered to ensure deterministic behavior across compiler runs.
+type TypeNames struct {
+	known map[*types.TypeName]struct{}
+	order []*types.TypeName
+}
+
+// Add a type name to the test. If the type name has been previously added,
```
flimzy, line 13: What "test" is referenced here?

compiler/utils.go
```diff
@@ -319,9 +328,30 @@ func isVarOrConst(o types.Object) bool {
...
-	return o.Parent() != nil && o.Parent().Parent() == types.Universe
+	// Note: named types are always assigned a variable at package level to be
+	// initialized with the rest of the package types, even the types declared
+	// in a statement inside a function.
+	_, isType := o.(*types.TypeName)
+	return (o.Parent() != nil && o.Parent().Parent() == types.Universe) || isType
 }
 
+// assignedObjectName checks if the object has been previously assigned a name
+// in this or one of the parent contexts. If not, found will be false false.
```
flimzy, line 339:
```suggestion
// in this or one of the parent contexts. If not, found will be false.
```

internal/testingx/must.go
```diff
@@ -0,0 +1,24 @@
+// Package testingx provides helpers for use with the testing package.
+package testingx
+
+import "testing"
+
+// Must provides a concise way to handle handle returned error in tests that
```
flimzy, line 6:
```suggestion
// Must provides a concise way to handle returned error in tests that
```

tests/gorepo/run.go
```diff
@@ -1029,7 +1047,7 @@ func (t *test) errorCheck(outStr string, fullshort ...string) (err error) {
 
 func (t *test) updateErrors(out string, file string) {
 	// Read in source file.
-	src, err := os.ReadFile(file)
+	src, err := ioutil.ReadFile(file)
```
flimzy, line 1050: Was it intentional to switch back to `ioutil` on this line and others like it?

flimzy (approved):

compiler/internal/typeparams/instance.go
nevkontakte: Done.

compiler/internal/typeparams/collect.go
nevkontakte: Done.

compiler/internal/typeparams/collect.go
nevkontakte: Ah, yes, this is a holdover from an earlier revision that used unexported go/types function that did not have the check. Removed.

compiler/internal/typeparams/collect.go
nevkontakte: Done.

compiler/package.go
nevkontakte: Done.

compiler/typesutil/typenames.go
nevkontakte: Should have been "set". Fixed!

compiler/utils.go
nevkontakte: Done.

internal/testingx/must.go
nevkontakte: Done.

tests/gorepo/run.go
nevkontakte: Nope, that was an unintended effect of me cherry-picking some changes from the original generics branch. Fixed.

PR #1279, opened 2024-03-23
Guard generics support behind an experiment flag.

nevkontakte:
This commit adds support for GOPHERJS_EXPERIMENT environment variable that can be used to enable experimental functionality in the compiler.

Experiment flags can be easily added or removed by updating fields of the internal/experiments.Flags struct. For now, only boolean flags are supported, but it should be easy enough to extend if needed.

Users need to have GOPHERJS_EXPERIMENT=generics set in their environment to enable support. Adding the flag will allow us to merge generics support into the main branch before it's fully production-ready without creating unnecessary confusion for the users.

Updates #1013

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -74,6 +74,8 @@ orbs:
 jobs:
   build:
     executor: gopherjs
+    environment:
+      GOPHERJS_EXPERIMENT: generics
     steps:
     - setup_and_install_gopherjs
     - run:
@@ -129,6 +131,8 @@ jobs:
   gopherjs_tests:
     executor: gopherjs
     parallelism: 4
+    environment:
+      GOPHERJS_EXPERIMENT: generics
     steps:
     - setup_and_install_gopherjs
     - run:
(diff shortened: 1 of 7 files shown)


circle.yml
```diff
@@ -204,6 +212,8 @@ jobs:
   darwin_smoke:
     macos:
       xcode: 13.4.1 # Mac OS 12.6.1, see https://circleci.com/docs/using-macos/
+    environment:
+      GOPHERJS_EXPERIMENT: generics
```
flimzy, line 216: Do we want to run our tests (or maybe a subset) both with and without generics support, to ensure we don't break non-generics support during this transition?

internal/experiments/experiments.go
```diff
@@ -0,0 +1,122 @@
+// Package experiments managed the list of experimental feature flags supported
```
flimzy, line 1:
```suggestion
// Package experiments manages the list of experimental feature flags supported
```

flimzy (approved):

circle.yml
nevkontakte: Running all tests seems prohibitively expensive, considering that they ballooned up to 30m again. But it may be worth having a smoke test of some sort. What do you think would be a reasonable subset?

Issue #1270, opened 2024-02-25
Add Go 1.20 support


nevkontakte:
Now that we've released 1.19 support, filing this issue to track work related to [1.20 support](https://go.dev/blog/go1.20).

Release notes: https://go.dev/doc/go1.20

Developer guide for new version support in [wiki](https://github.com/gopherjs/gopherjs/wiki/Development:-New-Go-version-support-guide).

PR #1282, opened 2024-04-02
[go1.20] Updated the linkname to handle new directives

grantnelson-wf:
The linkname directive, `//go:linkname localname [importpath.name]`, now optionally allows one-argument forms. [Details found in compiler directives document](https://pkg.go.dev/cmd/compile#hdr-Compiler_Directives). There was also a `TODO` in `linkname.go` to check for any additional arguments. The `os` package had a new insert-link to `net` that needed to be ignored and handled in the native overrides.

- I reorganized the `parseGoLinknames` method to separate it into:
  - reading the linkname from the comment
  - checking if the linkname is supported by GopherJS
- I added two methods to check for known links to ignore:
  - one method is for specific var-links
  - one method is for specific insert-links or any insert-links in specific packages
- I added the native override for `net.newUnixFile`

The introduction of the one-argument form (and the new part in the compiler directives document) makes me feel like there will be more insert-style links on the horizon since the one-argument is to "suppresses the usual error for a function that lacks a body." So by having a better way to deal with specific links when needed, we will be able to track new insert-links and var-links better.

If enough insert-links are added we will probably have to start handling them directly instead of ignoring them and handling them in the native overrides. However, that's seems to be outside of the scope of this PR and https://github.com/gopherjs/gopherjs/issues/1270.

### DIFF
--- a/compiler/linkname.go
+++ b/compiler/linkname.go
@@ -75,6 +75,83 @@ func (n SymName) IsMethod() (recv string, method string, ok bool) {
 	return
 }
 
+// readLinknameFromComment reads the given comment to determine if it's a go:linkname
+// directive then returns the linkname information, otherwise returns nil.
+func readLinknameFromComment(pkgPath string, comment *ast.Comment) (*GoLinkname, error) {
+	if !strings.HasPrefix(comment.Text, `//go:linkname `) {
+		return nil, nil // Not a linkname compiler directive.
+	}
+
+	fields := strings.Fields(comment.Text)
+
+	// Check that the directive comment has both parts and is on the line by itself.
+	switch len(fields) {
+	case 2:
+		// Ignore one-argument form //go:linkname localName
+		// This is typically used with "insert"-style links to
+		// suppresses the usual error for a function that lacks a body.
+		// The "insert"-style links aren't supported by GopherJS so
+		// these bodiless functions have to be overridden in the natives anyway.
+		return nil, nil
+	case 3:
+		// Continue for two-argument form //go:linkname localName importPath.extName
+		break
+	default:
+		return nil, fmt.Errorf(`gopherjs: usage requires 2 arguments: //go:linkname localName importPath.extName`)
+	}
+
+	localPkg, localName := pkgPath, fields[1]
+	extPkg, extName := ``, fields[2]
+
+	if localName == extName {
+		// Ignore self referencing links, //go:linkname localName localName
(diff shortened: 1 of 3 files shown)


compiler/linkname.go
```diff
@@ -98,63 +166,37 @@ func parseGoLinknames(fset *token.FileSet, pkgPath string, file *ast.File) ([]Go
...
-		if !strings.HasPrefix(comment.Text, "//go:linkname ") {
-			return nil // Not a linkname compiler directive.
+		link, err := readLinknameFromComment(pkgPath, comment)
+		if err != nil || link == nil {
+			return err
 		}
 
-		// TODO(nevkontakte): Ideally we should check that the directive comment
-		// is on a line by itself, line Go compiler does, but ast.Comment doesn't
-		// provide an easy way to find that out.
```
grantnelson-wf, line 107: @nevkontakte Please double check that the `if len(fields) != 3 {` does this. It looks like it was already checking that there wasn't anything after the `extname`, so I'm assuming this has already been managed. However, I don't want to remove this `TODO` if I'm just misunderstanding. I did add a unit-test to check for stuff after the `extname`.

compiler/linkname.go
```diff
@@ -75,6 +75,80 @@ func (n SymName) IsMethod() (recv string, method string, ok bool) {
...
+		if len(fields) == 2 {
+			// Ignore one-argument form //go:linkname localname
+			// This is typically used with "insert"-style links to
+			// suppresses the usual error for a function that lacks a body.
+			// The "insert"-style links aren't supported by GopherJS so
+			// these bodiless functions have to be overridden in the natives anyway.
+			return nil, nil
+		}
+		return nil, fmt.Errorf(`gopherjs: usage requires 2 arguments: //go:linkname localname importpath.extname`)
+	}
```
flimzy, line 98: IMO, a switch would be more readable. The if-within-an-if makes me do a double take to be sure I understand what's going on.  But use your judgement.

grantnelson-wf: That's a good idea 👍

compiler/linkname.go
nevkontakte: I don't remember exactly what I meant here 🤔 It may have been related to multi-line comments and how compiler directive must be separated in them?.. But since the meaning of the TODO is uncertain, it's value is also questionable, so I don't mind dropping it either way.

nevkontakte (approved):

PR #1283, opened 2024-04-02
[WIP][go1.20] Working on go1.20 native overrides

grantnelson-wf:
## WIP - Do Not Merge

🚧 This is a temporary PR that I'm using to work on some of the updates to natives and anything else for go1.20. This branch contains many ideas that may or may not be cut out into it's own PR. This allows for fast iteration whilst getting CI testing and feedback. Once I get a self-contained idea working, I'll then cut a PR containing one commit to add that idea to branch go1.20.


grantnelson-wf:
Super Green

Woohoo!! The CI is finally passing! This is a big mile stone for go1.20. With this we'll have a generic-less-ish go1.20. This still requires the generics experiments enabled and still had a lot of generic overrides. Now we have a passing CI it will be easier watch for regressions and improve development speed.

grantnelson-wf:
The branch is no longer needed since I'm switching to help with generics which is being added to master. When we're getting closer to go1.20 being finished I may open up a similar branch like this then.... until then... bye!

PR #1284, opened 2024-04-03
[go1.20] Override cypto/subtle XORBytes

grantnelson-wf:
The crypto/subtle package added the `XORBytes` in go1.20. This method XOR's the content of two slices into a third. This is done using unsafe slices to cast `[]byte` slices to `[]uintptr` slices (done to pack registers as full as possible to reduce the number of XOR's being done) or use an architecture specific method for XOR'ing the data.

I added native overrides to just XOR one byte at a time. I didn't find a better way using `Uint8Array` but there maybe one that I just am not aware of.

### DIFF
--- a/compiler/natives/src/crypto/subtle/xor.go
+++ b/compiler/natives/src/crypto/subtle/xor.go
@@ -1,0 +1,48 @@
+//go:build js
+// +build js
+
+package subtle
+
+func XORBytes(dst, x, y []byte) int {
+	n := len(x)
+	if len(y) < n {
+		n = len(y)
+	}
+	if n == 0 {
+		return 0
+	}
+	if n > len(dst) {
+		panic("subtle.XORBytes: dst too short")
+	}
+
+	// xorBytes(&dst[0], &x[0], &y[0], n) // arch-specific
+	// The above uses unsafe and generics for specific architecture
+	// to pack registers full instead of doing one byte at a time.
+	// We can't do the unsafe conversions from []byte to []uintptr
+	// so we'll simply do it one byte at a time.
+
+	x = x[:len(dst)] // remove bounds check in loop
+	y = y[:len(dst)] // remove bounds check in loop
+	for i := range dst {
+		dst[i] = x[i] ^ y[i]
+	}
+	return n
+}
+
+//gopherjs:purge
+const (
+	wordSize          = 0
+	supportsUnaligned = false
+)
+
+//gopherjs:purge
+func xorBytes(dstb, xb, yb *byte, n int)
+
+//gopherjs:purge
+func aligned(dst, x, y *byte) bool
+
+//gopherjs:purge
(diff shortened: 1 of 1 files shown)


flimzy (approved):

nevkontakte:
One possible way to speed this up for `UInt8Array` is to take its underlying `ArrayBuffer` and wrap it into `UInt32Array()`, do xors through it, and then return the result as a `UInt8Array` again. https://stackoverflow.com/a/68005553/759134 illustrates the idea. However, we do have to be careful about cases when the number of uin8 elements doesn't divide by 4.

LMK if you want to give it a try, otherwise I am okay merging this as-is.

compiler/natives/src/crypto/subtle/xor.go
```diff
@@ -0,0 +1,48 @@
...
+		panic("subtle.XORBytes: dst too short")
+	}
+
+	// xorBytes(&dst[0], &x[0], &y[0], n) // arch-specific
+	// The above uses unsafe and generics for specific architecture
+	// to pack registers full instead of doing one byte at a time.
+	// We can't do the unsafe conversions from []byte to []uintptr
+	// so we'll simply do it one byte at a time.
+
+	x = x[:len(dst)] // remove bounds check in loop
```
nevkontakte, line 24: nit: I assume this comment got copied over from the original source. However, gopherjs compiler is not sophisticated enough to eliminate the check, so the comment is a bit misleading in this context.

grantnelson-wf: Oh, yep, that's a copy/paste. I'll remove it

grantnelson-wf:
> One possible way to speed this up for `UInt8Array` is to take its underlying `ArrayBuffer` and wrap it into `UInt32Array()`, do xors through it, and then return the result as a `UInt8Array` again. https://stackoverflow.com/a/68005553/759134 illustrates the idea. However, we do have to be careful about cases when the number of uin8 elements doesn't divide by 4.
> 
> LMK if you want to give it a try, otherwise I am okay merging this as-is.

That's a really great idea. I didn't know that JS could do that kind of cast. That would be better and the existing code in the natives shows how to deal with any dangling bytes (i.e. if `len%4` is not zero). I'll see what I can do with JS to make that work, thanks.

grantnelson-wf:
Got a version using  is faster `UInt32Array`. I tested  it with playground ([link to test](https://gopherjs.github.io/playground/#/XMBnhImSOQ)) and using `UInt32Array`s is definitely faster (estimate about 3 to 4 times faster). I'll push that one up.

compiler/natives/src/crypto/subtle/xor.go
grantnelson-wf: Removed

flimzy (approved):

nevkontakte (approved):

PR #1290, opened 2024-04-13
Foundations of generics support

nevkontakte:
As discussed [in the issue](https://github.com/gopherjs/gopherjs/issues/1013#issuecomment-2002033648), merging foundations of the generics support into master to keep it in sync with Go 1.20 work and general development.

The general approach taken in the PR is closer to [monomorphization](https://en.wikipedia.org/wiki/Monomorphization): each generic function or type is translated multiple types, once for each combination of type arguments discovered in the program. The main advantages of this approach are:

 - Performance of generic code is equivalent to non-generic.
 - No significant changes to prelude - programs that don't use generics pay no output size penalty.
 - Better maintenance story: there is no divergence between handling of generic and non-generic code, and no need in duplicating parts of compiler logic in the runtime (hence, bigger prelude).

This differs from the [original proposal](https://github.com/gopherjs/gopherjs/issues/1013#issuecomment-1217237123) I made, which was partially implemented in the `generics` branch. The reason that proposal was abandoned are explained in [this comment](https://github.com/gopherjs/gopherjs/issues/1013#issuecomment-1962785024).

Currently generics implementation is NOT production-ready and hidden behind the `GOPHERJS_EXPERIMENT=generics` flag. Namely, some of the major missing features:

 - Generic instantiation discovery doesn't work across package boundary. This means that you can't use generic types or functions defined in a different package, which is a very common practical scenario. Changes to the build system are required to support that.
 - Blocking function analysis hasn't been updated to be generics-aware. The main effect is that a lot of generic calls are considered blocking when they don't need to be, leading to a less efficient code.
 - Dead code elimination doesn't work on individual instantiations. Namely, if a single instance of a generic object is considered alive, all other instantiations of the same object would be considered alive as well.

These issues will be addressed in the upcoming RPs.

Updates #1013.


nevkontakte:
Note: all the changes in this PR have been previously reviewed in the `generics-ng` branch, but I would like to solicit an overall approval to get this into master.

compiler/analysis/info.go
```diff
@@ -357,6 +358,7 @@ func (fi *FuncInfo) visitCallExpr(n *ast.CallExpr) ast.Visitor {
 func (fi *FuncInfo) callToNamedFunc(callee types.Object) {
 	switch o := callee.(type) {
 	case *types.Func:
+		o = typeparams.OriginMethod(o) // TODO(nevkontakte): Can be replaced with o.Origin() in Go 1.19.
```
flimzy, line 361: I guess we could do this now?

compiler/internal/typeparams/subst/export.go
```diff
@@ -0,0 +1,49 @@
+// Package subst is an excerpt from x/tools/go/ssa responsible for performing
```
flimzy, line 1: I wonder if it might make sense to store these sorts of files in a directory that makes it obvious that it's a vendored/forked upstream file/excerpt.  I've noticed that the `gofumpt` project does [something similar](https://github.com/mvdan/gofumpt/tree/master/internal/govendor), which can make it easier to know where to update files from upstream, as well as a nice indicator to humans and tools (such a linters) not to modify such files.

compiler/internal/typeparams/utils.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil
+	}
+}
+
+var (
+	errInstantiatesGenerics = errors.New("instantiates generic type or function")
+	errDefinesGenerics      = errors.New("defines generic type or function")
+)
+
+// RequiresGenericsSupport an error if the type-checked code depends on
```
flimzy, line 27:
```suggestion
// RequiresGenericsSupport returns an error if the type-checked code depends on
```

compiler/utils.go
```diff
@@ -319,9 +328,30 @@ func isVarOrConst(o types.Object) bool {
 }
 
 func isPkgLevel(o types.Object) bool {
-	return o.Parent() != nil && o.Parent().Parent() == types.Universe
+	// Note: named types are always assigned a variable at package level to be
+	// initialized with the rest of the package types, even the types declared
+	// in a statement inside a function.
```
flimzy, line 333: Is it worth explaining why?

flimzy (approved):
Didn't do a thorough review, since as mentioned, this has been reviewed before, but I did skim through every file and left a few comments.

compiler/analysis/info.go
nevkontakte: Done.

compiler/internal/typeparams/subst/export.go
nevkontakte: Good idea. Created `internal/govendor` and moved `subst` into it.

compiler/internal/typeparams/utils.go
nevkontakte: Done.

compiler/utils.go
nevkontakte: It is, but alas I don't know the answer :)

PR #1294, opened 2024-04-18
[go1.20] Removing generic overrides that are no longer needed

grantnelson-wf:
**We'll hold off removing the overrides until generics is further along**
and we can assess the impact of removing individual overrides

---

With the amazing updates made to master for generics, many of the generics overrides for go1.19 and go1.20 are no longer needed. There are a few that I have not removed because they hit some strange bugs that we can look into in another ticket.

We're getting closer, it was really cool to see `go1.20.14 install -v && gopherjs build -v net/http` pass without issue. `gopherjs test -v fmt log` still fails but that's because of other go1.20 changes that's I'm working through.

This is part of https://github.com/gopherjs/gopherjs/issues/1270


nevkontakte:
I think we may have to hold off on that one until generics are no longer guarded by the GOPHERJS_EXPERIMENT flag. Otherwise the standard library would no longer build for those who don't have the experiment enabled.

grantnelson-wf:
@nevkontakte In the go1.20 branch, the standard libraries don't built without the experiment flag as is. I have a bunch of overrides for a generic-less go1.20 in my [WIP for g1.20](https://github.com/gopherjs/gopherjs/pull/1283). If you want to, I could pull those out and get them a PR for the go1.20 branch. I was getting close to a generic-less go1.20 (with the exception of that `var _ = &Pointer[int]{}` from [go1.20.14/src/sync/atomic/type.go:40](https://cs.opensource.google/go/go/+/refs/tags/go1.20.14:src/sync/atomic/type.go;l=40) that has to be manually removed right now). I wasn't PR'ing them since it felt like we were close enough to target a generic capable go1.20.

So, would you like me to cut some PR's for an almost generic-less go1.20, I could do that? Or I could just hold off on removing these for now but also not add more generic-less overrides, whatever helps the most.

grantnelson-wf:
We'll hold off on these changes until generics and go1.20 is further along.

nevkontakte:
The rule of thumb is that master branch must be usable by the general population at all times, even between releases. As such, we have two options:

 - Implement a generic-less 1.20, which can be merged into master and released as another beta.
 - Assume that we are okay holding off on 1.20 until generics are done. Then we can go all-in on using generics as much as we can in the 1.20 branch, but won't be able to merge it into master as soon.

I don't think I have a strong preference one way or another. It really depends on how deeply embedded generics are in the 1.20 version, and you probably have a better grasp on this than me :)

grantnelson-wf:
I was thinking we'd hold off on 1.20 until generics are done.


The big thing is the `var _ = &Pointer[int]{}` from [go1.20.14/src/sync/atomic/type.go:40](https://cs.opensource.google/go/go/+/refs/tags/go1.20.14:src/sync/atomic/type.go;l=40) that has to be manually removed right now. If we were going to target a generic-less 1.20, we'd have to figure out a way to delete that line from the natives. Since that line doesn't have a unique id for the augmenter to match, we'd either:

- match it by line number (which is not future resilient)
- find it by comparing specifications (which seems like it'd be slow)
- have something that looks for `_` on lhs and checks that the rhs has no function body or anything which can have a side effect, e.g. `var _ = func() int { /* do thing that has side effects */ }()`
- make a PR into go's repo to remove that check

I'm not fond of any of those, so we'll just have to hold off on 1.20 until generics are done.

nevkontakte:
> I was thinking we'd hold off on 1.20 until generics are done.

Alright, I have no objections in this case. I'll take another look at this PR when I have a bit more time, and see if we can merge it.

compiler/natives/src/go/token/position.go
```diff
@@ -1,22 +0,0 @@
...
-
-import "sync"
-
-type FileSet struct {
-	mutex sync.RWMutex
-	base  int
-	files []*File
-
-	// replaced atomic.Pointer[File] for go1.19 without generics.
-	last atomicFilePointer
```
nevkontakte, line 14: Removing this override may be premature, since in their current state generics don't support instantiations across package boundaries.

nevkontakte (changes requested):
Could you please double-check that this PR doesn't actually introduce any regressions? I am pretty sure `atomic.Pointer` overrides are still necessary, and I suspect that at least some of the overrides in the `nistec` package need to stay.

compiler/natives/src/go/token/position.go
grantnelson-wf: It didn't seem to, but I'm still working through the errors for go1.20 in my [WIP PR](https://github.com/gopherjs/gopherjs/pull/1283). There seems to be a few that may be related to removing the crypto overrides and at least one that looks like it may be related to atomic.Pointer.

I was just so excited that generics worked for so many issues that I probably am just being preemptive on this. Like a said before, we can hold off on this. We'll still target a go1.20 with generics. I don't want to add problems while generics is being worked on.

grantnelson-wf:
We'll try to remove the overrides later in smaller chunks once the generics are further along.

go1.20 branch will still need `GOPHERJS_EXPERIMENT="generics"` since the go1.20.14 repo has stuff in it which won't compile without it and we can't (without a lot of effort) override that stuff.

PR #1298, opened 2024-04-25
[go1.20] Added support for unsafe.SliceData

grantnelson-wf:
go1.20 added [`unsafe.SlideData`](https://pkg.go.dev/unsafe@go1.22.2#SliceData), [`unsafe.String`](https://pkg.go.dev/unsafe@go1.22.2#String), and [`unsafe.StringData`](https://pkg.go.dev/unsafe@go1.22.2#StringData).

For the following reasons, this PR only adds `unsafe.SliceData`. The tests for this change are in the Go repo tests of this builtin function. This is part of https://github.com/gopherjs/gopherjs/issues/1270

### SliceData

`unsafe.SliceData` is something that GopherJS can handle because of how, even when empty, GopherJS slices have an array that can be pointed to.

The only thing we can't handle is how using the "unspecified memory address" works right now, [seen in this playground](https://goplay.tools/snippet/vr02QEJUWAG), but there are no promises for how that should work and no one should be using that pointer like that. In GopherJS assigning to the pointer of the empty slice simply writes to `[0]` on the empty slice which then isn't visible, always returns zero, and is a pointless, but doesn't break anything.

The Go compiler ensures that the argument into `SliceData` will be a slice.

### String

`unsafe.String` may be possible to support with `Buffer` but would require copying the data into a string. We should probably override it where possible because the point of this function is to not require a copy to be faster than simply performing a conversion, `string([]byte{...})`.

We may need to we can add this later (with the copy) since it seams like some natives are using this instead of the cast to save memory and/or time. Instead of overriding each location, we could just stub out the cast (kind of like how we didn't override all the atomics but just stubbed out the atomic methods).

### StringData

`unsafe.StringData` may be possible via a copy to a byte array first but is the same as above, doing the copy makes this no better than a conversion. Or we could make some JS code similar to `$indexPtr` that wraps a string into a byte pointer. Until then we should probably override them instead.

Also, this is super unsafe because in Go if you get the string data for a literal and attempt to modify it (despite the comment in the docs saying `*byte` should be treated as immutable) it causes a segmentation fault because you can't modify the literal stored in the memory with the opcode. The fault signal can not be recovered from. [Here's a playground with that](https://goplay.tools/snippet/4BvMksBpaKa).

### DIFF
--- a/compiler/expressions.go
+++ b/compiler/expressions.go
@@ -1059,11 +1059,8 @@ func (fc *funcContext) translateBuiltin(name string, sig *types.Signature, args 
 		sel, _ := fc.selectionOf(astutil.RemoveParens(args[0]).(*ast.SelectorExpr))
 		return fc.formatExpr("%d", typesutil.OffsetOf(sizes32, sel))
 	case "SliceData":
-		// SliceData returns nil if the slice is nil, otherwise returns a pointer to the first index of the array, &s[0].
-		// If the slice is empty (cap == 0), it returns an "unspecified memory address" or in our case, a pointer to the empty array.
 		t := fc.typeOf(args[0]).Underlying().(*types.Slice)
-		elemPtrType := types.NewPointer(t.Elem())
-		return fc.formatExpr("(%1e === %2s.nil) ? %3s.nil : $indexPtr(%1e.$array, %1e.$offset, %3s)", args[0], fc.typeName(t), fc.typeName(elemPtrType))
+		return fc.formatExpr(`$sliceData(%e, %s)`, args[0], fc.typeName(t))
 	default:
 		panic(fmt.Sprintf("Unhandled builtin: %s\n", name))
 	}
--- a/compiler/expressions.go
+++ b/compiler/expressions.go
@@ -1058,6 +1058,12 @@ func (fc *funcContext) translateBuiltin(name string, sig *types.Signature, args 
 	case "Offsetof":
 		sel, _ := fc.selectionOf(astutil.RemoveParens(args[0]).(*ast.SelectorExpr))
 		return fc.formatExpr("%d", typesutil.OffsetOf(sizes32, sel))
+	case "SliceData":
+		// SliceData returns nil if the slice is nil, otherwise returns a pointer to the first index of the array, &s[0].
+		// If the slice is empty (cap == 0), it returns an "unspecified memory address" or in our case, a pointer to the empty array.
+		t := fc.typeOf(args[0]).Underlying().(*types.Slice)
+		elemPtrType := types.NewPointer(t.Elem())
+		return fc.formatExpr("(%1e === %2s.nil) ? %3s.nil : $indexPtr(%1e.$array, %1e.$offset, %3s)", args[0], fc.typeName(t), fc.typeName(elemPtrType))
 	default:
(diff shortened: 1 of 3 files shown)


compiler/expressions.go
```diff
@@ -1058,6 +1058,12 @@ func (fc *funcContext) translateBuiltin(name string, sig *types.Signature, args
 	case "Offsetof":
 		sel, _ := fc.selectionOf(astutil.RemoveParens(args[0]).(*ast.SelectorExpr))
 		return fc.formatExpr("%d", typesutil.OffsetOf(sizes32, sel))
+	case "SliceData":
+		// SliceData returns nil if the slice is nil, otherwise returns a pointer to the first index of the array, &s[0].
+		// If the slice is empty (cap == 0), it returns an "unspecified memory address" or in our case, a pointer to the empty array.
+		t := fc.typeOf(args[0]).Underlying().(*types.Slice)
+		elemPtrType := types.NewPointer(t.Elem())
+		return fc.formatExpr("(%1e === %2s.nil) ? %3s.nil : $indexPtr(%1e.$array, %1e.$offset, %3s)", args[0], fc.typeName(t), fc.typeName(elemPtrType))
```
nevkontakte, line 1066: I think you need a temporary variable here for the `%1e` here, otherwise if it's a complex expression it would be evaluated multiple times, for example: `unsafe.SliceData(readLinesFromFile())` would insert the `readLinesFromFile()` call in every `%1e` location.

nevkontakte (changes requested):

nevkontakte:
In general, I am in favor of implementing `String()` and `StringData()` through the usual type conversion. Although it will cause a copy, it will probably work in more cases than not.

compiler/expressions.go
grantnelson-wf: Oh, great call, I didn't realize it did that for each instance of `%1e`. I'll fix that.

nevkontakte:
> I wonder if `String()` and `StringData()` might be the kinds of things we should support, but emit an optional warning when they are used.

That sounds like a perfect use case for https://github.com/gopherjs/gopherjsvet 😉

compiler/expressions.go
grantnelson-wf: Fixed it so that the `%e` formatting is done prior to the return and the formatting for the return uses `%1s` instead.

flimzy:
> That sounds like a perfect use case for https://github.com/gopherjs/gopherjsvet 😉

I hit a blocker with that project, with a lack of support for abritrary tags in the upstream tools. But I think we can work around it just by forking the upstream (as I've seen done in a few other projects)... so when I have time, I'll try to get that project back off life support.

compiler/expressions.go
nevkontakte: Sorry, when I mentioned a temporary variable, I should've clarified I meant a JS variable. Unfortunately, with your fix the expression would still be evaluated multiple times, since the compiled expression would be inserted into the generated code several times.

However, on a second thought I think a cleaner option would be to define a `$sliceData` helper function in prelude and insert its call instead of the complex expression. This would both generate a more compact code, as well as avoid a multiple evaluation problem. You can see an example of that with the `close()` builtin a few lines above.

grantnelson-wf: Awesome, I'll give that a try.

nevkontakte (approved):

PR #1312, opened 2024-05-28
[go1.20] Updating generics overrides for crypto

grantnelson-wf:
I was holding off on these changes until I could figure out what was wrong with a large part of crypto (turns out the issue was in `uint` and unrelated to the native overrides for crypto). Since that is now merged, here is the updates to the generics native overrides for crypto.

In go1.20 crypto the curves were broken up into nearly identical parts. Our original go1.19 overrides could be reused for most of these changes with minor rework, so that's what I did. I imported the shared override implementation each place the curve was needed.

This also includes an update to the cache removal. These are not generics related, we just can't handle the garbage collect hooks for the crypto cache, so we just make it always a cache miss.

CI will still fail for go1.20. We're missing an unrelated method (I'm working on it as part of another PR), but once that's in the build passes. Then this will remove most, but not all, of the problems in crypto. There is still a buffer overflow issue that I haven't looked into yet but felt unrelated to the curves.

This is part of #1270

### DIFF
--- a/compiler/natives/src/crypto/ecdh/nist.go
+++ b/compiler/natives/src/crypto/ecdh/nist.go
@@ -1,0 +1,58 @@
+//go:build js
+// +build js
+
+package ecdh
+
+import (
+	"crypto/internal/nistec"
+	"io"
+)
+
+//gopherjs:purge for go1.20 without generics
+type nistPoint[T any] interface{}
+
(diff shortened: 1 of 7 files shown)


nevkontakte (approved):

nevkontakte:
Thank you!

PR #1313, opened 2024-05-31
[go1.20] internal/godebug env watcher update

grantnelson-wf:
In `internal/godebug` the `setUpdate` method is used to watch for changes to the environment variable `GODEBUG`. This is used by tests such as the `archive/zip` test `TestInsecurePaths` or any test using [`t.Setenv`](https://pkg.go.dev/testing@go1.22.3#T.Setenv).

To solve this I added a way to inject a JS `Proxy` into `process.env` that calls a callback when any environment variable is set, the callback will handle the rest of the update and call the function provided to `setUpdate`.

Testing this was a little hard since `godebug` is used by tests. I solved this by creating a go:link from a test that is external to `godebug`.

This is part of #1270

### DIFF
--- a/compiler/natives/src/internal/godebug/godebug.go
+++ b/compiler/natives/src/internal/godebug/godebug.go
@@ -5,6 +5,7 @@ package godebug
 
 import (
 	"sync"
+	_ "unsafe" // go:linkname
 )
 
 type Setting struct {
@@ -36,6 +37,9 @@ func (s *Setting) Value() string {
 	})
 	return *s.value.Load()
 }
+
+//go:linkname setUpdate runtime.godebug_setUpdate
+func setUpdate(update func(def, env string))
 
 func update(def, env string) {
 	updateMu.Lock()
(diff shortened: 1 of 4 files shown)


nevkontakte (changes requested):
I'd like to avoid modifying the global environment, since it creates a risk of conflicting with some other library, even if the chance of that happening in stdlib tests is slim.

Since we are already overriding Setenv/Getenv implementations, can we implement the hook at Go level instead?

grantnelson-wf:
That makes sense. I was trying to make the injection such that any other proxy could also inject on top or below it, but it is worrying that it is modifying `process.env`. I'll see if I can implement this inside of `syscall/syscall_js_wasm.go` with the `runtimeSetenv`/`runtimeUnsetenv` methods using just go.

It does appear that anytime this callback is being used is being done via testing `SetEnv` and not something else like command to set environment variables some other way, so having a more broad watcher shouldn't be needed.

I'll rework it.

nevkontakte:
Yeah, I can also imagine runtimes that doesn't allow reassigning `process.env`, or some code trying to use reflection on it and having issues as a result.

nevkontakte (approved):

PR #1315, opened 2024-06-04
[go1.20] Fixing reflect Value.Grow

grantnelson-wf:
Fixing the `reflect` method `Value.Grow` that grows a slice to a contain a minimum specified capacity. This required the [`grow`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.14:src/reflect/value.go;l=2795-2807) method to be updated so that it can call into JS. I split the `append` method to create a `growSlice` method which functions similar to [`reflect.growslice`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.14:src/reflect/value.go;l=3846).

CI will still fail right now because `internal/godebug.setUpdate` needs to be implemented. It is implemented in another PR. With both this and the `setUpdate` change the `TestGrow` tests pass.

This does not fix the issue with `Value.extendSlice` used by `Value.Append`. I'm fixing that in another PR.

This is part of #1270

### DIFF
--- a/compiler/natives/src/reflect/reflect.go
+++ b/compiler/natives/src/reflect/reflect.go
@@ -1325,6 +1325,26 @@ func getJsTag(tag string) string {
 	return ""
 }
 
+func (v Value) grow(n int) {
+	if n < 0 {
+		panic(`reflect.Value.Grow: negative len`)
+	}
+
+	s := v.object()
+	len := s.Get(`$length`).Int()
+	if len+n < 0 {
+		panic(`reflect.Value.Grow: slice overflow`)
+	}
+
+	cap := s.Get(`$capacity`).Int()
+	if len+n > cap {
+		ns := js.Global.Call("$growSlice", s, len+n)
+		s.Set(`$capacity`, ns.Get(`$capacity`))
+		s.Set(`$array`, ns.Get(`$array`))
+		s.Set(`$offset`, ns.Get(`$offset`))
+	}
+}
+
 func (v Value) Index(i int) Value {
 	switch k := v.kind(); k {
 	case Array:
--- a/compiler/natives/src/reflect/reflect.go
+++ b/compiler/natives/src/reflect/reflect.go
@@ -1830,3 +1830,13 @@ func verifyNotInHeapPtr(p uintptr) bool {
 	// always return true.
 	return true
 }
+
+// typedslicecopy is implemented in prelude.js as $copySlice
+//
+//gopherjs:purge
+func typedslicecopy(elemType *rtype, dst, src unsafeheader.Slice) int
+
+// growslice is implemented in prelude.js as $growSlice.
+//
+//gopherjs:purge
+func growslice(t *rtype, old unsafeheader.Slice, num int) unsafeheader.Slice
--- a/compiler/natives/src/reflect/reflect.go
+++ b/compiler/natives/src/reflect/reflect.go
@@ -1339,9 +1339,7 @@ func (v Value) grow(n int) {
 	cap := s.Get(`$capacity`).Int()
 	if len+n > cap {
 		ns := js.Global.Call("$growSlice", s, len+n)
-		s.Set(`$capacity`, ns.Get(`$capacity`))
-		s.Set(`$array`, ns.Get(`$array`))
-		s.Set(`$offset`, ns.Get(`$offset`))
+		js.InternalObject(v.ptr).Call("$set", ns)
 	}
 }
 
--- a/compiler/prelude/prelude.js
+++ b/compiler/prelude/prelude.js
@@ -428,55 +428,75 @@ var $appendSlice = (slice, toAppend) => {
     return $internalAppend(slice, toAppend.$array, toAppend.$offset, toAppend.$length);
 };
 
+// Internal helper function for appending to a slice.
+// The given slice will not be modified.
+//
+// If no values are being appended, the original slice will be returned.
+// Otherwise, a new slice will be created with the appended values.
+//
+// If the underlying array has enough capacity, it will be used.
+// Otherwise, a new array will be allocated with enough capacity to hold
+// the new values and the original array will not be modified.
 var $internalAppend = (slice, array, offset, length) => {
     if (length === 0) {
         return slice;
     }
 
-    var newLength = slice.$length + length;
-    slice = $growSlice(slice, newLength);
+    let newLength = slice.$length + length;
+    let newSlice = $growSlice(slice, newLength);
     
-    var newArray = slice.$array;
-    $copyArray(newArray, array, slice.$offset + slice.$length, offset, length, slice.constructor.elem);
-
-    var newSlice = new slice.constructor(newArray);
-    newSlice.$offset = slice.$offset;
+    let newArray = newSlice.$array;
+    $copyArray(newArray, array, newSlice.$offset + newSlice.$length, offset, length, newSlice.constructor.elem);
+   
     newSlice.$length = newLength;
-    newSlice.$capacity = slice.$capacity;
     return newSlice;
 };
 
+// Calculates the new capacity for a slice that is expected to grow to at least
+// the given minCapacity. This follows the Go runtime's growth strategy.
+// The oldCapacity is the current capacity of the slice that is being grown.
 const $calculateNewCapacity = (minCapacity, oldCapacity) => {
     return Math.max(minCapacity, oldCapacity < 1024 ? oldCapacity * 2 : Math.floor(oldCapacity * 5 / 4));
 };
 
+// Potentially grows the slice to have a capacity of at least minCapacity.
+//
+// A new slice will always be returned, even if the given slice had the required capacity.
+// If the slice didn't have enough capacity, the new slice will have a
+// new array created for it with the required minimum capacity.
+//
+// This takes the place of the growSlice function in the reflect package.
 var $growSlice = (slice, minCapacity) => {
-    const oldCapacity = slice.$capacity;
-    if (minCapacity <= oldCapacity) {
-        return slice
-    }
-     
-    const newCapacity = $calculateNewCapacity(minCapacity, oldCapacity);
-    
-    let newArray;
-    if (slice.$array.constructor === Array) {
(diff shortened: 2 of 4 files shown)


compiler/natives/src/reflect/reflect.go
```diff
@@ -1325,6 +1325,26 @@ func getJsTag(tag string) string {
...
+	}
+
+	s := v.object()
+	len := s.Get(`$length`).Int()
+	if len+n < 0 {
+		panic(`reflect.Value.Grow: slice overflow`)
+	}
+
+	cap := s.Get(`$capacity`).Int()
+	if len+n > cap {
```
flimzy, line 1340: Do we need a `if cap+n < 0 {` overflow check, too?

flimzy (commented):
One comment, otherwise LGTM.

compiler/natives/src/reflect/reflect.go
grantnelson-wf: I based this update off of the original code...
```Go
func (v Value) grow(n int) {
	p := (*unsafeheader.Slice)(v.ptr)
	switch {
	case n < 0:
		panic("reflect.Value.Grow: negative len")
	case p.Len+n < 0:
		panic("reflect.Value.Grow: slice overflow")
	case p.Len+n > p.Cap:
		t := v.typ.Elem().(*rtype)
		*p = growslice(t, *p, n)
	}
}
```
...but it is always better to check our assumptions. So, here is my thought train (ramblings to myself):

We know:
- $len \le cap$, so we could say $len + x = cap$ where $x \ge 0$.
- $len + n \gt cap$ otherwise we won't grow, so the smallest $n$ that will still cause a grow is $cap - len + 1$.
- To have a negative value we must roll over past a $max$ value.

Okay, if we get a big $x$ by setting $len = 0$, we only need $n = cap + 1$ to grow.
If we set $cap = \frac{max}{2}$ (or any value smaller than $max$ but still with room to grow),
Then $len + n = 0 + (cap +1) = cap + 1 \lt max$
But $cap + n = cap + (cap + 1) = 2 cap + 1 = 2 \frac{max}{2} + 1 = max + 1 \gt max$.
Therefore, we can have a $len + n$ which isn't too big but still more than $cap$, and a $cap + n$ that is too big. Since we grow to $len + n$ we would be getting a false positive by checking $cap + n$.

I'm not sure if that stroll through my thoughts make sense (I can try to clean it up if you'd like), but it seems like we don't need to check `cap+n < 0` since `cap + n` may be much larger than `len + n` and we are only growing to `len + n`.

grantnelson-wf:
### HOLD

I think I found an issue with how I implemented grow. I'm going to triple check and add some tests. When `TestGrow` is run before `TestFieldByName` the latter fails, but both pass when run by themselves. I think the issue is that `TestGrow` runs

```Go
v := ValueOf(new([]int)).Elem()
v.Grow(1)
```

Then in TestFieldByName there is a `[]int` which is expected to be empty, but when run together it is `[]int{ 0 }`. I'm not sure but I think grow is modifying the `nil` prototype for `[]int` in JS somehow so that the next time an empty `[]int` is gotten, it isn't empty.

The following is a rough test which I'm going to turn into a better test for this PR:

```Go
v1 := ValueOf(new([]int)).Elem()
print("v0 = ", v1.Interface().([]int), "\n")
v1.Grow(1)
print("v1 = ", v1.Interface().([]int), "\n")

v2 := ValueOf(new([]int)).Elem()
print("v2 = ", v2.Interface().([]int), "\n")
```
That outputs
```Plain
v0 =  <ref *1> typ {
  '$array': Int32Array(0) [],
  '$offset': 0,
  '$length': 0,
  '$capacity': 0,
  '$val': [Circular *1]
} 
v1 =  <ref *1> typ {
  '$array': Int32Array(1) [ 0 ],
  '$offset': 0,
  '$length': 0,
  '$capacity': 1,
  '$val': [Circular *1]
} 
v2 =  <ref *1> typ {
  '$array': Int32Array(1) [ 0 ],   //< This seems wrong and I suspect that should be `Int32Array(0) []`
  '$offset': 0,
  '$length': 0,
  '$capacity': 1,
  '$val': [Circular *1]
} 
```

compiler/natives/src/reflect/reflect.go
```diff
@@ -1325,6 +1325,26 @@ func getJsTag(tag string) string {
...
+		ns := js.Global.Call("$growSlice", s, len+n)
+		s.Set(`$capacity`, ns.Get(`$capacity`))
+		s.Set(`$array`, ns.Get(`$array`))
+		s.Set(`$offset`, ns.Get(`$offset`))
```
grantnelson-wf, line 1344: @nevkontakte and @flimzy, I'm going to need some advice/help with this one. I can't figure out how to assign the `Value` to the new grown slice without either, modifying the `typ.nil` for the slice type, or the variable inside the `Value` isn't updated. I think that we may need to make a copy of the `nil` slice at some point, but I don't know enough about how all the JS works together to figure it out yet.

I added a test and have been using
```Bash
go1.20.14 install . && gopherjs test ./tests -v --run TestNilPrototypeNotModified
```
with generics enabled. After adding the `Object.freeze(typ.nil)` the tests return the following:
```
=== RUN   TestNilPrototypeNotModifiedByPointer
--- PASS: TestNilPrototypeNotModifiedByPointer (0.00s)
=== RUN   TestNilPrototypeNotModifiedByReflectGrow
--- FAIL: TestNilPrototypeNotModifiedByReflectGrow (0.00s)
TypeError: Cannot assign to read only property '$capacity' of object '#<typ>'
    at Object.$packages.reflect.Value.ptr.grow (/reflect/gopherjs__reflect.go:1342:4)
    at Object.$packages.reflect.Value.ptr.Grow (/reflect/value.go:2792:3)
    at TestNilPrototypeNotModifiedByReflectGrow (/github.com/Workiva/gopherjs/tests/arrays_test.go:111:3)
    ...
FAIL    github.com/gopherjs/gopherjs/tests      0.226s
```
without the freeze it will fail the capacity check.

nevkontakte: Actually, freezing nil values by default seems like a very good idea to catch correctness bugs. In theory, to grow a nil slice we need to allocate a new slice of the given capacity, basically the equivalent of `make([]T, 0, X)`. I'll have to poke around the code more to get more specific though.

compiler/prelude/prelude.js
```diff
@@ -433,37 +433,52 @@ var $internalAppend = (slice, array, offset, length) => {
...
+    var newArray = slice.$array;
+    $copyArray(newArray, array, slice.$offset + slice.$length, offset, length, slice.constructor.elem);
 
     var newSlice = new slice.constructor(newArray);
-    newSlice.$offset = newOffset;
+    newSlice.$offset = slice.$offset;
     newSlice.$length = newLength;
-    newSlice.$capacity = newCapacity;
+    newSlice.$capacity = slice.$capacity;
     return newSlice;
```
nevkontakte, line 446: Considering $growSlice already returns a new slice it seems suboptimal to create it once again. I understand that it's necessary to avoid modifying the original slice object when the slice didn't need growing, but it feels kind of inelegant.

compiler/natives/src/reflect/reflect.go
nevkontakte: Okay, I think I understand what's the problem here. This fragment of the code tries to modify the slice object that's contained within `reflect.Value()`, and if that object corresponds to a nil slice, bad things happen.

What we need to do is to _replace_ the object with the new one returned by `$growSlice()` instead of copying its properties. I think all you need to do is `js.InternalObject(v.ptr).Call("$set", ns)`.

nevkontakte (changes requested):

compiler/natives/src/reflect/reflect.go
grantnelson-wf: That worked, I thought I had tried something like that but I must of done it wrong, oh well. Thank you so much for helping with that.

compiler/prelude/prelude.js
```diff
@@ -433,37 +433,52 @@ var $internalAppend = (slice, array, offset, length) => {
...
+        for (let i = slice.$length; i < newCapacity; i++) {
+            newArray[i] = zero();
+        }
+    } else {
+        newArray = new slice.$array.constructor(newCapacity);
+        newArray.set(slice.$array.subarray( slice.$offset,  slice.$offset + slice.$length));
+    }
+
+    const oldLength = slice.$length;
+    slice = new slice.constructor(newArray);
```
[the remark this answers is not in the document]
grantnelson-wf: Renamed and commented

compiler/prelude/prelude.js
grantnelson-wf: Added a check to see if `$growSlice` would create a new slice so that only one new slice is made. I felt like that would work better than trying to return a tuple, doing an "old slice/ new slice" comparison to see if it changed, or duplicating code. I like this better even though it does the same check twice. Is this more what you were thinking for a better solution?

grantnelson-wf:
Ready for Review again

compiler/prelude/prelude.js
nevkontakte: Would things get simpler if `$growSlice()` always returned a new slice, even if it wasn't grown? Not a big deal either way, just trying to streamline the code a bit.

grantnelson-wf: Me thinking it through: When [`func (v Value) Grow(n int)`](https://pkg.go.dev/reflect@go1.20.14#Value.Grow) is called with an $n$ such that $len + n \lt cap$, it has no effect the slice since the slice already has enough capacity to handle $len +n$. In that case having `$growSlice()` always return a new slice creates an unneeded slice, however we check for that case in the Go part of Grow prior to ever calling `$growSlice()`, so always returning a new slice won't effect Grow but would make the append logic simpler.

Okay, I'll make that change.

grantnelson-wf: Changed

nevkontakte: Yep, that looks much cleaner. Thanks!

nevkontakte (approved):

PR #1317, opened 2024-06-20
Another attempt to move tests from CircleCI to GitHub Actions 

grantnelson-wf:
I thought I'd take a swing at moving from CircleCI to GHA. I'm taking some ideas and solutions from #1243.
I'm new to GHA and CircleCI so this has a lot of room for improvement. Any ideas, feedback, and help is greatly appreciated.

- [x] Move environment setup
- [x] Move initial build test on Ubuntu
- [x] Move lint so that the Go versions constant could be shared. (This could have also been done with [config vars for multiple workflows](https://docs.github.com/en/actions/learn-github-actions/variables#defining-configuration-variables-for-multiple-workflows))
- [x] Move MacOS (Darwin) smoke test
- [x] Move Windows smoke test
- [x] Move Ubuntu tests
- [x] Move Go repo tests
- [x] Move Go issue/fixbugs tests
- [x] Move TodoMVC comparison between GOPATH and Go Modules output

---

_Note: Part of the reason I build this WIP PR is because Workiva's fork is under our enterprise plan, meaning it has different rules for actions. For example [it allows different sized runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners) but Workiva doesn't allow 3rd party tools like golangci-lint without several additional steps. So to properly test it I have to have a PR in gopherjs/gopherjs under gopherjs's cloud plan. Additionally, I did this as a separate PR from #1243 because I don't have push permission to that branch and, therefore, needed my own branch._


.github/actions/setup-gopherjs/action.yml
```diff
@@ -0,0 +1,68 @@
+name: Setup GopherJS
+description: Sets up Go, Node.js, and GopherJS
+
+inputs:
+  includeSyscall:
```
nevkontakte, line 5: TBH, I'm entirely happy to drop the node-syscall extension from the CI. It's meed deprecated since GopherJS 1.18, and I was going to remove it in the next stable release entirely.

nevkontakte:
@grantnelson-wf thanks for doing this! FYI, I'm going to unsubscribe from this issue for now to stop it flooding my inbox, but please feel free to @ me whenever you need my input 🙂

.github/actions/setup-gopherjs/action.yml
grantnelson-wf: That makes sense. Should we do the same with the `GO111MODULE` tests as well?

grantnelson-wf:
@nevkontakte and @flimzy I've got the GHA CI working!

I did leave some things out, like some of the report generation and the comparison of TodoMVC builds. I couldn't tell which reports were for CircleCI logging and which, if any, were not. I also didn't bump versions much because of things like the legacy syscall test and a problem in `Example_compressingReader` with node version 20.

What else should we be doing for CI and should I leave the `circle.yaml` file for now?

.github/actions/setup-gopherjs/action.yml
flimzy: It may not be relevant for GopherJS users, but there are still (rare) legitimate times when it makes sense to turn off modules, IIRC.  It's a bit unfortunate that the way to do that is with a flag that's ostentiably outdated by 5+ years.

Curious your opinion, @nevkontakte

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,183 @@
+name: CI
```
flimzy, line 1: 100% pure personal preference here. Do with it as you wish:

I like to keep my tests and SA/linters in separate actions files, simply to make it obvious at a glance where a failure occurred.  And then name them accordingly.

I.e. one called 'Tests' and one called 'Linter' or 'Static Analysis'.

If we do that, I'm not sure where that leaves `todomvc_check` to go.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,183 @@
...
+      - name: Setup GopherJS
+        uses: ./.github/actions/setup-gopherjs/
+      - name: Run GopherJS tests 
+        working-directory: ${{ env.GOPHERJS_PATH }}
+        run: |
+          PACKAGE_NAMES=$( \
+              GOOS=js GOARCH=wasm go list std github.com/gopherjs/gopherjs/js/... github.com/gopherjs/gopherjs/tests/... \
+              | grep -v -x -f .std_test_pkg_exclusions \
+            )
+          gopherjs test -p 2 --minify -v --short $PACKAGE_NAMES
```
flimzy, line 168: I'm curious how you got this to run fast enough to not hit the GHA timeout.

flimzy (commented):
Thanks for picking this up! I'll be happy to be rid of CircleCI!

LGTM, aside from the open question regarding `GO111MODULE`

.github/actions/setup-gopherjs/action.yml
nevkontakte: AFAIK, Go team decided to support GOPATH indefinitely, so we have to test both ways too, at least unless we find a really compelling reason to drop it.

nevkontakte:
The purpose of the diff test for the TODO MVC app is to verify that our GOPATH support doesn't regress. The same app build on GOPATH or Modules mode should produce the same output, which ensures that the GOPATH build is at least as correct as the better-tested and exercised Go Modules mode. Since upstream decided to keep supporting GOPATH fore the foreseeable future, we should continue doing so as well.

.github/workflows/ci.yaml
grantnelson-wf: [GHA's usage limits](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits) says that a job is limited to 6 hours by default. (We can limit it more with [timeout-minutes](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) if we need something to timeout before 6 hours.) For this latest commit [the run times for the jobs](https://github.com/gopherjs/gopherjs/actions/runs/9769066747/usage?pr=1317) say the GopherJS Tests job took about 40 mins.

I don't know how much overhead the `go-junit-report` added to convert the output logs for CircleCI but we no longer have to do that. I'm also not running `circleci tests split --split-by=timings --timings-type=classname` thing since I were not using CircleCI. Not sure if those shaved milliseconds or minutes off of the total time.

I also don't know what environment CircleCI is using. According to [the default runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) we get 4 cores and 16 GB of RAM for the GopherJS Test job. (Meaning we could probably change `-p 2` to `-p 4`.)

flimzy: > job is limited to 6 hours

Interesting. My very fuzzy memory was that it was about 10 minutes. But maybe we had imposed that upon ourselves.

> about 40 mins

:fearful: 

Yipes. That's a long time.  Maybe that's why we had run them in parallel in Circle, just to not be so slow.  I wonder how much we care about that going forward.

.github/workflows/ci.yaml
grantnelson-wf: That makes sense, personally I'm fine with the naming right now:
image

If we put the `Lint Checks` into it's own workflow file it would look like `Measure ...` and just not have the `CI/` prefix.

The reason I put the linter into this workflow is because it has the same `on: push:` settings (where `Measure ...` has a different one), so that all the jobs that are run together are in one workflow file (_not that important_). Also so that we could share the same `env` values. Sharing values means way we have fewer locations to update versions and experiment flags in (_doesn't happen too often to do not an issue_).

I'm cool with breaking out the Linter or leave it alone in this PR, either works for me.

flimzy: That looks fine to me.

.github/actions/setup-gopherjs/action.yml
grantnelson-wf: Cool, that's good to know. I'll move over the TodoMVC GO111MODULE build comparison job too.

.github/workflows/ci.yaml
grantnelson-wf: 40 mins is painful. I'll see if `-p 4` improves it. We could also split the list of package names in 2 or more. Cutting it by number of packages wouldn't balance the jobs but it would run faster. The usage says we can have up to 20 concurrent jobs. One redeeming quality of GHA is that it runs all jobs in parallel (unless we tell it otherwise) so we get all the lint and smoke tests as feedback in only a few minutes.


One of the embedded software companies I worked at did a full build of a custom linux, gcc, a bunch of other tools, our app, then ran tests. In total it took 5 hours. I'd only run it right before heading home for the day unless I had a good book. At Workiva we also have a few hour long CI jobs for testing all the consumers of a common repo. We need a new version of [the XKCD compiling comic](https://imgs.xkcd.com/comics/compiling.png) where it says "waiting for CI to pass". 😆

flimzy: > We could also split the list of package names in 2 or more. Cutting it by number of packages wouldn't balance the jobs but it would run faster.

This was the hangup I ran into with my attempt.  I found some tools aide in balanced splitting of these jobs, as CircleCI does, but had difficulty getting any of them to work.

nevkontakte: Yeah, balanced test splitting was a major reason we used the `go-junit-report` and `circleci tests split` combo. There are a few outlier packages which run unreasonably slow tests. A few years ago I did some optimizations to make some of the slower tests pass faster, but since then new Go versions added more slow tests and I haven't gotten around to fixing them.

Part of the idea in running the tests in several smaller machines instead of one big one is that we'd only accrue execution minutes for a small container while finishing those outlier tests, instead of a whole multi-core machine. I don't know how much run time we get on an opensource plan, if we managed to stay within limits, it's probably fine to use a bigger machine with more parallelism.

flimzy: Another option: I've never tried running my own runners for GitHub, but on GitLab it's quite simple.  I have a ~3 EUR/mo VM at hetzner.de that I use to run GitLab runners for a large number of my personal projects.

A cheap VM like that wouldn't speed up execution, but it would give us effectively unlimited CI minutes at a low monthly cost (which I would be happy to cover, if we decide to go that route).

grantnelson-wf: It would be awesome if we could get it to work without costing anything and I think we can do that, if I'm reading and understanding [the billing page](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes) correctly. It doesn't sound like there are monthly limits on public free repos:

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes ...

If it wasn't free then big OSS with lots of PRs per month would probably leave Github, people could do a DoS attack by opening a bunch of PRs and using up the limit quickly, and I have used a lot of minutes working on this CI PR but haven't hit a cap yet 🤞 (I'm really sorry if I'm wrong and just consumed a lot of mins for this PR).

The limit for free (I'm assuming private) repos is 2,000 mins per month, and since each full CI run takes an accumulated 50 mins for all the jobs to finish, that'd mean we could only run 40 full CI runs per month. However, (again assuming I understand this correctly) the minutes are the sum of all jobs' minutes, not just the longest running job, meaning if we broke up the 40 min job into 4 x 10 min jobs it would still count as the same amount of minutes used, plus the overhead of setting up the runner.

If it turns out that GopherJS does have a limit per month as a public repo, we could put all ubuntu jobs (linter, smoke tests, full tests) into the same job so that the setup and installs only happens once, we could put things like the full tests into a workflow that runs only on demand so we can run it before a merge but not on every push, and we could then setup a self-hosted runner like flimzy said. (If GopherJS ever costs money, I'm not sure how it works but I did notice some repos, like [templ](https://github.com/a-h/templ), have a "Sponsor this project" that might offset costs, but it probably would require a gopherjs LLC.)

grantnelson-wf: Well, that's good to know it's 0. That's also why I put so much equivocation above; it's seems vague and is confusing to me.

.github/actions/setup-gopherjs/action.yml
grantnelson-wf: Added the TodoMVC GO111MODULE build comparison

.github/workflows/ci.yaml
grantnelson-wf: I ran an experiment where I just did a simple break up of the packages into even number of packages (I know you said that isn't great but I was curious how bad the balance was... you were right). I broke it up into 4 parts and was hoping for 10 mins +/- 2 mins. [The results](https://github.com/gopherjs/gopherjs/actions/runs/9880583597/job/27289482414?pr=1317) showed that's not great because the vast majority of the time is spent in the crypto packages, about 28 mins. Meaning to balance the tests better we'd need to use prior runs to estimate the next runs like [go-test-split-action](https://github.com/hashicorp-forge/go-test-split-action) says it can do using junit, sounds similar to what you had setup for CircleCI.

grantnelson-wf: I took the test results from one of the runs and sorted it by duration. The worst two packages are the `crypto/internal/edwards25519` package taking 22m 3.455s and the `crypto/internal/edwards25519/field` package taking 25m 22.895s!

<details>
  <summary>Test Measurements</summary>

| packages | secs |
|:--|--:|
crypto/internal/edwards25519/field | 1,522.896
crypto/internal/edwards25519 | 1,323.456
image/gif | 498.345
crypto/x509 | 268.931
net/http | 242.588
crypto/tls | 156.691
compress/gzip | 74.228
crypto/internal/nistec | 70.028
crypto/ed25519 | 37.949
math/big | 34.969
compress/flate | 30.578
image/jpeg | 20.398
archive/tar | 14.639
encoding/pem | 13.955
mime/multipart | 13.229
go/doc/comment | 12.941
strconv | 12.609
regexp/syntax | 12.074
crypto/elliptic | 11.493
time | 10.504
index/suffixarray | 8.334
crypto/ecdsa | 8.148
github.com/gopherjs/gopherjs/tests | 7.824
go/printer | 7.306
crypto/rsa | 6.126
go/parser | 5.847
encoding/json | 4.629
image/png | 4.375
image/draw | 4.183
reflect | 3.722
archive/zip | 3.703
bytes | 3.391
crypto/rc4 | 3.348
html/template | 2.773
image | 2.689
go/doc | 2.651
bufio | 2.541
database/sql | 2.498
regexp | 2.303
os | 2.116
debug/dwarf | 2.003
net/smtp | 1.928
testing/quick | 1.916
text/template | 1.705
image/color | 1.684
math/rand | 1.633
internal/trace | 1.600
testing | 1.566
debug/pe | 1.554
hash/maphash | 1.551
container/heap | 1.522
net | 1.505
debug/elf | 1.470
io/fs | 1.376
crypto/rand | 1.375
fmt | 1.367
path/filepath | 1.346
strings | 1.341
net/netip | 1.321
encoding/gob | 1.298
sync | 1.264
sort | 1.216
internal/xcoff | 1.213
os/exec | 1.167
compress/bzip2 | 1.039
go/token | 1.015
compress/zlib | 1.013
text/template/parse | 1.007
syscall | 1.001
net/rpc/jsonrpc | 0.907
encoding/base64 | 0.896
net/http/cgi | 0.894
net/http/fcgi | 0.876
sync/atomic | 0.862
go/format | 0.837
net/http/cookiejar | 0.831
crypto/des | 0.817
crypto/cipher | 0.811
context | 0.801
encoding/ascii85 | 0.793
encoding/csv | 0.788
internal/fuzz | 0.783
encoding/base32 | 0.770
math | 0.758
embed | 0.755
expvar | 0.753
math/bits | 0.741
net/http/httptrace | 0.735
compress/lzw | 0.732
net/url | 0.724
crypto/dsa | 0.700
internal/poll | 0.690
go/ast | 0.671
text/scanner | 0.669
crypto/sha512 | 0.664
crypto/hmac | 0.661
debug/gosym | 0.658
io | 0.647
debug/macho | 0.641
encoding/binary | 0.622
go/constant | 0.622
internal/reflectlite | 0.620
crypto/subtle | 0.614
text/tabwriter | 0.614
go/internal/gcimporter | 0.610
mime | 0.610
encoding/hex | 0.600
crypto/aes | 0.596
unicode | 0.581
go/build/constraint | 0.570
container/list | 0.567
crypto | 0.566
crypto/md5 | 0.566
encoding/asn1 | 0.559
net/mail | 0.557
hash/crc64 | 0.554
syscall/js | 0.551
hash | 0.549
hash/crc32 | 0.539
flag | 0.535
unicode/utf8 | 0.535
net/textproto | 0.526
go/importer | 0.524
crypto/sha1 | 0.522
crypto/internal/boring/bcache | 0.521
os/user | 0.521
io/ioutil | 0.519
testing/fstest | 0.519
go/internal/gccgoimporter | 0.516
log | 0.515
mime/quotedprintable | 0.514
errors | 0.504
testing/iotest | 0.500
crypto/internal/boring | 0.493
crypto/sha256 | 0.489
container/ring | 0.481
hash/adler32 | 0.475
go/scanner | 0.465
hash/fnv | 0.459
math/cmplx | 0.457
embed/internal/embedtest | 0.447
html | 0.439
net/http/internal | 0.431
path | 0.430
internal/profile | 0.423
debug/plan9obj | 0.416
internal/diff | 0.411
crypto/internal/nistec/fiat | 0.407
runtime/metrics | 0.399
unicode/utf16 | 0.394
internal/fmtsort | 0.390
debug/buildinfo | 0.389
internal/buildcfg | 0.389
database/sql/driver | 0.386
internal/singleflight | 0.378
os/exec/internal/fdtest | 0.378
internal/itoa | 0.365
internal/unsafeheader | 0.363
log/syslog | 0.360
net/http/internal/ascii | 0.359
internal/safefilepath | 0.358
os/signal | 0.353
github.com/gopherjs/gopherjs/tests/gorepo | 0.348
github.com/gopherjs/gopherjs/tests/main | 0.347
crypto/internal/subtle | 0.346
internal/godebug | 0.346
internal/cpu | 0.344
runtime/internal/sys | 0.344
runtime/internal/math | 0.341
plugin | 0.339
github.com/gopherjs/gopherjs/js | 0.338

</details>

So I creates jobs to run "crypto" and "non-crypto" GopherJS tests. The non-crypto job takes 13m 7s and crypto job takes 28m 37s. This way of breaking up the tests isn't as balanced as it could be, but it works fairly well without having to try to store the previous run's timings as an artifact. I setup the job matrix so that we could split out other groups of tests into different jobs, if we want to.

grantnelson-wf:
Ready for review again

- I broke up GopherJS Test job into crypto and non-crypto
- I updated the TodoMVC test job to do the module build comparison
- I updated documentation

.github/workflows/ci.yaml
flimzy: Would it make sense to filter on `edwards25519` specifically?  Maybe we can get the long-running task down by a couple more minutes?

Either way, this is indeed a big improvement, for small effort. :+1:

nevkontakte: I did some googling and it does seem like there is [no actions limit on public repos](https://github.com/orgs/community/discussions/70492#discussioncomment-7362186), including Windows and Mac runners, as long as we are using standard runners. So at this point I'm reasonably convinced we won't run into billing issues if we lean on GHA more heavily.

Splitting crypto tests into a separate workflow is a reasonable fix for the short term, although at some point we should figure out what's so slow about it. I feel like users may thank us for a performance improvement :)

.github/actions/setup-gopherjs/action.yml
nevkontakte: LGTM, thanks!

nevkontakte (approved):
I'm happy with the current setup, thanks @grantnelson-wf! @flimzy unless you have any other suggestions, I am happy to merge.

tests/gorepo/gorepo_test.go
```diff
@@ -24,7 +24,7 @@ func TestGoRepositoryCompilerTests(t *testing.T) {
 	shards := os.Getenv("CIRCLE_NODE_TOTAL")
 	shard := os.Getenv("CIRCLE_NODE_INDEX")
 	if shards != "" && shard != "" {
-		// We are running under CircleCI parallel test job, so we need to shard execution.
+		// When we are running under CircleCI parallel test job, we need to shard execution.
```
flimzy, line 27: Should this be deleted, or is that a follow-up task after Circle CI is finally removed?

flimzy (approved):

tests/gorepo/gorepo_test.go
grantnelson-wf: I was hoping that we could get another pass at CI work sometime to do some cleanup and tryout features like [go-test-split-action](https://github.com/hashicorp-forge/go-test-split-action). I'm sure there are some other things we haven't noticed yet that we'll run across since GHA is fairly new to us.

grantnelson-wf: Oooh, another one which I literally ran across this morning about 5 mins ago. During my teams standup I was telling my team about what I was doing. For my [WIP] PR I have a copy of this GHA bumped to go1.20. I made a comment that I only have a few fails left but wasn't sure because CircleCI had been timing out and the GHA was cancelling `GopherJS Test (crypto)` when non-crypto failed. A QA on our team said "just add `fail-fast: false`." After looking at the [fail-fast docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast), they're right, we should probably add that into GopherJS's GHA. Things like that we wouldn't know about that until we were having some failing tests. I'll add it into the WIP PR to try it out as something we can add in a follow up with any other improvements we run across.

nevkontakte: I am okay with keeping it for now. It doesn't add _that_ much complexity and would make a rollback easier if we have to.

nevkontakte: @grantnelson-wf actually, mind adding it into this PR too? I've updated repo settings to require GHA checks instead of CircleCI, and I'd like to make sure the CircleCI Pipeline stops showing up after you push something into this PR.

grantnelson-wf: Can do!

grantnelson-wf: Updated `TestGoRepositoryCompilerTests` to not have the CircleCI flags.
Added in `fail-fast: false` to let all the jobs of GopherJS Tests finish even when one fails.

nevkontakte (approved):

tests/gorepo/gorepo_test.go
nevkontakte: LGTM, thank you!

PR #1318, opened 2024-06-23
Move function translation code into a separate file.

nevkontakte:
The logic is already very complicated and breaking it up into more manageable methods will improve readability.

 - Broke out nestedFunctionContext() and translateStandaloneFunction()
 - Removed the need to return function param names from translateFunctionBody()

(based on commit 2e9a5ce26e5938a33e449d144ed6751a296373c6)


compiler/functions.go
```diff
@@ -0,0 +1,320 @@
...
+	"github.com/gopherjs/gopherjs/compiler/astutil"
+	"github.com/gopherjs/gopherjs/compiler/internal/typeparams"
+	"github.com/gopherjs/gopherjs/compiler/typesutil"
+)
+
+// newFunctionContext creates a new nested context for a function corresponding
+// to the provided info and instance.
+func (fc *funcContext) nestedFunctionContext(info *analysis.FuncInfo, sig *types.Signature, inst typeparams.Instance) *funcContext {
+	if info == nil {
+		panic(fmt.Errorf("missing *analysis.FuncInfo"))
```
flimzy, line 24: optional

```suggestion
		panic(errors.New("missing *analysis.FuncInfo"))
```

flimzy (approved):

compiler/functions.go
nevkontakte: I tend to use `fmt.Errorf()` even by default, because in practice it's likely at the error message would be extended at some point anyway and I'd have to update the function. `errors.New()` is really only worth it when you don't want to depend on the `fmt` package for some reason, and our code already depends on it.

PR #1322, opened 2024-07-12
[go1.20] Removing remaining uses of `unsafeheader.Slice` from reflect

grantnelson-wf:
Overrode the remaining uses of `unsafeheader.Slice` since GopherJS implemented slices differently. The remaining use was in `cvtSliceArray` used by methods like [Value.Convert](https://pkg.go.dev/reflect@go1.20.14#Value.Convert).

The CI will still not pass yet, but this will fix the tests like `TestConvert` so they pass. With the prior tickets for reflect merged, this should also get the `reflect` package to pass.

This is a follow up to #1321. This is part of #1270

### DIFF
--- a/compiler/natives/src/internal/unsafeheader/unsafeheader.go
+++ b/compiler/natives/src/internal/unsafeheader/unsafeheader.go
@@ -1,0 +1,16 @@
+//go:build js
+// +build js
+
+package unsafeheader
+
+// Slice and String is Go's runtime representations which is different
+// from GopherJS's runtime representations. By purging these types,
+// it will prevent failures in JS where the code compiles fine but
+// expects there to be a constructor which doesn't exist when casting
+// from GopherJS's representation into Go's representation.
+
+//gopherjs:purge
+type Slice struct{}
+
+//gopherjs:purge
+type String struct{}
--- a/compiler/natives/src/reflect/reflect.go
+++ b/compiler/natives/src/reflect/reflect.go
@@ -832,6 +832,21 @@ func cvtSliceArrayPtr(v Value, t Type) Value {
 	return Value{t.common(), unsafe.Pointer(array.Unsafe()), v.flag&^(flagIndir|flagAddr|flagKindMask) | flag(Ptr)}
 }
 
+// convertOp: []T -> [N]T
+func cvtSliceArray(v Value, t Type) Value {
+	n := t.Len()
+	if n > v.Len() {
+		panic("reflect: cannot convert slice with length " + itoa.Itoa(v.Len()) + " to array with length " + itoa.Itoa(n))
+	}
+
+	slice := v.object()
+	dst := MakeSlice(SliceOf(t.Elem()), n, n).object()
+	js.Global.Call("$copySlice", dst, slice)
+
+	arr := dst.Get("$array")
+	return Value{t.common(), unsafe.Pointer(arr.Unsafe()), v.flag&^(flagAddr|flagKindMask) | flag(Array)}
+}
+
 func Copy(dst, src Value) int {
 	dk := dst.kind()
 	if dk != Array && dk != Slice {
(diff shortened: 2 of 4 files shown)


compiler/natives/src/reflect/reflect.go
```diff
@@ -832,6 +832,21 @@ func cvtSliceArrayPtr(v Value, t Type) Value {
 	return Value{t.common(), unsafe.Pointer(array.Unsafe()), v.flag&^(flagIndir|flagAddr|flagKindMask) | flag(Ptr)}
 }
 
+// convertOp: []T -> [N]T
+func cvtSliceArray(v Value, t Type) Value {
```
nevkontakte, line 836: Just to double-check, does the upstream also copy the data, or return a pointer to the shared array?

compiler/natives/src/internal/unsafeheader/unsafeheader.go
```diff
@@ -0,0 +1,16 @@
+//go:build js
+// +build js
+
+package unsafeheader
+
+// Slice and String is Go's runtime representations which is different
+// from GopherJS's runtime representations. By purging these types,
+// it will prevent failures in JS where the code compiles fine but
+// expects there to be a constructor which doesn't exist when casting
+// from GopherJS's representation into Go's representation.
```
nevkontakte, line 10: I think I like this idea in theory, but I wonder if it may break user code that references those symbols but doesn't use them in a way that's unsupported. For example, an unsafe conversion `[]int -> unsafe.Slice -> []int` would still work.

compiler/natives/src/reflect/reflect.go
grantnelson-wf: So, I ran into this. I was trying to follow what Flimzy had done for `[]T -> *[N]T` which makes a shared array and [the `TestConvertSlice2Array` test](https://cs.opensource.google/go/go/+/refs/tags/go1.20.14:src/reflect/all_test.go;l=4706-4707) caught that. The test says,
> Converting a slice to non-empty array needs to return a non-addressable copy of the original memory.

So upstream makes a copy? And by making a copy here the tests pass.

It is odd that `[]T -> *[N]T` would act differently than `[]T -> [N]T` but I think it's because `*[N]T` is a pointer to a potential `[N]T` where as `[N]T` is allocated memory for the variable. So you could point `*[N]T` at another footprint but `[N]T` is a new one. Or something like that?

compiler/natives/src/internal/unsafeheader/unsafeheader.go
grantnelson-wf: I felt it was safe to remove these since they are in an internal package and the comments on [the source code](https://cs.opensource.google/go/go/+/refs/tags/go1.20.14:src/internal/unsafeheader/unsafeheader.go) say,
> Slice is the runtime representation of a slice. It cannot be used safely or portably and its representation may change in a later release.

I actually based this off of [your commit for `TestWriteThroughHeader`](https://github.com/gopherjs/gopherjs/commit/4a391b4bbf145c9053c05016a25f65a492c92ff4) where you say,
> This test relies on assumptions about string and slice implementation that do not in fact hold true for GopherJS.

I figured that it would help me track down things that compile fine and may not be tested fully, but leverage those structures and therefore would break if used. By removing them it will keep them from compiling. It turns out, it was hardly used.

compiler/natives/src/reflect/reflect.go
nevkontakte: Yep, I think you are correct, thanks for confirming!

compiler/natives/src/internal/unsafeheader/unsafeheader.go
nevkontakte: Right, I missed this was an internal package. Then we don't risk breaking anyone besides ourselves, which we can deal with :) I think some public package defines similar structures, which threw me off.

nevkontakte (approved):

PR #1327, opened 2024-07-19
Make slow stdlib tests faster.

nevkontakte:
The main improvement here applies to crypto/internal/edwards25519/..., which use testing/quick for fuzz-style tests. Because those tests rely heavily on 64-bit math, they are much, much slower under GopherJS than upstream.

I added a hook into testing/quick that allows us to cap the number of iterations a test would be allowed to execute. Although that does technically reduce coverage, it's better than disabling the test entirely and we don't change any logic in those packages in the first place.

I deleted some overlays that corresponded to the old location of the same packages, the changes in this commit serve the same purpose.

I also shortened a couple of tests in image/gif and crypto/x509, for the same reasons.


grantnelson-wf:
FWIW, it looks great to me.

flimzy (approved):

PR #1332, opened 2024-07-23
[go1.20] Added copy from slice to array copy

grantnelson-wf:
The GopherJS transpiler will converts
```Go
slice := []byte{1, 2, 3, 4}
array := [4]byte(slice)
```
into
```JS
array = $clone(($clone(slice, arrayType)), arrayType);
```
the `$clone` calls the copy method on the array type. The copy method used `src.length` that is undefined for a slice. I added to the array type copy method to also allow for conversion from slices into arrays. The secondary `$clone` is from the assignment of an array to an array.

This will fix the issues in `net/netip` where they are converting slices of different lengths into arrays for ip addresses. This change should fix the last issue in GopherJS Tests so it should pass. Most of CI should pass, but `measurement_size.yml` will still fail in this PR (it's fixed by another PR).

This is part of #1270

### DIFF
--- a/compiler/prelude/types.js
+++ b/compiler/prelude/types.js
@@ -150,7 +150,16 @@ var $newType = (size, kind, string, named, pkg, exported, constructor) => {
                     }), "$");
                 };
                 typ.copy = (dst, src) => {
-                    $copyArray(dst, src, 0, 0, src.length, elem);
+                    if (src.length === undefined) {
+                        // copy from a slice, the slice may be bigger but not smaller than the array
+                        if (src.$length < dst.length) {
+                            $throwRuntimeError("cannot convert slice with length "+src.$length+" to array or pointer to array with length "+dst.length);
+                        }
+                        $copyArray(dst, src.$array, 0, 0, dst.length, elem);
+                    } else {
+                        // copy from another array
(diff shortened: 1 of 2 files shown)


flimzy (approved):

compiler/prelude/types.js
```diff
@@ -150,7 +150,16 @@ var $newType = (size, kind, string, named, pkg, exported, constructor) => {
                     }), "$");
                 };
                 typ.copy = (dst, src) => {
-                    $copyArray(dst, src, 0, 0, src.length, elem);
+                    if (src.length === undefined) {
+                        // copy from a slice, the slice may be bigger but not smaller than the array
+                        if (src.$length < dst.length) {
+                            $throwRuntimeError("cannot convert slice with length "+src.$length+" to array or pointer to array with length "+dst.length);
+                        }
+                        $copyArray(dst, src.$array, 0, 0, Math.min(src.$length, dst.length), elem);
```
grantnelson-wf, line 158: It just occurred to me that `Math.min(src.$length, dst.length)` could simplified to just `dst.length`, since we know `dst.lenght` is less than or equal to the `src.$length`. I'll push up that fix in a second.

grantnelson-wf: Fixed

PR #1337, opened 2024-07-31
[go1.20] Moving DCE to isolate filter naming

grantnelson-wf:
The dead-code elimination (DCE) code has two filter strings and dependencies as a set of strings. The strings have special meanings such as both empty strings indicate always alive. To help facilitate future changes to DCE, I'm moving the DCE code to it's own package. Also I'm making some methods to set a decl's DCE filters and deps to ensure that the names are continued to be set the same as new changes are made.

These changes should have only minor functional changes, everything should continue to function as before. These changes are mostly a minor refactor to encapsulate DCE design decisions. The CI should continue to pass.

This is related to #1270 work since the future changes to DCE is to include removal of dead generic code instantiations, e.g. allow `List[int]` to be alive and `List[string]` be dead separately.

### DIFF
--- a/compiler/internal/dce/info.go
+++ b/compiler/internal/dce/info.go
@@ -1,0 +1,108 @@
+package dce
+
+import (
+	"fmt"
+	"go/types"
+	"sort"
+	"strings"
+
+	"github.com/gopherjs/gopherjs/compiler/typesutil"
+)
+
+// Info contains information used by the dead-code elimination (DCE) logic to
+// determine whether a declaration is alive or dead.
+type Info struct {
+
+	// alive indicates if the declaration is marked as alive
+	// and will not be eliminated.
+	alive bool
+
+	// importPath is the package path of the package the declaration is in.
+	importPath string
+
+	// Symbol's identifier used by the dead-code elimination logic, not including
+	// package path. If empty, the symbol is assumed to be alive and will not be
+	// eliminated. For methods it is the same as its receiver type identifier.
+	objectFilter string
+
+	// The second part of the identified used by dead-code elimination for methods.
+	// Empty for other types of symbols.
+	methodFilter string
+
+	// List of fully qualified (including package path) DCE symbol identifiers the
+	// symbol depends on for dead code elimination purposes.
+	deps []string
+}
+
+// String gets a human-readable representation of the DCE info.
+func (d *Info) String() string {
+	tags := ``
+	if d.alive {
+		tags += `[alive] `
+	}
+	if d.unnamed() {
+		tags += `[unnamed] `
+	}
+	fullName := d.importPath + `.` + d.objectFilter
+	if len(d.methodFilter) > 0 {
+		fullName += `.` + d.methodFilter
+	}
+	return tags + fullName + ` -> [` + strings.Join(d.deps, `, `) + `]`
+}
+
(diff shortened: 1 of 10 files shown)


nevkontakte:
I think I'd like to take a step further and move all DCE-related logic into a separate `compiler/internal/dce` package.

grantnelson-wf:
Working on fixing some edge cases after the move to it's own package.

compiler/utils.go
```diff
@@ -103,12 +104,12 @@ func (fc *funcContext) Delayed(f func()) {
...
 // CollectDCEDeps captures a list of Go objects (types, functions, etc.)
-// the code translated inside f() depends on. The returned list of identifiers
-// can be used in dead-code elimination.
+// the code translated inside f() depends on. Then sets those objects
+// as dependencies of the given dead-code elimination info.
 //
 // Note that calling CollectDCEDeps() inside another CollectDCEDeps() call is
 // not allowed.
-func (fc *funcContext) CollectDCEDeps(f func()) []string {
+func (fc *funcContext) CollectDCEDeps(dce *dce.Info, f func()) {
```
nevkontakte, line 112: Consider this an optional request, but I would move this method into the dce package as well.

compiler/internal/dce/info.go
```diff
@@ -0,0 +1,75 @@
...
+	deps []string
+}
+
+// isAlive returns true if the declaration is marked as alive.
+func (d *Info) isAlive() bool {
+	return d.objectFilter == "" && d.methodFilter == ""
+}
+
+// SetAsAlive marks the declaration as alive, meaning it will not be eliminated.
+// This must be done after the SetName method is called.
```
nevkontakte, line 37: Let's make this method panic if this invariant is violated.

compiler/internal/dce/selector.go
nevkontakte, line 1: Since we are factoring this this logic out into its own package, it would be good to add at least some rudimentary tests

nevkontakte (changes requested):
I left a few suggestions, but generally I'm quite happy with this direction. Thanks!

compiler/utils.go
grantnelson-wf: Moved to `dce.Collector`

compiler/internal/dce/info.go
grantnelson-wf: Instead of a panic, I made the explicitly marking as alive use an `alive` flag instead of clearing out the names. Now, if the names aren't set or if the `alive` flag is set true, the decl is alive. This allows the decl to be marked as alive before or after naming without effecting the naming.

compiler/internal/dce/selector.go
grantnelson-wf: Added some tests

grantnelson-wf:
@nevkontakte and @flimzy as I've been working on the generics DCE, I've pushed a few updates to this PR, but I think it has reached a stable point where this PR (hopefully) is a good foundation to build the generics DCE off of. However, I haven't fully figured out the generics DCE yet; I just didn't want to get this out and immediately change it all for generics 😆. Anyway, I think I've address the prior change requests and this PR should be ready for review again.

grantnelson-wf:
Moving this over to master in https://github.com/gopherjs/gopherjs/pull/1339

PR #1339, opened 2024-08-22
Moving DCE into its own package

grantnelson-wf:
The dead-code elimination (DCE) code has two filter strings and dependencies as a set of strings. The strings have special meanings such as both empty strings indicate always alive. To help facilitate future changes to DCE, I'm moving the DCE code to it's own package. Also I'm making some methods to set a decl's DCE filters and deps to ensure that the names are continued to be set the same as new changes are made.

These changes should have only minor functional changes, everything should continue to function as before. These changes are mostly a minor refactor to encapsulate DCE design decisions. The CI should continue to pass.

This is related to #1013 and #1270 work since the future changes to DCE is to include removal of dead generic code instantiations, e.g. allow `List[int]` to be alive and `List[string]` be dead separately.

### DIFF
--- a/compiler/internal/dce/info.go
+++ b/compiler/internal/dce/info.go
@@ -5,8 +5,6 @@ import (
 	"go/types"
 	"sort"
 	"strings"
-
-	"github.com/gopherjs/gopherjs/compiler/typesutil"
 )
 
 // Info contains information used by the dead-code elimination (DCE) logic to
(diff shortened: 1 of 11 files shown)


compiler/internal/dce/info.go
```diff
@@ -0,0 +1,108 @@
...
+	importPath string
+
+	// Symbol's identifier used by the dead-code elimination logic, not including
+	// package path. If empty, the symbol is assumed to be alive and will not be
+	// eliminated. For methods it is the same as its receiver type identifier.
+	objectFilter string
+
+	// The second part of the identified used by dead-code elimination for methods.
+	// Empty for other types of symbols.
+	methodFilter string
```
nevkontakte, line 30: optional: I am happy to merge this RP as-is to limit the amount of changes in the logic, but it may be a worthwhile refactoring to consider.

It may be worthwhile grouping these three strings into a `dce.Key` type, and moving logic such as key derivation from a `types.Object` and key comparison into that type. I think that way we could use the same key representation everywhere in the code. Currently, for example, `dceInfo.objectFilter` is subtly different from `Info.objectFilter` because the former is qualified by the import path. I have a hunch that having `dce.Key` type may make things a little clearer instead of having to rely on untyped strings.

Anyway, this is just food for thought, I am okay with merging as is, if you prefer.

grantnelson-wf: So, I did this abstraction prior to really understanding the DCE very well. I was trying to make it just a simple abstraction with minimal changes. However, I think if I update towards what I have in the change to handle instances PR, #1340, it will be closer to what you were thinking, but still without a key object.

For example, I later realized it didn't make sense to keep `importPath` separate, so the names are back to just `objectFilter` and `methodFilter`. I also added a `filters.go` that does all the name generation stuff. Eventually the dependencies and names both use names returned from the same filter generation function. The filter generation does get somewhat complicated so it might be better to rethink creating a `Key` later when reviewing #1340.

nevkontakte (approved):

PR #1340, opened 2024-08-23
Adding instance tracking to the DCE

grantnelson-wf:
This extends dead-code elimination (DCE) to handle type arguments for instantiations of generic types. Since this change reworked how the DCE works, I added a README.md that discusses the idea and direction for these changes whilst giving help to any future extensions of the DCE.

The CI should continue to pass and I added a lot of new tests in the DCE package. We should probably at some point add some integration tests in the compiler package to ensure that the correct information is being given to the DCE, but because of how large that could be, I did not do that in this PR.


compiler/internal/dce/README.md
```diff
@@ -92,10 +93,14 @@ All the types in the function signatures and embedded interfaces are the
...
 Interfaces may contain exported and unexported function signatures.
-If an interface is alive then all of the functions, even the unexported
-functions, are alive.
+If an interface is alive then all of the functions are alive.
 Since there are many ways to wrap a type with an interface, any alive type that
-duck-types to an interface must have all of the matching methods alive.
+duck-types to an interface must have all of the matching methods also alive.
+
+In theory the unexported functions are also alive however, for GopherJS there
+is an exception because duck-typing is handled separately the method
```
flimzy, line 101: Missing word?

```suggestion
is an exception because duck-typing is handled separately from the method
```

flimzy (approved):

compiler/internal/dce/README.md
grantnelson-wf: Fixed

compiler/internal/dce/filters.go
```diff
@@ -2,31 +2,339 @@ package dce
...
+	if f, ok := o.(*types.Func); ok {
+		sig := f.Type().(*types.Signature)
+		if recv := sig.Recv(); recv != nil {
+			// The object is a method so the object filter is the receiver type
+			// if the receiver type is named, otherwise it's an unnamed interface.
+			typ := recv.Type()
+			if ptrType, ok := typ.(*types.Pointer); ok {
+				typ = ptrType.Elem()
+			}
+			if len(tArgs) <= 0 {
```
flimzy, line 29: nit/optional
```suggestion
			if len(tArgs) == 0 {
```

compiler/internal/dce/filters.go
```diff
@@ -2,31 +2,339 @@ package dce
...
 }
 
-// getDepFilter returns the filter for the given object to be used as a dependency.
-func getDepFilter(o types.Object) string {
-	qualifiedName := o.Pkg().Path() + "." + o.Name()
-	if typesutil.IsMethod(o) {
-		qualifiedName += "~"
+// getObjectFilter returns the object filter that functions as the primary
+// name when determining if a declaration is alive or not.
+// See [naming design](./README.md#naming) for more information.
```
flimzy, line 51: Markdown won't render properly here, and elsewhere.  Consider instead:

```suggestion
// See [naming design] for more information.
//
// [naming design]: https://github.com/gopherjs/gopherjs#naming
```

Or maybe even write the new documentation in GoDoc, rather than the README, for easier linking.

compiler/internal/dce/filters.go
```diff
@@ -2,31 +2,339 @@ package dce
...
+	for i := 0; i < inter.NumEmbeddeds(); i++ {
+		if union, ok := inter.EmbeddedType(i).(*types.Union); ok {
+			parts = append(parts, gen.Union(union))
+		}
+	}
+	// Sort the parts of the interface since the order doesn't matter.
+	// e.g. `interface { a(); b() }` is the same as `interface { b(); a() }`.
+	sort.Strings(parts)
+
+	if len(parts) <= 0 {
```
flimzy, line 301: nit/optional
```suggestion
	if len(parts) == 0 {
```

flimzy (approved):
Looks good to me. Just some stylistic suggestions, that I'm happy to ignore and merge. Just LMK.

compiler/internal/dce/filters.go
grantnelson-wf: 😆 That's muscle memory from working in C/C++ for so many years (same with me having to fix any `++i` before it compiles). I fixed all the cases like that.

compiler/internal/dce/filters.go
grantnelson-wf: Fixed

PR #1343, opened 2024-09-25
Adding DCE integration tests in compiler

grantnelson-wf:
This adds tests for the dead-code elimination (DCE) package into the compiler. The compiler performs many tasks that are related to DCE such as setting Decl's names and dependencies. These tests are designed like an integration test to ensure that the correct information is being given to the DCE and that the DCE is doing the correct things with that information.

This will introduce some TODO's that are resolved by the next part #1340. These TODO's exemplify the changes needed to be made to the DCE to properly handle generics and type instances. Double check that I'm not taking crazy pills with these TODO's and that they make sense to be the thing that needs to be fixed in the next part.

---

🍰 As an added bonus; I upgraded the existing test to stop using the deprecated `golang.org/x/tools/go/loader` and instead start using `golang.org/x/tools/go/packages`. It's nothing that exciting since it only is for source code for strings, but it's a start 😄

### DIFF
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -2,9 +2,7 @@ package compiler
 
 import (
 	"bytes"
-	"fmt"
 	"go/types"
-	"path/filepath"
 	"regexp"
 	"sort"
 	"testing"
@@ -13,38 +11,40 @@ import (
 	"golang.org/x/tools/go/packages"
 
 	"github.com/gopherjs/gopherjs/compiler/internal/dce"
+	"github.com/gopherjs/gopherjs/internal/srctesting"
 )
 
 func TestOrder(t *testing.T) {
 	fileA := `
-package foo
+		package foo
 
-var Avar = "a"
+		var Avar = "a"
 
-type Atype struct{}
+		type Atype struct{}
 
-func Afunc() int {
-	var varA = 1
-	var varB = 2
-	return varA+varB
-}
-`
+		func Afunc() int {
+			var varA = 1
+			var varB = 2
+			return varA+varB
+		}`
 
 	fileB := `
-package foo
+		package foo
 
-var Bvar = "b"
+		var Bvar = "b"
 
-type Btype struct{}
+		type Btype struct{}
 
-func Bfunc() int {
-	var varA = 1
-	var varB = 2
-	return varA+varB
-}
-`
+		func Bfunc() int {
+			var varA = 1
+			var varB = 2
+			return varA+varB
+		}`
 
-	files := []source{{"fileA.go", []byte(fileA)}, {"fileB.go", []byte(fileB)}}
+	files := []srctesting.Source{
+		{Name: "fileA.go", Contents: []byte(fileA)},
+		{Name: "fileB.go", Contents: []byte(fileB)},
+	}
 
 	compareOrder(t, files, false)
 	compareOrder(t, files, true)
@@ -64,7 +64,7 @@ func TestDeclSelection_KeepUnusedExportedMethods(t *testing.T) {
 			Foo{}.Bar()
 		}`
 
-	srcFiles := []source{{`main.go`, []byte(src)}}
+	srcFiles := []srctesting.Source{{Name: `main.go`, Contents: []byte(src)}}
 	sel := declSelection(t, srcFiles, nil)
 
 	sel.DeclCode.IsAlive(`^\s*Foo = \$newType`)
@@ -86,7 +86,7 @@ func TestDeclSelection_RemoveUnusedUnexportedMethods(t *testing.T) {
 			Foo{}.Bar()
 		}`
 
-	srcFiles := []source{{`main.go`, []byte(src)}}
+	srcFiles := []srctesting.Source{{Name: `main.go`, Contents: []byte(src)}}
 	sel := declSelection(t, srcFiles, nil)
 
 	sel.DeclCode.IsAlive(`^\s*Foo = \$newType`)
@@ -122,14 +122,14 @@ func TestDeclSelection_KeepUnusedUnexportedMethodForInterface(t *testing.T) {
 			}
 		}`
 
-	srcFiles := []source{{`main.go`, []byte(src)}}
+	srcFiles := []srctesting.Source{{Name: `main.go`, Contents: []byte(src)}}
 	sel := declSelection(t, srcFiles, nil)
 
 	sel.DeclCode.IsAlive(`^\s*Foo = \$newType`)
 	sel.DeclCode.IsAlive(`^\s*\$ptrType\(Foo\)\.prototype\.Bar`)
 
-	// `baz` is used to duck-type (via method list) against IFoo
-	// but the method itself is not used so can be removed.
+	// `baz` signature metadata is used to check a type assertion against IFoo,
+	// but the method itself is never called, so it can be removed.
 	sel.DeclCode.IsDead(`^\s*\$ptrType\(Foo\)\.prototype\.baz`)
 	sel.MethodListCode.IsAlive(`^\s*Foo.methods = .* \{prop: "baz", name: "baz"`)
 }
@@ -152,7 +152,7 @@ func TestDeclSelection_KeepUnexportedMethodUsedViaInterfaceLit(t *testing.T) {
 			f.baz()
 		}`
 
-	srcFiles := []source{{`main.go`, []byte(src)}}
+	srcFiles := []srctesting.Source{{Name: `main.go`, Contents: []byte(src)}}
 	sel := declSelection(t, srcFiles, nil)
 
 	sel.DeclCode.IsAlive(`^\s*Foo = \$newType`)
@@ -172,7 +172,7 @@ func TestDeclSelection_KeepAliveUnexportedMethodsUsedInMethodExpressions(t *test
 			fb(Foo{})
 		}`
 
-	srcFiles := []source{{`main.go`, []byte(src)}}
+	srcFiles := []srctesting.Source{{Name: `main.go`, Contents: []byte(src)}}
 	sel := declSelection(t, srcFiles, nil)
 
 	sel.DeclCode.IsAlive(`^\s*Foo = \$newType`)
@@ -196,7 +196,7 @@ func TestDeclSelection_RemoveUnusedFuncInstance(t *testing.T) {
 			println(Sum(1.1, 2.2, 3.3))
 		}`
 
-	srcFiles := []source{{`main.go`, []byte(src)}}
+	srcFiles := []srctesting.Source{{Name: `main.go`, Contents: []byte(src)}}
 	sel := declSelection(t, srcFiles, nil)
 
 	sel.DeclCode.IsAlive(`^\s*Sum\[\d+ /\* float64 \*/\]`)
@@ -205,7 +205,7 @@ func TestDeclSelection_RemoveUnusedFuncInstance(t *testing.T) {
 	sel.DeclCode.IsDead(`^\s*Foo = function`)
 	sel.DeclCode.IsDead(`^\s*sliceType(\$\d+)? = \$sliceType\(\$Int\)`)
 
-	// TODO(gn): This should not be alive because it is not used.
+	// TODO(grantnelson-wf): This should not be alive because it is not used.
 	sel.DeclCode.IsAlive(`^\s*Sum\[\d+ /\* int \*/\]`)
 }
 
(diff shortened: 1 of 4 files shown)


compiler/compiler_test.go
```diff
@@ -73,44 +384,116 @@ type source struct {
...
+	pkgs, err := packages.Load(config, patterns...)
+	if err != nil {
+		t.Fatal(`error loading packages:`, err)
+	}
+
+	hasErrors := false
+	packages.Visit(pkgs, nil, func(pkg *packages.Package) {
+		for _, err := range pkg.Errors {
+			hasErrors = true
+			fmt.Println(err)
```
nevkontakte, line 457: t.Log() or t.Errorf() is more appropriate here.

compiler/compiler_test.go
```diff
@@ -119,28 +502,146 @@ func compile(path string, sourceFiles []source, minify bool) ([]byte, error) {
...
+	root := parseSources(t, sourceFiles, auxFiles)
+	archives := compileProject(t, root, false)
+	mainPkg := archives[root.PkgPath]
+
+	paths := make([]string, 0, len(archives))
+	for path := range archives {
+		paths = append(paths, path)
+	}
+	sort.Strings(paths)
+	packages := make([]*Archive, 0, len(archives)-1)
```
nevkontakte, line 551: Why is the capacity `len(archives)-1`. Considering the array is filled by iterating over `pats`, which has size `len(archives)`, it seems like an odd choice.

compiler/compiler_test.go
```diff
@@ -119,28 +502,146 @@ func compile(path string, sourceFiles []source, minify bool) ([]byte, error) {
...
+		t:            t,
+		mainPkg:      mainPkg,
+		archives:     archives,
+		packages:     packages,
+		dceSelection: dceSelection,
+	}
+
+	st.DeclCode = &selectionCodeTester{st, `DeclCode`, func(d *Decl) []byte { return d.DeclCode }}
+	st.InitCode = &selectionCodeTester{st, `InitCode`, func(d *Decl) []byte { return d.InitCode }}
+	st.MethodListCode = &selectionCodeTester{st, `MethodListCode`, func(d *Decl) []byte { return d.MethodListCode }}
```
nevkontakte, line 574: Why assign separately instead of using the literal above?

compiler/compiler_test.go
```diff
@@ -119,28 +502,146 @@ func compile(path string, sourceFiles []source, minify bool) ([]byte, error) {
...
+	return st
+}
+
+func (st *selectionTester) PrintDeclStatus() {
+	st.t.Helper()
+	for _, pkg := range st.packages {
+		fmt.Println(`Package`, pkg.ImportPath)
```
nevkontakte, line 581: Same in this method, probably should use `t.Log()`.

compiler/compiler_test.go
nevkontakte, line 574: nit: with the exception of trivial single-field structs, I prefer having fields explicitly named in the literal. This makes it harder to mix up consecutive fields of the same type if they get added at a later time.

compiler/compiler_test.go
```diff
@@ -41,27 +43,336 @@ func Bfunc() int {
...
+			println("baz")
+		}
+		func main() {
+			Foo{}.Bar()
+		}`
+
+	srcFiles := []source{{`main.go`, []byte(src)}}
+	sel := declSelection(t, srcFiles, nil)
+
+	sel.DeclCode.IsAlive(`^\s*Foo = \$newType`)
```
nevkontakte, line 70: Depending on the generated code is somewhat fragile. Maybe we could use `Decl.FullName` instead, at least for decls that support it?

compiler/compiler_test.go
```diff
@@ -41,27 +43,336 @@ func Bfunc() int {
...
+	sel.DeclCode.IsAlive(`^\s*\$ptrType\(Foo\)\.prototype\.Bar`)
+
+	// `baz` is used to duck-type (via method list) against IFoo
+	// but the method itself is not used so can be removed.
```
nevkontakte, line 132: Clarified phasing a little bit.

```suggestion
	// `baz` signature metadata is used to check a type assertion against IFoo,
	// but the method itself is never called, so it can be removed.
```

compiler/compiler_test.go
```diff
@@ -41,27 +43,336 @@ func Bfunc() int {
...
+	srcFiles := []source{{`main.go`, []byte(src)}}
+	sel := declSelection(t, srcFiles, nil)
+
+	sel.DeclCode.IsAlive(`^\s*Sum\[\d+ /\* float64 \*/\]`)
+	sel.DeclCode.IsAlive(`^\s*sliceType(\$\d+)? = \$sliceType\(\$Float64\)`)
+
+	sel.DeclCode.IsDead(`^\s*Foo = function`)
+	sel.DeclCode.IsDead(`^\s*sliceType(\$\d+)? = \$sliceType\(\$Int\)`)
+
+	// TODO(gn): This should not be alive because it is not used.
```
nevkontakte, line 208: I recommend using a full GitHub username in TODOs to make the author unambiguous. This is more important for longer-lived TODOs, where it may point at who to ask for context/clarifications if need be. But for the sake of consistency, I would do it here too.

compiler/compiler_test.go
```diff
@@ -73,44 +384,116 @@ type source struct {
 	contents []byte
 }
 
-func compile(path string, sourceFiles []source, minify bool) ([]byte, error) {
-	conf := loader.Config{}
-	conf.Fset = token.NewFileSet()
-	conf.ParserMode = parser.ParseComments
+func compile(t *testing.T, sourceFiles []source, minify bool) []byte {
```
nevkontakte, line 387: Some of these helper methods seem to be general enough to be useful for testing other packages. Maybe move them into `internal/srctesting`?

nevkontakte (changes requested):
Thanks, this is excellent work! Some test cases revealed DCE behavior that was surprising, even though it made sense in the end.

compiler/compiler_test.go
grantnelson-wf: Oops, that is odd. That was left over from something I tried. I had the code more like `ImportDependencies` where the root package is added to the end of the slice, then in this code was taking the root package off the end. Later I removed that complexity since, although more similar to the real thing, it wasn't needed for the tests.

I thought I had cleaned it all up, oh well, thanks for catching that.

compiler/compiler_test.go
grantnelson-wf: I just passed `st` into the `selectionCodeTester` so that I didn't have to pass in several of `selectionTester`'s fields individually, but that meant if it was in the literal I'd get a `st` not defined. It looks like all that `selectionCodeTester` uses is `t *testing.T`, `packages []*Archive`, and `dceSelection map[*Decl]struct{}` so it wouldn't actually be too much to send those in separately, so I'll just do that instead.

compiler/compiler_test.go
grantnelson-wf: Many of the `Decl.FullNames` are blank or the generic name duplicated on each instance. I totally agree, it seems fragile to use the generated code but it's the only thing that I could used to differentiate the decls sometimes. 🤷‍♂️ 

I tried to make it less fragile by using regular expressions so that code like `Bar[42 /* int */]` skips checking the number and only checks the "Bar" and "int" part, i.e. `Bar\[\d+ /\* int \*/\]`. It also made it easy to jump over those position marks (`\b`) in the code. But I realize that's not perfect. I thought about making methods like `IsAliveInstance("Bar", "int")` so that if it did change we'd have less to update, but it obscured what was being checked.

I created `PrintDeclStatus()` as one of the first things to see what Decls were in the package and what information they had. (Running debug wasn't very helpful to me since those were `[]byte`s) It prints the `Decl.FullNames` on the same line as `[Alive]` or `[Dead]`. The following is the output from adding `PrintDeclStatus` to `TestDeclSelection_RemoveUnusedStructTypeInstances`:

```Plain
Package command-line-arguments
  [Alive] ""
     DeclCode: "\tFoo = {};\n"
     TypeInitCode: "\t$pkg.Foo = Foo;\n"
     Vars: [Foo]
  [Alive] ""
     DeclCode: "\tFoo[0 /* float64 */] = $newType(0, $kindStruct, \"main.Foo[float64]\", true, \"command-line-arguments\", true, function(v_) {\n\t\tthis.$val = this;\n\t\tif (arguments.length === 0) {\n\t\t\tthis.v = 0;\n\t\t\treturn;\n\t\t}\n\t\tthis.v = v_;\n\t});\n"
     MethodListCode: "\tFoo[0 /* float64 */].methods = [{prop: \"Bar\", name: \"Bar\", pkg: \"\", typ: $funcType([], [], false)}];\n"
     TypeInitCode: "\tFoo[0 /* float64 */].init(\"command-line-arguments\", [{prop: \"v\", name: \"v\", embedded: false, exported: false, typ: $Float64, tag: \"\"}]);\n"
  [Alive] ""
     DeclCode: "\tFoo[2 /* int */] = $newType(0, $kindStruct, \"main.Foo[int]\", true, \"command-line-arguments\", true, function(v_) {\n\t\tthis.$val = this;\n\t\tif (arguments.length === 0) {\n\t\t\tthis.v = 0;\n\t\t\treturn;\n\t\t}\n\t\tthis.v = v_;\n\t});\n"
     MethodListCode: "\tFoo[2 /* int */].methods = [{prop: \"Bar\", name: \"Bar\", pkg: \"\", typ: $funcType([], [], false)}];\n"
     TypeInitCode: "\tFoo[2 /* int */].init(\"command-line-arguments\", [{prop: \"v\", name: \"v\", embedded: false, exported: false, typ: $Int, tag: \"\"}]);\n"
  [Dead]  ""
     InitCode: "\b\x00\x00\x00i\t\t$unused(new Foo[0 /* float64 */].ptr(3.14));\n"
     Vars: [_]
  [Alive] "(command-line-arguments.Foo[T]).Bar"
     DeclCode: "\t$ptrType(Foo[0 /* float64 */]).prototype.Bar = function Foo·Bar() {\n\t\tvar f;\n\t\tf = this;\n\b\x00\x00\x00O\t\tconsole.log(f.v);\n\b\x00\x00\x00\x00\t};\n\tFoo[0 /* float64 */].prototype.Bar = function(...$args) { return this.$val.Bar(...$args); };\n"
  [Alive] "(command-line-arguments.Foo[T]).Bar"
     DeclCode: "\t$ptrType(Foo[2 /* int */]).prototype.Bar = function Foo·Bar$1() {\n\t\tvar f;\n\t\tf = this;\n\b\x00\x00\x00O\t\tconsole.log(f.v);\n\b\x00\x00\x00\x00\t};\n\tFoo[2 /* int */].prototype.Bar = function(...$args) { return this.$val.Bar(...$args); };\n"
  [Alive] ""
     Vars: [main]
  [Alive] "command-line-arguments.main"
     DeclCode: "\tmain = function main$1() {\n\b\x00\x00\x00\xa1\t\t$clone(new Foo[2 /* int */].ptr(7), Foo[2 /* int */]).Bar();\n\b\x00\x00\x00\x00\t};\n"
  [Alive] ""
     InitCode: "\b\x00\x00\x00\x00\t\tif ($pkg === $mainPkg) {\n\b\x00\x00\x00\x95\t\t\tmain();\n\b\x00\x00\x00\x00\t\t\t$mainFinished = true;\n\b\x00\x00\x00\x00\t\t}\n"
```

compiler/compiler_test.go
grantnelson-wf: I won't be able to move any of them other than `ParseSources` without making `srctesting` depend on `compiler` because of the `Archive`'s. We could move more by using `package *_test` tests but for now I'll just `ParseSources`, if that's okay.

compiler/compiler_test.go
nevkontakte: I see, so a principled fix would be to go and generate sensible identifiers for each kind of decl we emit, and use them in the test. I realize that could be tricky in the general case because it reduces to the problem of unique type names we discussed in the chat.

We could consider generating identifiers that are "good enough" to distinguish decls in our test cases, but I'll let you decide if you want to do that or not.

compiler/compiler_test.go
nevkontakte: Ah, makes sense, I missed the reference to `st`. Either way would have been fine then.

compiler/compiler_test.go
nevkontakte: That's a good point. Let's keep it as is then.

nevkontakte (approved):
Overall I'm happy with this PR. If you don't think that generating stable identifiers for Decls is worth it, I'll merge it as-is.

compiler/compiler_test.go
grantnelson-wf: @nevkontakte, since I'm not sure what it would take to name those (like you said about the sensible identifiers) and since this is a dependency for #1340, I'd prefer to get this in sooner and do a naming decls / stability upgrade as a later effort.

_(And, since I have managers that track the Jira tickets I make for these PRs, I know they'd like to see this progressing, even if it means I open a bunch of follow up Jira tickets. Besides they understand we're on your schedule not ours, and I care more about helping out with good code then about what they want. 😁 Hi, Dave, if you're seeing this 😆)_

nevkontakte: No worries, like I said, this was an optional suggestion.

PR #1345, opened 2024-10-11
Adding more functions to InstanceMap

grantnelson-wf:
To finish updating blocking for instances of generic functions, I needed an instance map that also has a way to iterate key/value pairs (or a way to get list of keys) and to delete keys. I updated the existing `InstanceMap` by pulling some more ideas from the [typeutil.Map](https://pkg.go.dev/golang.org/x/tools/go/types/typeutil#Map).

### DIFF
--- a/compiler/internal/typeparams/map.go
+++ b/compiler/internal/typeparams/map.go
@@ -1,8 +1,10 @@
 package typeparams
 
 import (
+	"fmt"
 	"go/types"
-	"sync"
+	"sort"
+	"strings"
 
 	"golang.org/x/tools/go/types/typeutil"
 )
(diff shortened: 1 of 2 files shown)


compiler/internal/typeparams/map.go
```diff
@@ -25,40 +27,35 @@ type (
 // instance equality, objects are compared by pointer equality, and type
 // arguments with types.Identical(). To reduce access complexity, we bucket
 // entries by a combined hash of type args. This type is generally inspired by
-// typeutil.Map.
+// golang.org/x/tools/go/types/typeutil/map.go
```
flimzy, line 30: Nit/optional: Could be made a GoDoc link (assuming linking to this specific type is appropriate):

```suggestion
// [golang.org/x/tools/go/types/typeutil.Map]
```

flimzy (approved):

compiler/internal/typeparams/map.go
grantnelson-wf: I forgot that's how GoDoc links are formatted, too much markdown 😄 I'll get that fixed up to a valid link and put it in GoDoc's format.

flimzy (approved):

PR #1347, opened 2024-10-22
Blocking for Function Instances

grantnelson-wf:
- I moved the analysis package into an internal package

- I updated the analysis info to determine blocking for instances of functions taking into account the type arguments.

- I added a bunch of tests to analysis info for many of the different blocking that is being checked. I left some TODO's for things that are known not to function correctly and a TODO for a test that exemplifies an issue with sharing generic information across packages. Most of these TODO's will be fixed as part of the ongoing generics effort. The issues marked with TODO's should not cause any issues with consumers since they require generics to be an issue and generics hasn't been released yet.


compiler/internal/typeparams/instance.go
```diff
@@ -147,6 +147,18 @@ func (iset *InstanceSet) ByObj() map[types.Object][]Instance {
 	return result
 }
 
+// ForObj returns instances for a given object type belong to. Order is not specified.
+// This returns the same values as `ByObj()[obj]`.
```
flimzy, line 151: I'm no sure what this is meant to say. Not sure if it's missing a word, or has too many words.  :)  Please just re-read and edit for clarity.

compiler/internal/analysis/info.go
```diff
@@ -211,18 +251,30 @@ type FuncInfo struct {
...
 
 	pkgInfo      *Info // Function's parent package.
 	visitorStack astPath
 }
 
+// HasBlocking indicates if this function may block goroutine execution.
+//
+// For example, a channel operation in a function or a call to another
+// possibly blocking function may block the function.
+func (fi *FuncInfo) HasBlocking() bool {
```
flimzy, line 274: Elsewhere we use `IsBlocking`. Should this be the same for consistency, or is there a meaningful difference in the naming convention?

grantnelson-wf: That's a good question. I was replacing the repeat code `len(fi.Blocking) != 0` in a bunch of places so I thought, "this is checking if the function info _has blocking_ nodes", but now I'm wondering if that means "the function _is blocking_ because it has blocking nodes."

For consistency, I'll change it to `IsBlocking`

compiler/internal/analysis/info.go
```diff
@@ -334,13 +396,19 @@ func (fi *FuncInfo) Visit(node ast.Node) ast.Visitor {
...
-		fi.callToNamedFunc(fi.pkgInfo.Uses[f])
+		fi.callToNamedFunc(fi.instanceForIdent(f))
 	case *ast.SelectorExpr:
-		if sel := fi.pkgInfo.Selections[f]; sel != nil && typesutil.IsJsObject(sel.Recv()) {
-			// js.Object methods are known to be non-blocking, but we still must
-			// check its arguments.
+		if sel := fi.pkgInfo.Selections[f]; sel != nil {
+			if typesutil.IsJsObject(sel.Recv()) {
+				// js.Object methods are known to be non-blocking, but we still must
+				// check its arguments.
```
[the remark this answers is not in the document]
grantnelson-wf: Fixed

flimzy (approved):

flimzy:
Thanks!

PR #1349, opened 2024-11-01
[WIP] Improving return with defers blocking

grantnelson-wf:


compiler/internal/analysis/info.go
```diff
@@ -248,16 +228,27 @@ type FuncInfo struct {
 	// List of continue statements in the function.
 	continueStmts []continueStmt
 	// List of return statements in the function.
-	returnStmts []astPath
+	returnStmts []returnStmt
+	// List of most of the deferred function calls in the function.
```
flimzy, line 232: Why "most"?  Which ones are excluded and why?

flimzy (approved):
I don't see anything that jumps out at me, but it's hard to validate this sort of code visually. I'm happy with merging it, but maybe @nevkontakte wants to take a look too.

compiler/internal/analysis/info.go
grantnelson-wf: We should be able record every `defer` but I didn't because the `defer`s that aren't blocking aren't needed when determining if a `return` is blocking. So, since calls from `js.Object`s and builtin functions (e.g. `len`) aren't blocking, I didn't add those. Instead I added some comments on [line 472](https://github.com/gopherjs/gopherjs/pull/1349/files#diff-8ca33eb0bc7ed01cd267a73c3e2557daa5b9bb71b5ebb20b81fb1fb7b691e668R472) and [line 623](https://github.com/gopherjs/gopherjs/pull/1349/files#diff-8ca33eb0bc7ed01cd267a73c3e2557daa5b9bb71b5ebb20b81fb1fb7b691e668R623) in info.go where we possibly are skipping defers. We could add them with `if forDefer { fi.deferStmts = append(fi.deferStmts, newDefer(false)) }` in those cases.

Since we still needed to know if there were any defers, I left the `HasDefer` as it was.

nevkontakte: Maybe rephrase it as "List of deferred function calls which could be blocking"?

compiler/internal/analysis/info.go
```diff
@@ -393,25 +459,27 @@ func (fi *FuncInfo) Visit(node ast.Node) ast.Visitor {
 	// needs to be analyzed.
 }
 
-func (fi *FuncInfo) visitCallExpr(n *ast.CallExpr) ast.Visitor {
+func (fi *FuncInfo) visitCallExpr(n *ast.CallExpr, forDefer bool) ast.Visitor {
```
nevkontakte, line 462: nit: it's a personal preference, but I think `deferredCall bool` would be a bit more descriptive.

compiler/internal/analysis/defer.go
```diff
@@ -0,0 +1,108 @@
...
+//
+// The result will be cached for future calls since each return statement
+// will check the same defers (plus any new defers) of prior return statements.
+func (d *deferStmt) IsBlocking(info *Info) bool {
+	blocking, cacheResult := d.isBlocking(info)
+	if cacheResult {
+		d.isBlocking = func(info *Info) (bool, bool) {
+			return blocking, false
+		}
+	}
```
nevkontakte, line 97: Does caching actually provide a significant performance boost?

If not, I would avoid it, since it could be a cause for a difficult to debug inconsistency if at some future point functions could be declared blocked at a later processing stage.

nevkontakte (approved):

compiler/internal/analysis/defer.go
grantnelson-wf: Good point, I'll make those more debuggable. The caching was only because I knew that each return would check all the defers the prior return checked plus some new ones (in general) and I was trying to reduce duplicate checks, but since those are just doing map lookups, it's probably not that much of a performance boost.

compiler/internal/analysis/info.go
grantnelson-wf: Changed

nevkontakte (approved):

nevkontakte:
LGTM, thank you!

PR #1350, opened 2024-11-19
Updating Decl's FullNames

grantnelson-wf:
The blocking analysis was looking up the generic name for a declaration and may have been finding the function variable declaration (used for storing all instances of a generic function). To fix this we need the declarations to have better full names. 

Changes:

- I added some methods used to generate names that could be used to lookup decls in most cases. In some cases it would be difficult to get a unique name, e.g. `init` functions, when performing a lookup in an `Archive`. Fortunately, those cases have no need to be looked up (yet or ever as far as I can figure).

- Updated the unit-tests that were using regular expression matching of the JS code to instead use these new names.

- Added `t.Helper()` calls into `info_test.go` assert methods, as asked for in https://github.com/gopherjs/gopherjs/pull/1351#discussion_r1875057369


compiler/declNames.go
```diff
@@ -0,0 +1,70 @@
...
+
+// mainFuncFullName returns the name for the declaration used to invoke the
+// main function of the program. There should only be one decl with this name.
+func mainFuncDeclFullName() string {
+	return `init:main`
+}
+
+// funcDeclFullName returns a name for a package-level function
+// declaration for the given instance of a function.
+// The name is unique except unless the function is an `init` function.
```
flimzy, line 46:
```suggestion
// The name is unique unless the function is an `init` function.
```

compiler/declNames.go
```diff
@@ -0,0 +1,70 @@
...
+func typeDeclFullName(inst typeparams.Instance) string {
+	return `type:` + inst.String()
+}
+
+// anonTypeDeclFullName returns a unique name for a package-level type
+// declaration for an anonymous type. Names are only unique per package.
+// These names are generated for types that are not named in the source code.
+func anonTypeDeclFullName(o types.Object) string {
+	return `anonType:` + symbol.New(o).String()
+}
```
flimzy, line 70: I don't see package level constants mentioned here. Are they a special case that doesn't apply to our use case?

nevkontakte: IIRC all constants are inlined, so they don't get decls generated.

nevkontakte (approved):
May be worth adding a sanity check somewhere that no archive has two decls with the same full name.

compiler/declNames.go
grantnelson-wf: Thanks, I had a moment of confusion 😕. I was going to add a test to see what happened when I got a chance, because I was quite confident that I got every Decl named and I hadn't remembered a constant. I had guessed they were inlined or added as a variable that never gets assigned. Makes sense if they are inlined.

compiler/declNames.go
grantnelson-wf: Finally got back around to fixing this grammar.

compiler/declNames.go
nevkontakte: For the inquisitive minds, this is the code that makes inlining happen: https://github.com/gopherjs/gopherjs/blob/428efba961576ae2f0aca3ebc48b478881dbf409/compiler/expressions.go#L38-L78

We rely on go/types computing static values for expressions, and where such value is available, we just emit it. This is also how `x := 1 + 3` turns into `x = 4` in the generated code.

PR #1351, opened 2024-11-21
Fixing issues in analysis around generics

grantnelson-wf:
While I was doing other work I ran into a few issues and fixed them:

1. **Unused Generic Functions**: Unused generic functions and methods that perform calls were making requests with the type parameters, e.g. `func Foo[T any]() { Bar[T]() }` will make a request for the function information for `Bar[T]` with `T` being the type parameter. This was because if there were no instances, I assumed that meant that it was a non-generic function, but I totally missed the possibility that it is a generic function that just isn't used. I now check for instances and check if the function is generic and if it is unused and generic, skip over the function and don't bother making a FuncInfo for it. Since it is unused, it can't be needed by anything else other than other unused functions.

2. **Function Literal Instances**: Function literals that are inside a generic function were not getting the instance information, e.g. `func Foo[T any]() { func() { Bar[T]() }() }` will make a request for the function information for `Bar[T]` with `T` being the type parameter. This was because when the function literal FuncInfo is created it wasn't getting the type arguments from the instance of the generic function that it is inside of. I changed the function literal to take on the instance it is inside of. To make this work I also made the function literal FuncInfo lookup require the type arguments of the specific instance of the function literal to lookup.

3. **Delay Remote Calls**: I make the `IsBlocking` method able to use the `isImportedBlocking` function if the instance of the function that is being looked up isn't from the current `Info`. With this change I could also delay the lookup of other remote calls and defers until the end of the analysis when function information is propagating. This change will help later for getting generic changes across package boundaries.

These changes will help prepare for some of the changes needed to make function lookup in the ImportContext look for the correct instance of a generic function Decl. This is related to #1013 and #1270

### DIFF
--- a/compiler/internal/analysis/info_test.go
+++ b/compiler/internal/analysis/info_test.go
@@ -156,10 +156,10 @@ func TestBlocking_GoRoutines_WithFuncLiterals(t *testing.T) {
 			}(<-c)
 		}`)
 	bt.assertNotBlocking(`notBlocking`)
-	bt.assertBlockingLit(4)
+	bt.assertBlockingLit(4, ``)
 
 	bt.assertBlocking(`blocking`)
-	bt.assertNotBlockingLit(10)
+	bt.assertNotBlockingLit(10, ``)
 }
 
 func TestBlocking_GoRoutines_WithNamedFuncs(t *testing.T) {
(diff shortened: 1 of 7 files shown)


flimzy (approved):
LGTM, but @nevkontakte may have additional comments.

compiler/internal/analysis/info_test.go
```diff
@@ -1529,21 +1694,29 @@ func (bt *blockingTest) assertFuncInstCount(expCount int) {
 	if got := bt.pkgInfo.funcInstInfos.Len(); got != expCount {
```
nevkontakte, line 1694: Here and other assert* functions should call t.Helper(), so that the test error shows the line in the surrounding test case and not the helper itself.

nevkontakte (approved):
Thank you! 🙏

compiler/internal/analysis/info_test.go
grantnelson-wf: Good point, I'll add that in as part of #1350

grantnelson-wf: Added https://github.com/gopherjs/gopherjs/pull/1350/commits/1fa608f5b1f374773e22d94c5d61d349b78137bd

PR #1354, opened 2024-12-05
Archive serialization updates and DCE fix

grantnelson-wf:
I made some serialization updates for archives to fix an issue I accidentally introduced with my DCE update (sorry 😥) . The DCE data wasn't being serialized because of how I encapsulated it and kept the fields un-exported. This meant an archive that was reloaded from memory would have no DCE info and think every Decl in it are always alive. This was discovered as I was exploring what kind of information we'd need to add to an archive for cross package generic resolution.

- I fixed the serialization of DCE whilst keeping the encapsulation.
- I fixed the "TODO"s in archives around getting access to that actual `types.Package` and `token.FileSet` instead of having byte arrays.
- I added some tests to the compiler that checks the JS after reloading from memory results in the same JS as without reloading, including DCE.
- I changed how build time and source mod time is passed into cache. This is partially done because reloading the `types.Package` from the byte array is now done earlier and we needed to still skip that if the archive is out-of-date.


flimzy (approved):

compiler/compiler.go
```diff
@@ -277,19 +255,98 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, gls goLinknameS
...
+	return &a, sa.BuildTime, nil
 }
 
 // WriteArchive writes compiled package archive on disk for later reuse.
-func WriteArchive(a *Archive, w io.Writer) error {
-	return gob.NewEncoder(w).Encode(a)
+//
+// The passed in buildTime is used to determine if the archive is out-of-date.
+// It should be set to time.Now() typically but it exposed for testing purposes.
+func WriteArchive(a *Archive, buildTime time.Time, w io.Writer) error {
```
nevkontakte, line 321: Why do we need to pass builtTime  here? In terms of cache invalidation, it doesn't matter when the cache was generated, it only matters when its inputs were most recently modified. So it seems redundant with the SrcModTime.

compiler/compiler.go
```diff
@@ -277,19 +255,98 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, gls goLinknameS
...
+func ReadArchive(importPath string, r io.Reader, srcModTime time.Time, imports map[string]*types.Package) (*Archive, time.Time, error) {
+	var sa serializableArchive
+	if err := gob.NewDecoder(r).Decode(&sa); err != nil {
+		return nil, time.Time{}, err
+	}
+
+	if srcModTime.After(sa.BuildTime) {
+		// Archive is out-of-date.
+		return nil, sa.BuildTime, nil
+	}
```
nevkontakte, line 288: From the separation of concerns perspective, I would expect ReadArchive() to always return the archive content, regardless of its age, and then its validity for a particular purpose to be decided separately.

compiler/compiler.go
```diff
@@ -277,19 +255,98 @@ func WritePkgCode(pkg *Archive, dceSelection map[*Decl]struct{}, gls goLinknameS
...
 
-	return &a, nil
+	a.ImportPath = sa.ImportPath
+	a.Name = sa.Name
+	a.Imports = sa.Imports
+	a.Declarations = sa.Declarations
+	a.IncJSCode = sa.IncJSCode
+	a.Minified = sa.Minified
+	a.GoLinknames = sa.GoLinknames
+	return &a, sa.BuildTime, nil
```
nevkontakte, line 314: If we think some sort of timestamp is relevant to the consumers of the archive, we should make it a part of the struct, instead of returning it separately. See also the comment below that it should be a derivative of SrcModTime.

compiler/compiler_test.go
```diff
@@ -405,6 +406,33 @@ func TestLengthParenthesizingIssue841(t *testing.T) {
...
+			println("do nothing")
+		}`
+	srcFiles := []srctesting.Source{{Name: `main.go`, Contents: []byte(src)}}
+	root := srctesting.ParseSources(t, srcFiles, nil)
+	rootPath := root.PkgPath
+	origArchives := compileProject(t, root, false)
+	readArchives := reloadCompiledProject(t, origArchives, rootPath)
+
+	origJS := renderPackage(t, origArchives[rootPath], false)
+	readJS := renderPackage(t, readArchives[rootPath], false)
```
nevkontakte, line 429: Can we just diff the archive slices? I'm reasonably certain cmp.Diff() can handle unexported fields if you pass it the right options.

compiler/compiler.go
grantnelson-wf: The reason was to try to keep the same functionality. Before the change we didn't deserialize the package and register the type until after checking that the Archive was up-to-date. Since [`gcexportdata.Read`](https://pkg.go.dev/golang.org/x/tools/go/gcexportdata#Read) says "Read may inspect and add to the imports map" I was worried what would happen if we deserialized a package that was out-of-date, then tried to overwrite it later with an up-to-date version.

I also wasn't sure of what changes, if any, it'd make to the import packages and types. If the only changes are to the import map, then we could make a copy of that map before the Read changes it, then only if the package is not old, merge the imports map while registering the type. Since I wasn't sure, I didn't want to risk it.

Anyway, I'm cool with whatever, including undoing those changes and leaving the package as serialized bytes. Since the package was already serialized, everywhere that needed the package just looks it up in the imports map using the path (as far as I could tell), so it's not like it is required to have that deserialized right now.

compiler/compiler.go
grantnelson-wf: My thought was, if we return a nil archive and nil error, there is no explanation for why it returned a nil archive, but if we needed to we could see the buildTime and determine that the archive was just out-of-date. Right now, that's just used for some logging right after load and then isn't used anymore.

I was trying to separate what is needed for reading and writing an Archive and what the Archive used for once loaded. I guess I figured that the Archive didn't need to know it's build time and that's just part of ensuring we don't load an out-of-date archive. Kinda like the build time is a little more resilient file mod time that can't be changed by a move or touch.

Seems like I might be thinking about it wrong. I don't mind undoing and putting the BuildTime field back on Archive. Since you definitely understand this better than I do, I'll do what you think makes the most sense.

compiler/compiler.go
grantnelson-wf: This buildTime is passed in because I was trying to separate the data needed to do a write/read from the Archive, so I changed the Archive to no longer has the BuildTime field on it. The BuildTime field was only used for some logging right after loading and also determining if the load only loaded an out-of-date Archive that needs to be thrown out.

I made it so that we can pass in the time that we want to use as "how old" the Archive is. Instead of passing it in, we could hardcode it internally to use `time.Now()` but it makes me nervous to write tests that depend on times and also use `time.Now()`.

So, the idea I was going for was; We pass in a time on write that is written as "when" it was written. Then we pass in a time on read that is used to as the "oldest" allowed time that will cancel the read if the file was written too long ago.

I was thinking of it like a MakeFile, were the file is rebuilt if old relative to something else, it doesn't matter what the file is, it's just old so rebuilt it. Except here we aren't using the mod time, so we need the build time to be passed in during the write.

(I'm not saying my thinking was right or wrong, I'm just letting you know what was in my mind so that you can determine what/if I'm thinking about it wrong.)

compiler/compiler_test.go
grantnelson-wf: Sure, I hadn't tried that and just used JS because I was basing this test off of the `TestOrder` test.

I also wanted to make sure decl DCE selection was working correctly since I was also fixing that. As long as `cmp.Diff` also checks unexpected data then it should be fine.

compiler/compiler.go
nevkontakte: Yeah, I think I get where you are coming from, and I agree in general. I was just thinking if we derive archive's validity timestamp from the SrcModTime, we make the logic even less time dependent, as in it wouldn't depend on when the build happened to run, only on when inputs got modified.

I'll leave it up to you if you think this direction is worth pursuing.

compiler/compiler.go
nevkontakte: Ah, right, I did not take the possible side-effects into account. Yeah, in this case such a behavior is justified. Maybe note it down somewhere in a comment for future archaeologists? :)

compiler/compiler.go
nevkontakte: Right, my comment was in conjunction with the one above, and your point about side effects in the imports map is a valid one. Ideally, each archive would decode gcexportdata into its own imports map, which could be merged downstream, but the interface of `gcexportdata.Read` doesn't really make it easy.

Given the constraints, the current approach makes sense.

compiler/compiler_test.go
nevkontakte: My thinking was that if the test is aimed at testing the invariant of `ReadArchive(WriteArchive(a)) == a`, it's best to test it directly. Adding code generation into the mix can cause additional complexity.

The tests for DCE you added in the other PR seemed comprehensive enough to me.

nevkontakte (approved):

compiler/compiler.go
grantnelson-wf: Oh, I see what you mean, that makes sense

compiler/compiler_test.go
grantnelson-wf: Oops, didn't mean "unexpected", I meant "unexported". So I gave this a try and apparently is does complain with unexported data. I added the following into the test prior to the `renderPackage`s:

```Go
if diff := cmp.Diff(origArchives, readArchives); diff != `` {
	t.Errorf("the reloaded archives didn't match:\n%s", diff)
}
```

and `cmp.Diff` panicked with

```Plain
panic: cannot handle unexported field at {map[string]*compiler.Archive}["command-line-arguments"].Package.path:
		"go/types".Package
	consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported
```

The `AllowUnexported` option requires all the unexported types to be passed in, but trying to pass in everything in "go/types" eventually reaches `types.object` that I can't pass into the option. The [`Exporter`](https://pkg.go.dev/github.com/google/go-cmp/cmp#Exporter) option does make it possible to test all unexported types (`cmp.Exporter(func(_ reflect.Type) bool { return true })`) but the documentation recommend against that and we're getting diffs in things like `pos`, `color_`, `id`, etc.

I think at least for right now it is easiest to just have the JS rendered as a way to confirm the important information is being reloaded and anything else not loaded identically will not affect the results.

compiler/compiler.go
grantnelson-wf: Changed it to use `srcModTime`

compiler/compiler_test.go
nevkontakte: That's fair, let's keep it as is then.

nevkontakte (approved):

PR #1361, opened 2025-02-10
Separate Parsing and Compiling

grantnelson-wf:
This is the start of several changes to get generic information propagated across package lines.

This changes how packages are loaded so that they are not converted into `Archive`s right away. This does mean that `Archive` caching has been disabled. With the `Archive`s creation delayed, we have two phases:

1. **Parse**: The parse phase uses `BuildPackage`s to determine the files that are needed to be parsed for a package and determine the imports that are required for a package. The result of the parse phase will be `ParsedPackage`s. All the `ParsedPackage`s needed for a project will be gotten prior to moving onto the compile phase.
2. **Compile**: The compile phase will take the root `ParsedPackage` and start compiling it, compiling any imported `ParsedPackage` as it goes. Each compile of a `ParsedPackage` results in an `Archive`. Once the `Archive`s have all been created, GopherJS continues as usual.

The point of this change is to prepare for following PR's. In the following PR's we can break up phase 2 into smaller steps. Right now the compile phase does Type Checking, code simplifying, link finding, anaylizing, then converting into JS. The analyzing includes determining flattening, blocking, etc information for the methods. At the point I left the `TODO` between phase 1 and 2. We can move several of those parts out of compile and do them to the whole project at once. We can run type checking and analyzing all at once to help propagate information across package lines. Once all the preliminary information has been gathered, then we output the `Archive`s.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -1175,7 +1175,7 @@ func (s *Session) getImportPath(path, srcDir string) (string, error) {
 		return importPath, nil
 	}
 
-	// Fall back to the slop import of the build package.
+	// Fall back to the slow import of the build package.
 	pkg, err := s.xctx.Import(path, srcDir, 0)
 	if err != nil {
 		return ``, err
--- a/build/build.go
+++ b/build/build.go
@@ -27,12 +27,11 @@ import (
 	"github.com/fsnotify/fsnotify"
 	"github.com/gopherjs/gopherjs/compiler"
 	"github.com/gopherjs/gopherjs/compiler/astutil"
+	"github.com/gopherjs/gopherjs/internal/testmain"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/neelance/sourcemap"
 	"golang.org/x/tools/go/buildutil"
-
-	"github.com/gopherjs/gopherjs/build/cache"
 )
 
 // DefaultGOROOT is the default GOROOT value for builds.
@@ -744,14 +743,60 @@ func (p *PackageData) InstallPath() string {
 	return p.PkgObj
 }
 
+// ParsedPackage is the results from building a package before it is compiled.
+type ParsedPackage struct {
+	ImportPath string // import path of package ("" if unknown)
+	Dir        string // directory containing package sources
+
+	// GoFiles is the parsed and augmented Go AST files for the package.
+	GoFiles []*ast.File
+	FileSet *token.FileSet
+	JSFiles []JSFile
+}
+
+// Imports calculates the import paths of the package's dependencies
+// based on all the imports in the augmented files.
+//
+// The given skip paths will not be returned in the results.
+// This will not return any `*_test` packages in the results.
+func (p *ParsedPackage) Imports(skip ...string) []string {
+	seen := make(map[string]struct{})
+	for _, s := range skip {
+		seen[s] = struct{}{}
+	}
+	imports := []string{}
+	for _, file := range p.GoFiles {
+		for _, imp := range file.Imports {
+			path := strings.Trim(imp.Path.Value, `"`)
+			if _, ok := seen[path]; !ok {
+				if !strings.HasSuffix(path, "_test") {
+					imports = append(imports, path)
+				}
+				seen[path] = struct{}{}
+			}
+		}
+	}
+	sort.Strings(imports)
+	return imports
+}
+
 // Session manages internal state GopherJS requires to perform a build.
 //
 // This is the main interface to GopherJS build system. Session lifetime is
 // roughly equivalent to a single GopherJS tool invocation.
 type Session struct {
-	options    *Options
-	xctx       XContext
-	buildCache cache.BuildCache
+	options *Options
+	xctx    XContext
+
+	// importPaths is a map of the resolved import paths given the srcDir
+	// and path. This is used to avoid redetermining build packages during
+	// compilation when we're looking up parsed packages.
+	importPaths map[string]map[string]string
+
+	// parsePackage is a map of parsed packages that have been built and augmented.
+	// This is keyed using resolved import paths. This is used to avoid
+	// rebuilding and augmenting packages that are imported by several packages.
+	parsedPackages map[string]*ParsedPackage
 
 	// Binary archives produced during the current session and assumed to be
 	// up to date with input sources and dependencies. In the -w ("watch") mode
@@ -767,6 +812,8 @@ func NewSession(options *Options) (*Session, error) {
 
 	s := &Session{
 		options:          options,
+		importPaths:      make(map[string]map[string]string),
+		parsedPackages:   make(map[string]*ParsedPackage),
 		UpToDateArchives: make(map[string]*compiler.Archive),
 	}
 	s.xctx = NewBuildContext(s.InstallSuffix(), s.options.BuildTags)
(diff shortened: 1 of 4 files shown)


flimzy (approved):
Read through it all, and it looks good to me. But happy to have @nevkontakte look at it too.

build/build.go
```diff
@@ -1035,21 +1159,50 @@ func (s *Session) BuildPackage(pkg *PackageData) (*compiler.Archive, error) {
...
+	if strings.HasSuffix(path, "_test") {
+		return path, nil
+	}
+
+	// Check if the import path is already cached.
+	if importPath, ok := s.importPaths[srcDir][path]; ok {
+		return importPath, nil
+	}
+
+	// Fall back to the slop import of the build package.
```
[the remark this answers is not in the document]
grantnelson-wf: oops, "slow" import, just a typo I didn't notice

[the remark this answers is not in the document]
grantnelson-wf: Fixed

grantnelson-wf:
I was continuing onto the compile break up and realized that `ParsedPackage` is nearly identical to `sources`. Since this one is already ready to go, I'm going to leave it as is, but the next PR will remove `ParsedPackage` and make `sources` exported as `Sources` with the extra fields in `ParsedPackage`.

However, if you'd like, I could just make that change in this PR instead of having another one.

build/build.go
```diff
@@ -744,14 +743,60 @@ func (p *PackageData) InstallPath() string {
...
+// The given skip paths will not be returned in the results.
+// This will not return any `*_test` packages in the results.
+func (p *ParsedPackage) Imports(skip ...string) []string {
+	seen := make(map[string]struct{})
+	for _, s := range skip {
+		seen[s] = struct{}{}
+	}
+	imports := []string{}
+	for _, file := range p.GoFiles {
+		for _, imp := range file.Imports {
```
nevkontakte, line 769: It's better to use resolved imports from https://pkg.go.dev/go/build#Package. Although import paths _usually_ match package name, they could also be relative paths and other sorts of funky.

build/build.go
```diff
@@ -744,14 +743,60 @@ func (p *PackageData) InstallPath() string {
...
-	options    *Options
-	xctx       XContext
-	buildCache cache.BuildCache
+	options *Options
+	xctx    XContext
+
+	// importPaths is a map of the resolved import paths given the srcDir
+	// and path. This is used to avoid redetermining build packages during
+	// compilation when we're looking up parsed packages.
+	importPaths map[string]map[string]string
```
nevkontakte, line 794: nit: If I understood this correctly:

 - The first key is the package source directory.
 - The second key is an unresolved import path from AST
 - The value is the corresponding resolved absolute import path.

If so, could you state this in the comment more explicitly?

build/build.go
```diff
@@ -910,19 +948,76 @@ func (s *Session) BuildFiles(filenames []string, pkgObj string, cwd string) erro
...
+	// TODO(grantnelson-wf): At this point we have all the parsed packages we
+	// need to compile the whole project, including testmain, if needed.
+	// We can perform analysis on the whole project at this point to propagate
+	// flatten, blocking, etc. information and check types to get the type info
+	// with all the instances for all generics in the whole project.
+
+	return s.compilePackages(parsed)
+}
+
+func (s *Session) buildTestPackage(pkg *PackageData) (*ParsedPackage, error) {
```
nevkontakte, line 978: Does this method actually perform compilation (directly or indirectly)? I don't think it does, in which case it may be better to name it something like `loadTestPackage`. Same consideration applies to the other sibling methods.

nevkontakte (approved):
Not a specific request, but something to consider and/or use as a reference:

https://pkg.go.dev/golang.org/x/tools/go/packages#Load both loads package metadata and (optionally) parses AST files there too. So, at a first glance, adding parsed sources to PackageData is not a bad idea.

More broadly, I suspect that we could simplify a lot of what build package does today by leaning more on go/packages. My idealized build process would look something like:

1) Use https://pkg.go.dev/golang.org/x/tools/go/packages#Load to load the entire program. Use the Confg.ParseFile hook to augment standard library on the fly [1].
2) Perform GopherJS-specific analysis.
3) Perform compilation.

I _think_ a lot of custom BuildContext stuff would simply disappear after that. We don't have to do that right now, but I thought I'd mention it in case it turns out to be a more expedient way to achieve what we need.

grantnelson-wf:
@nevkontakte (in response to [this comment](https://github.com/gopherjs/gopherjs/pull/1361#pullrequestreview-2632918663)),

I had a WIP ([this closed PR](https://github.com/gopherjs/gopherjs/pull/1357)) where I was attempting to use [golang.org/x/tools/go/packages#Load](https://pkg.go.dev/golang.org/x/tools/go/packages#Load) but was having trouble figuring out how to perform the augmentation during loading so that we could have `Load` perform the type checks on the augmented files with the augmented imports. I was using `ParseFile` but I think it was an issue with `FileSet`.

The main issue I was having seemed to be that `FileSet` didn't report enough bytes for the augmented file. The resulting augmented file could be longer than the original so we couldn't override the location with [`/token#File.AddLineColumnInfo`](https://pkg.go.dev/go/token#File.AddLineColumnInfo). The experiment I wanted to try was to use a separate `FileSet` when loading the original and override files, then perform the augmentation, then create a "new file" for the `FileSet` that is the correct length for the augmented file. I just haven't been able to try it out to see if it fixes the issues I was hitting. I also think it could have been because of how I was trying to test it and we could avoid those issues without fixing the `FileSet`... I don't know yet.

Another thing I wanted to try, if I couldn't figure out the above, was to just load the packages as if not in GopherJS with `packages#Load`, then, after getting the packages back, augment the files for GopherJS, add the additional packages for GopherJS, and lastly run type checks on them.

Anyway, because I was having issues getting it to work, I decided to try to finish that later, if no one else has figured it out. However, I feel you are right, a lot of the Config stuff and stuff getting up to the `ParsedPackage`s could be removed if we did that.

build/build.go
grantnelson-wf: I added this to try to automatically determine the imports that were added via augmentation or the generation of the testmain. In `loadTestPackage` and `loadPackages` (called `buildTestPackage` and `buildPackages` just before I renamed them per request), we use the `build#Package` to determine the original imports, we augment the files, then I use this method to find any Imports we added in the augmented files. I know we limit which ones can be added, I just was trying to avoid adding ones we don't need. Especially in testmain since that can add `os` and other imports. The passed in `skip` is so that we can skip over any already imported from `build#Package`.

Although, it probably would be a good idea to update the `build#Package` as well or perform one set of imports after augmentation. I was trying not to disrupt too much of the `SrcModTime` stuff right away even through we won't be caching, maybe we could cache the `ParsedPackage`s someday.

build/build.go
grantnelson-wf: I tried to make it clearer. Let me know if that's still not explicit enough.

build/build.go
grantnelson-wf: Renamed:
- `buildTestPackage` -> `loadTestPackage`
- `buildPackages` -> `loadPackages`
- `buildImportPathWithSrcDir` -> `loadImportPathWithSrcDir`

build/build.go
nevkontakte: Alright. The main thing I want to ensure is to maintain correctness - now and going forward. From what I could see in the code, the use of this method shouldn't create correctness issues in the scope of the current PR. Maybe to avoid this turning into a surprise rake in the future we could rename the method to more clearly indicate dangers? Something like `unresolvedImports`? And add a warning in the doc comment.

build/build.go
nevkontakte: Thank you!

nevkontakte:
@grantnelson-wf thanks for the details! That all sounds reasonable. If I have some spare time one of the weekends, I may try to experiment with it myself. Perhaps I will be able to find a working solution, or perhaps prove definitively that it's not viable :)

build/build.go
nevkontakte: LGTM, thanks!

nevkontakte (approved):

build/build.go
grantnelson-wf: This is a great idea. I'll rename it in the follow up ticket, https://github.com/gopherjs/gopherjs/pull/1362.

PR #1364, opened 2025-03-05
Perform cross package type analysis

grantnelson-wf:
#### These changes are for the 3rd task towards generics

> 3. build system needs to be changed to collect generic instance information and propagate it across package boundaries. This would require reorganizing build package into loading all packages first, analyzing them, and then passing to the compiler, which is different from today's behavior where imported packages are loaded and compiled on-demand when compiler needs them. The tricky part here would be making sure build cache accounts for the fact that dependent packages may influence the set of generic instances of the dependency packages. - [from this slack comment by nevkontakte](https://gophers.slack.com/archives/C039C0R2T/p1721859938577319)

#### What these changes do

- This leverages the prior work to making Sources a more important part of the Build process.
  - This add the type information and linkname information into the Sources.
- This pulls the preparation of the Sources out of the compile phase. During preparation:
  - All the source files are sorted within each package by name
  - The Sources type information and type package are determined and stored in Sources
  - All the Go linknames are gotten
  - All the source files are simplified
  - All the instance information is collected
  - Analysis is run and propagated across package boundaries
- Once all the sources have been prepared, they can be compiled

### DIFF
--- a/compiler/package.go
+++ b/compiler/package.go
@@ -15,7 +15,6 @@ import (
 	"github.com/gopherjs/gopherjs/compiler/sources"
 	"github.com/gopherjs/gopherjs/compiler/typesutil"
 	"github.com/gopherjs/gopherjs/internal/errorList"
-	"github.com/gopherjs/gopherjs/internal/experiments"
 )
 
 // pkgContext maintains compiler context for a specific package.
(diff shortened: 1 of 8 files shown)


compiler/package.go
```diff
@@ -295,16 +214,74 @@ func Compile(srcs sources.Sources, importContext *ImportContext, minify bool) (_
...
 }
 
+// PrepareAllSources prepares all sources for compilation by
+// parsing go linknames, type checking, sorting, simplifying, and
+// performing cross package analysis.
+// The results are stored in the provided sources.
+//
+// All sources must be given at the same time for cross package analysis to
+// work correctly. For consistency, the sources should be sorted by import path.
+func PrepareAllSources(allSources []*sources.Sources, importer sources.Importer, tContext *types.Context) error {
```
grantnelson-wf, line 234: This is the big change. In prior tickets I changed the code so we'd get all the Sources for all the packages prior to doing compile. This is all the preparation that was done in compile, instead being run on all the Sources at the same time prior to any package compiling.

To prevent this method from having tons of maps to store all the data per package, I just changed Sources to have the data stored in them and changes the Sources methods to populate those data field instead of returning the data. I also broke up some of those methods as needed to get all the parts prepared in the correct order.

flimzy (approved):
I have read through this PR a couple of times, and nothing stands out to me to me as a problem.

nevkontakte (approved):
Apologies for the delay. Something was bothering me about thing change, but I couldn't quite figure out what exactly. However, I think I have :)

That thing is that build.Session API is still mostly trying to pretend it works on single packages, despite the fact that it loads the full dependency tree and stores it inside. However, in reality we are working on a full program, not a single package.

The dissonance is made worse by the fact that we sometimes synthesize parts of the program (like test packages), but that fact is almost unobservable.

A better API (for testability and understanding both) would be where each processing stage is (at least conceptually) a pure function with its input a full-program representation of the previous stage, and the return value, again, a full program of the next stage. Compare this with today, where are a just passing the reference to an entry package, and keep the majority of the information internally.

So, at some point, we probably should build on the idea you've started with Sources and implement types like LoadedProgram, AnalyzedProgram, CompiledProgram and GeneratedSource.

All that said... From practical perspective, you PR is a fine incremental step forward. Now that I understand what was bothering me, I also know that it's not an immediate concern, just a future direction.

grantnelson-wf:
Totally, @nevkontakte. I totally agree, that this is not the best solution. I was going for three goals; 1. get it working, 2. do it with least numbers of line changes possible, 3. it must still be maintainable. I'm completely onboard with this being just a stepping stone towards a better design. Thank you for explaining your concern.

I think one thing that would help would be to get the [`x/tools/go/packages`](https://pkg.go.dev/golang.org/x/tools/go/packages#Load) working. That would reduce the "working on single packages" feeling since we'd get all the packages in one go. Also, you're right about using types to enforce the order preparation steps so that it doesn't have as many unenforceable "Do this before..." comments.

Issue #1352, opened 2024-11-22
GopherJS does not support Go 1.23 due to new restrictions.


different55:
When installing according to the instructions in the README:

```
go install github.com/gopherjs/gopherjs@v1.19.0-beta1
```

On an amd64 darwin machine running Go 1.23.1, I get this final line of output:

```
link: github.com/visualfc/goembed/parser: invalid reference to go/build.parseGoEmbed
```

Doing a little digging, it seems like this is because goembed uses the `//go:linkname` directive to access Go's own internal `parseGoEmbed` function, which as of 1.23 is apparently [not allowed](https://github.com/golang/go/issues/67401).

As a workaround, I can confirm that `-ldflags=-checklinkname=0` works to roll back this new behavior temporarily, maybe this should be documented in the README until this issue is resolved more cleanly?

nevkontakte:
Looks like you've already reported https://github.com/visualfc/goembed/issues/8 to the package author. Hopefully they fix it, if not the only real option we have is to reimplement this library ourselves, which is quite a bit of work.

IngwiePhoenix:
Ran into this too. Seems like the dependent library hasn't been maintained in three years... Quite unfortunate.

Naatan:
Any plans to address this? This means GopherJS is effectively two Go versions behind now, unless I'm missing something. Surprised this isn't a bigger deal.

IngwiePhoenix:
Basically it boils down to:

- One dependency (a parser for `//go:embed` statements)
- ...that requires one function (`build.parseGoEmbed`) by using the [`linkname` directive](https://pkg.go.dev/cmd/compile#hdr-Linkname_Directive)
- ...which isn't exported: https://github.com/golang/go/blob/3c05790d4847ba2d0d06b5ba3f82b311a99d531c/src/go/build/read.go#L556

This means, that this particular function, would have to be copied ("hoisted" if you will) into the goembed package. But since that maintainer is no longer maintaining that package, it would have to go one level up further and instead become part of gopherJS itself.

...any takers?

msvitok77:
Hi, if interested you can use this: https://github.com/msvitok77/goembed. I've released version **v0.3.5**

nevkontakte:
Thanks @IngwiePhoenix and @msvitok77 for your contribution! The fix has been merged now.

PR #1367, opened 2025-04-08
Fix: `linkname` no longer causes issues for 1.24.x and probably others

IngwiePhoenix:
This is a dead simple PR to address #1352 by updating the deps.

Not sure if I should also bump a patch version - and if so, where? - so here is the super simple, two commits, three files and three lines PR. =)

### DIFF
--- a/build/embed.go
+++ b/build/embed.go
@@ -8,7 +8,7 @@ import (
 	"go/token"
 	"strconv"
 
-	"github.com/visualfc/goembed"
+	"github.com/msvitok77/goembed"
 )
 
 func buildIdent(name string) string {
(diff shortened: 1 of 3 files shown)


IngwiePhoenix:
Long-term, I'd recommend doing that - and considering how limited the use of this function is, it could possibly be integrated as a sub-package in this repo and thus "internalized" rather than vendored. Just a thought tho.

nevkontakte (approved):
I agree that adopting a fork, or simply pulling in the logic into gopherjs would be ideal in the long term. In the meantime, I've reviewed the changes in the fork and they seem okay, so we can go ahead with this PR.

PR #1370, opened 2025-05-12
Adding nested type support

grantnelson-wf:
## Summary

Adding type parameters from a function into the nested types inside of that function. For example:

```Go
func Foo[T any]() {
	type Bar struct { x T }
	type Baz[U any] struct { x T; y U }
	...
}
```

The type `Bar` is concrete (i.e. not generic) however it acts like it is generic because it is nested in a generic function. If the function is called like `Foo[string]()` then `Bar` innately is different from if the function is called like `Foo[int]()`. The nesting type parameters is shown like `Bar[T;]` and instantiated like `Bar[string;]` and `Bar[int;]`.

In Go the `;` in the type parameters isn't present if the nested type is concrete. `Baz` is generic meaning the `;` indicates the separation between the nested type parameters and the objects own type parameters. For example if `Baz[string]{}` is expressed inside of `Foo[int]`, then a `Baz[int; string]` is instantiated. If `Baz[T]{}` is expressed inside of `Foo[X]`, then a `Baz[X; X]` is instantiated.

When creating instances for a nested type, all combinations of nest function instances and nesting type instances need to be determined. For example if `Foo` has the instances `Foo[string]` and `Foo[int]`, and `Baz` has the instances `Baz[bool]` and `Baz[string]`, then the resolved instances for `Baz` are `Baz[string; bool]`, `Baz[string; string]`, `Baz[int; bool]`, and `Baz[int; string]`. While reading the instances from `types.Info`, an instance may have a type parameter in it, e.g. `Baz[T]`. We replace the type parameter with the nest function's instance type arguments, meaning `Baz[T]` resolves to `Baz[int; int]` and `Baz[string; string]`, but not `Baz[int; string]` nor `Baz[string; int]`.

Note: I used `TNest` or similar to store the type arguments from the nest function or method as a separate set from `TArgs`. This is different from what Go does which is using an index for the "implicit" type arguments to differentiate between the two parts, [seen here](https://cs.opensource.google/go/go/+/release-branch.go1.19:src/cmd/compile/internal/noder/reader.go;l=693-713;bpv=1;bpt=0). I felt it was easier to maintain and to fit into the existing code better than using an index.

----

### Fixing Nested Type Arguments

This change does not include a fix for when a type argument itself is nested. For example:

```Go
func Foo[T any]() {
	type Bar struct { X T }
	type Baz[U any] struct { }
	_ = Baz[Bar]{}
	...
}
```

The `Baz` is instantiated correctly using the nesting type argument for `Foo` and the argument given to it `Bar`. However, `Bar` has not been instantiated correctly with the nesting type argument for `Foo`, meaning `Bar` is still generic with the underlying type of `struct { X T }`. If `Foo[int]()` is called, the `Bar` used for the type argument for `Baz` needs to be `Bar[int;]` with `struct { X int }`.

I didn't fix this issue yet because it feels very unlikely, although possible, to happen, and I found it while finishing up this PR. So, instead of delaying this initial work on nested types, I left some tests that demonstrate the issue and skipped them with a `TODO(grantnelson-wf)` comment.

### Fixing DCE

This change does not include an update to the DCE to remove unused instances caused by nested types. For example:

```Go
func Foo[T any]() {
	type Bar struct { }
	...
}

func aliveCode() {
	Foo[string]()
}

func deadCode() {
	Foo[int]()
}
```

In the above code the instances `Foo[string]` and `Foo[int]` are created, then `Foo[int]` is found to be dead and eliminated. However, the instance `Bar[string;]` and `Bar[int;]` are created by their respective nest function instances. Right now our DCE doesn't take into account the nesting type parameters so will see all instances of `Bar` as alive because `"Bar"` is used in the alive `Foo[string]`. This leaves the dead code `Bar[int]`, also seen as `"Bar"`, alive when it should be eliminated.

A following ticket will fix the DCE to take into acount the nesting type parameters so that dead nested types will be properly removed.

### Numbering Nested Types Issue

This change does not include any of the numbering of types like Go does. For example:

```Go
type Biz[U any] struct{}

func Foo() any {
	type Baz struct{}
	return Biz[Baz]{}
}

func Bar() any {
	type Baz struct{}
	return Biz[Baz]{}
}
```

When the types returned from `Foo` and `Bar` are printed, they are `main.Biz[main.Baz·1]` and `main.Biz[main.Baz·2]` respectively. The `·1` and `·2` are indications of which declaration of `Baz` is being used. They are not shown in all cases and never show for package level declarations. The numbering is applied in the order the files are parsed and nested types are declared in those files.

The number increases monotonically with each new type declaration regardless of their name and type, meaning if the `Baz` in `Bar` is changed to `type Cat interface{}` and `Biz` is instantiated with `Cat` in `Bar`, the types are `main.Biz[main.Baz·1]` and `main.Biz[main.Cat·2]`. The type from `Foo` didn't change and `Cat` is still denoted with a `·2` since it is the second nested type.

Since these numbers aren't being collected nor printed, we can not remove `typeparam/nested.go` from the known failures of the Gorepo Tests yet. The test will not pass because the outputs will not match without the nested type numbering. So, not only are does that test hit some "Nested Type Arguments" (the `main.U`'s aren't being instantiated correctly), but the difference in the outputs looks like:

```Diff
+ 0,3: main.T[int;int]
+ 4,7: main.T[int;main.U[int]]
+ 22,23: main.T[main.Int;main.Int]
+ 26,27: main.T[main.Int;main.U[main.Int]]
- 0,3: main.T·2[int;int]
- 4,7: main.T·2[int;main.U·3[int;int]]
- 22,23: main.T·2[main.Int;main.Int]
- 26,27: main.T·2[main.Int;main.U·3[main.Int;main.Int]]
```

### DIFF
--- a/compiler/internal/typeparams/collect.go
+++ b/compiler/internal/typeparams/collect.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"go/ast"
 	"go/types"
+	"strings"
 
 	"github.com/gopherjs/gopherjs/compiler/typesutil"
 	"github.com/gopherjs/gopherjs/internal/govendor/subst"
@@ -12,27 +13,67 @@ import (
 // Resolver translates types defined in terms of type parameters into concrete
 // types, given a mapping from type params to type arguments.
 type Resolver struct {
+	tParams *types.TypeParamList
+	tArgs   []types.Type
+	parent  *Resolver
+
+	// subster is the substitution helper that will perform the actual
+	// substitutions. This maybe nil when there are no substitutions but
+	// will still usable when nil.
 	subster *subst.Subster
 	selMemo map[typesutil.Selection]typesutil.Selection
 }
 
 // NewResolver creates a new Resolver with tParams entries mapping to tArgs
 // entries with the same index.
-func NewResolver(tc *types.Context, tParams []*types.TypeParam, tArgs []types.Type) *Resolver {
+func NewResolver(tc *types.Context, tParams *types.TypeParamList, tArgs []types.Type, parent *Resolver) *Resolver {
 	r := &Resolver{
+		tParams: tParams,
+		tArgs:   tArgs,
+		parent:  parent,
 		subster: subst.New(tc, tParams, tArgs),
 		selMemo: map[typesutil.Selection]typesutil.Selection{},
 	}
 	return r
 }
 
+// TypeParams is the list of type parameters that this resolver
+// (not any parent) will substitute.
+func (r *Resolver) TypeParams() *types.TypeParamList {
+	if r == nil {
+		return nil
+	}
+	return r.tParams
+}
+
+// TypeArgs is the list of type arguments that this resolver
+// (not any parent) will resolve to.
+func (r *Resolver) TypeArgs() []types.Type {
+	if r == nil {
+		return nil
+	}
+	return r.tArgs
+}
+
+// Parent is the resolver for the function or method that this resolver
+// is nested in. This may be nil if the context for this resolver is not
+// nested in another generic function or method.
+func (r *Resolver) Parent() *Resolver {
+	if r == nil {
+		return nil
+	}
+	return r.parent
+}
+
 // Substitute replaces references to type params in the provided type definition
 // with the corresponding concrete types.
 func (r *Resolver) Substitute(typ types.Type) types.Type {
-	if r == nil || r.subster == nil || typ == nil {
+	if r == nil || typ == nil {
 		return typ // No substitutions to be made.
 	}
-	return r.subster.Type(typ)
+	typ = r.subster.Type(typ)
+	typ = r.parent.Substitute(typ)
+	return typ
 }
 
 // SubstituteAll same as Substitute, but accepts a TypeList are returns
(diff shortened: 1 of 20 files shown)


compiler/internal/typeparams/collect.go
```diff
@@ -131,21 +192,65 @@ func (c *visitor) Visit(n ast.Node) (w ast.Visitor) {
...
+	c.addInstance(obj, nil, c.resolver.TypeArgs())
+}
+
+func (c *visitor) addInstance(obj types.Object, tArgList *types.TypeList, tNest []types.Type) {
+	tArgs := c.resolver.SubstituteAll(tArgList)
+	if isGeneric(tArgs...) {
+		// Skip any instances that still have type parameters in them after
+		// substitution. This occurs when a type is defined while nested
+		// in a generic context and is not fully instantiated yet.
+		// We need to wait until we find a full instantiation of the type.
```
grantnelson-wf, line 234: This also happens when the type argument itself is a nested type that hasn't been instantiated correctly. This is one of the fixes that I'm doing as a follow up PR. See "Nested Type Arguments" in summary for more information.

flimzy (approved):

PR #1374, opened 2025-05-28
Handling nested type arguments

grantnelson-wf:
This fixes some issues (see "Fixing Nested Type Arguments" in summary of #1370).
This also cleans up some of the resolver code to make it easier to prepare for resolution of types.

This change fixes when a type argument itself is nested. For example:

```Go
func Foo[T any]() {
	type Bar struct { X T }
	type Baz[U any] struct { }
	_ = Baz[Bar]{}
	...
}
```

The `Baz` is instantiated correctly using the nesting type argument for `Foo` and the argument given to it `Bar`. However, `Bar` has not been instantiated correctly with the nesting type argument for `Foo`, meaning `Bar` is still generic with the underlying type of `struct { X T }`. If `Foo[int]()` is called, the `Bar` used for the type argument for `Baz` needs to be understood to be `Bar[int;]` with `struct { X int }`.

Initially I intended to change the substitution to always modify the underlying struct in a named type. However, that causes problems in the instance map since the instance map uses the object pointer as a key and the fully substituted object has a different pointer. Instead, following how `types.Instantiate` works, I made the instances allow for lazy substitution, i.e. unsubstituted type parameters are allowed in the context of a named type with those type parameters and properly substituted type arguments.

In the above example this would mean `Bar` would be seen as `struct { X T }` but that is now acceptable to the code since it is in context where `T` is defined. This is acceptable since the transformation code later on continues to substitute as needed, so the `T` will be substituted to `int` before it is used.

The only draw back to not fully substituting the underlying types is that when printing `Baz[Bar]` in `Foo[int]`; Go will output `Baz[Bar[int;]]` (ignoring the dot numbering and the extra `;`), whilst now, GopherJS will output `Baz[Bar]` without properly indicating the nesting context. We can swing back around to this later, I didn't want get stuck on this too long when it is rare that nested type arguments or deep nesting is used in normal code.

### DIFF
--- a/compiler/decls.go
+++ b/compiler/decls.go
@@ -451,13 +451,7 @@ func (fc *funcContext) newNamedTypeVarDecl(obj *types.TypeName) *Decl {
 func (fc *funcContext) newNamedTypeInstDecl(inst typeparams.Instance) (*Decl, error) {
 	originType := inst.Object.Type().(*types.Named)
 
-	var nestResolver *typeparams.Resolver
-	if len(inst.TNest) > 0 {
-		fn := typeparams.FindNestingFunc(inst.Object)
-		tp := typeparams.SignatureTypeParams(fn.Type().(*types.Signature))
-		nestResolver = typeparams.NewResolver(fc.pkgCtx.typesCtx, tp, inst.TNest, nil)
-	}
-	fc.typeResolver = typeparams.NewResolver(fc.pkgCtx.typesCtx, originType.TypeParams(), inst.TArgs, nestResolver)
+	fc.typeResolver = typeparams.NewResolver(fc.pkgCtx.typesCtx, inst)
 	defer func() { fc.typeResolver = nil }()
 
 	instanceType := originType
@@ -469,10 +463,7 @@ func (fc *funcContext) newNamedTypeInstDecl(inst typeparams.Instance) (*Decl, er
 			}
 			instanceType = instantiated.(*types.Named)
 		}
-		if len(inst.TNest) > 0 {
-			instantiated := nestResolver.Substitute(instanceType)
-			instanceType = instantiated.(*types.Named)
-		}
+		instanceType = fc.typeResolver.Substitute(instanceType).(*types.Named)
 	}
 
 	underlying := instanceType.Underlying()
--- a/compiler/expressions.go
+++ b/compiler/expressions.go
@@ -531,7 +531,7 @@ func (fc *funcContext) translateExpr(expr ast.Expr) *expression {
 		case *types.Signature:
 			return fc.formatExpr("%s", fc.instName(fc.instanceOf(e.X.(*ast.Ident))))
 		default:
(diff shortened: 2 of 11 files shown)


grantnelson-wf:
@nevkontakte Sorry, I wasn't very clear in my comment about the pointers. Instead of the example from the comment, lets use this one:

```Go
func Foo[T any]() {
	type Bar[U any] struct {
		X T
		Y U
	}
}
```

In the above we have the type `Bar[T; U]` where `T` is the implicit type. When we use `types.Instantiate` on a named type it will not perform the substitution of the underlying type (like I has mistakenly thought in the prior PR). Named types are ["substituted lazily"](https://cs.opensource.google/go/go/+/refs/tags/go1.19.13:src/go/types/instantiate.go;l=118), meaning that `Bar[T; int]` would still have the underlying type of `struct { X T; Y U }`, not `struct { X T; Y int }` like I had thought it would. (This is what I was trying to say when I said "this would mean Bar would be seen as struct { X T }", the underlying type wouldn't be substituted like I had expected it to be.)

I had to fix the code checking for concrete types in the visitor, so that it wouldn't have false negatives caused by the underlying types. But I still wanted to detect that `Bar[T; int]` is generic. Since the named type only had type parameters and type arguments but no implicit/nest type parameters/arguments, we couldn't stop at a named type and assume it was concrete because it had concrete type arguments, we had to still check the underlying struct to find the `T`. (If the type didn't have `X T`, then it would be concrete because all the type parameters had been realized, even though it is nested in a generic type.)

When substituting `T` in `Bar[T; U]`, a new named type with the underlying struct is created. For example, if `T` is substituted with `bool`, we would get `Bar[bool; U]` with the underlying type being `struct { X bool; Y U }`. This new named type would not have an origin type and no type arguments. It appears like a unique generic `Bar[U]` in every instance of `Foo`. Each instance would have different pointers because of the new types and underlying types.

This became an issue with how I had updated the `Instance` to include nesting types. Having a brand new `Bar[U]` object for every instance of `Foo[T]` caused problems because `Bar[bool; U]` wasn't anywhere in `types.Info`, only the original `Bar[T; U]` object is (i.e. the `Bar` object created by the type checker). Since we use the object in the Instances for various things,  it was easier to make the object in an `Instance` be that original `Bar`, than it would be trying to add an origin to the `Instance` struct and have the `Instance`'s object point to the substituted version.

Overall this means that the whole underlying type was "substituted lazily" including the type arguments themselves. To use it, we'd have to keep a map from type parameters to type arguments, including the implicit ones from the nest, so we can substitute while processing the underlying types. Which is great since the code you had already written (i.e. the replacer) already did that; I just had to adjust how I was thinking about nested types from the prior PR to this PR and adjust the code I wrote to match these changes.

nevkontakte:
Ah, I see now, that makes sense. As an aside, in other places we haven't shied away from adding more data into `types.Info`, so that's also always an option.

nevkontakte (approved):

PR #1378, opened 2025-07-02
Adding Type Initialization Grouping and Ordering

grantnelson-wf:
When the JS packages are being setup, type instances could require type information from packages which have not been setup yet. For example, `list.List[Cat]` could be defined inside the `cat` package and import the `list` package. The type instance `List[cat.Cat]` is initialized inside the `list` package. This will cause a failure because `cat.Cat` hasn't been setup yet since it requires `list` to be setup first as an import. We can't move `List[Cat]` to the cat package because `List[T]` may access unexposed data inside the `list` package.

This is the first part of two (or more) tickets. To solve the type initialization, each declaration will be given a group number. There will be $n$ groups and each group number is $g_i$ where $`0 \le i < n`$. For any group, $g_i$, the prior groups, $`\{g_0 ... g_{i-1}\}`$, will have to be initialized first. All the declarations with the same group number are part of the same group. However within that group, the declarations will still need to be initialized in the current ordering where imports are done before importers and types are initialized before they are used in pointers, slices, etc.

This PR determines the grouping and order of the groups, then sets the group number for each `Decl`. This PR mainly consists of two new packages: `sequencer` and `grouper`. The `sequencer` is a generalized tool for determining ordering and grouping of items depending on the items' dependencies. The `grouper` stores information about the declarations' types so that it can use the `sequencer` to group and order the declarations based on type.

Following ticket will use that group number to populate a "startup map" where the key is the group number and the value is a set of functions that need to be run. Each function will initialize all the `Decl`s for a particular group for a single package. Then once all the packages have been added, the runtime will call each group of functions in order.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/gopherjs/gopherjs/compiler/internal/dce"
+	"github.com/gopherjs/gopherjs/compiler/internal/grouper"
 	"github.com/gopherjs/gopherjs/compiler/linkname"
 	"github.com/gopherjs/gopherjs/compiler/prelude"
 	"golang.org/x/tools/go/gcexportdata"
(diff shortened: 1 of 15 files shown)


grantnelson-wf:
I think I've figured out what you were thinking and think I may have a solution where the "cheat" is done programmatically. I'm breaking up the `DeclCode` into more specific types of declarations and adding a `$finishSetup` method. While trying to understand what you meant I learned about how JS handles variables in closures (like those for constructors) and that was the "ohhhh, I get it" moment for me. I was trying to get every type initialized prior to it being used anywhere, including in a closure, but in JS you don't have to do that. I'm testing my code-up of your solution and will have a new PR up soon with that work and I'll close this PR (unless I run into something that may require type ordering still, then I'll leave it open so we can discuss alternatives more).

Issue #1330, opened 2024-07-22
Proposal: More strict native override checking


grantnelson-wf:
# More strict native override checking

## Problem

Currently, anything in the native overrides, without a directive saying otherwise, overrides anything that has the same identifiers in the same package. The Go repo has strict rules about when an how exported types, variable, const, methods, etc are modified, however, non-exported methods may be changed each revision. Since many of the overrides are for non-exported methods, it would be helpful to write warnings when native overrides are out-of-date. This issue proposes a new set of directives that will help contributors keep track of changes in Go.

This proposed change would be most helpful when bumping the versions of GopherJS to a new version of Go. The goal is to help save time and produce better results when doing version bumps. This change would also help a little when doing updates within a revision for any update which changes the native overrides. This change will also help contributors know what the prior contributor intended to be overriding code verse helper code.

The drawback to this change is we'd have to go through all the native overrides and apply compiler directives. This cost should be a onetime effort to implement the strict checking and very minimum overhead otherwise.

## Proposed Solution

I propose a tool that can be run on demand that walks through every package in the Go repo and runs the build process on it. This will allow us to check every package in the native overrides for GopherJS repo. The builder should have a `strictOverrides` flag added to it that defaults to false. The warnings described below will be outputted when the `strictOverrides` flag is set to true. When this new tool is run, `strictOverrides` will be set to true, so that these new warnings are only outputted when a contributor is working on a version bump. This tool could also be run as a GHA CI job.

A warning should be outputted for any packages found in the overrides which does not exist in the Go repo. This indicates a package was deleted or moved and that override code should also be deleted or moved. The warning could be something like "warning: the package at [package path] no longer exists in the Go repo".

To make this work, we'll need two new directives: `replace` and `new`. These names are just preliminary names for this proposal. These two directives output warnings (or errors) when specific conditions aren't met. Since a whole package may be moved, deleted, refactored, etc. we should only output the first N (probably 5-10) warnings from any package.

### `replace` directive

The `replace` directive (`//gopherjs:replace`) would tell the builder that it **must find a matching identifier** in the same package that is being overridden, otherwise a warning should be outputted. The `replace` directive checks for overridden code being renamed, migrated, or removed from the code base. The outputted warning could be something like "warning: the native override for [identifier] in [package path] was not found: [override file]:[lineNo]".

[...]

PR #1414, opened 2026-02-20
[go1.21] Adding `new` / `replace` optional override directives

grantnelson-wf:
This is part of the proposal [issue #1330, "Proposal: More strict native override checking"](https://github.com/gopherjs/gopherjs/issues/1330). This adds the `gopherjs:new` and `gopherjs:replace` optional directives. Additionally, I added the new missing method to bytealg and added these directives as a test. Since this is being added to the go1.21 integration branch, CI will continue to fail. This code includes [the update to master](https://github.com/gopherjs/gopherjs/pull/1413) so that PR should be merged before this one.

Between Go1.20 and Go1.21, the Go developers refactored reflect and internal/reflectlite by moving a lot of code over to internal/abi. Since so much code was in internal packages; the developers weren't required to keep exported code consistent other than in reflect. Since there was so much code being moved, it was difficult for me to determine what was new code to go1.20, code being replaced in go1.20, then which code is new in go1.21 and replaced in go1.21. Any code that was a replacement for go1.20 of code that got moved, removed, or renamed in go1.21 would suddenly become new code we are adding.

I didn't want to leave around a bunch of code that hopefully had no effect (and hopefully just removed with dead code elimination), so I implemented this part of the proposal to help me track how our overrides are being applied.

These overrides are compiler enforced "developer intensions" just like the [override modifier in C#](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/override) and other modifiers in many languages. The code could quietly decide if we are overriding code or adding new code, or with these directives we can ensure that overrides work as the developer intended at compile time.

The full proposal is probably overkill at this time because of how strict is. These directives are not required so we don't have to add them to all of the overrides in one pass. The only change that this will effect is any of the existing override directives (i.e. `purge`, `keep-original`, and `override-signature`) will start acting like `gopherjs:replace` and raise an error if the code that they are attached to doesn't exist.

Instead of adding this code, then removing it once I have reflect, reflectlite, and abi working, I felt this would be a good time to add these directives. We already have to go through the native overrides while upgrading to go1.21 so any of the existing override directives can be updated as needed and we can add in these new directives if we feel it will help us when we upgrade to go1.22.

However, if you think that this is not a good idea to add; I have no problem skipping it and removing it from the reflect, reflectlite, and abi updates.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -124,13 +124,26 @@ func ImportDir(dir string, mode build.ImportMode, installSuffix string, buildTag
 // overrideInfo is used by parseAndAugment methods to manage
 // directives and how the overlay and original are merged.
 type overrideInfo struct {
+	// replace indicates that the original code is expected to exist.
+	// If the original code is not present, then error to let us know
+	// the original code has been removed or renamed.
+	// This will be set to false once the original code has been found.
+	replace bool
+
+	// new indicates that the override code is expected to be new.
+	// If the original code is present, then error to let us know
+	// that we are overriding something that we did not expect to replace.
+	// This lets us know of new or renamed code in the original that happens
+	// to collide with our overrides.
+	new bool
+
 	// keepOriginal indicates that the original code should be kept
 	// but the identifier will be prefixed by `_gopherjs_original_foo`.
 	// If false the original code is removed.
 	keepOriginal bool
 
-	// purgeMethods indicates that this info is for a type and
-	// if a method has this type as a receiver should also be removed.
+	// purgeMethods indicates that this info is for a type and if a method has
+	// this type as a receiver then the method should also be removed.
 	// If the method is defined in the overlays and therefore has its
 	// own overrides, this will be ignored.
 	purgeMethods bool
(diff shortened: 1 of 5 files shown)


build/build.go
```diff
@@ -346,12 +379,14 @@ func augmentOriginalImports(importPath string, file *ast.File) {
 // augmentOriginalFile is the part of parseAndAugment that processes an
 // original file AST to augment the source code using the overrides from
 // the overlay files.
-func augmentOriginalFile(file *ast.File, overrides map[string]overrideInfo) {
+func augmentOriginalFile(file *ast.File, overrides map[string]overrideInfo, found map[string]struct{}) {
```
nevkontakte, line 382: Could you update the comment to document `overrideInfo` and `found`? In particular, since the key is a plain string, it would be useful to write down the expected key structure.

nevkontakte (approved):
I like this idea, it can help us get to a better-defined overrides incrementally. Thanks!

build/build.go
grantnelson-wf: Good call. I'll add to that comment.

Since this is going into an integration branch and I have more PRs to follow, I've added that change into the next PR I'm working on

Issue #1415, opened 2026-02-20
Add Go 1.21 support


grantnelson-wf:
Now that we've released 1.20 support, filing this issue to track work related to [1.21 support](https://go.dev/blog/go1.21).

Release notes: https://go.dev/doc/go1.21

Developer guide for new version support in [wiki](https://github.com/gopherjs/gopherjs/wiki/Development:-New-Go-version-support-guide).

These changes are being added to the [go1.21 integration branch](https://github.com/gopherjs/gopherjs/tree/go1.21)

grantnelson-wf:
[Go 1.21 is now supported](https://github.com/gopherjs/gopherjs/releases/tag/v1.21.0)

PR #1417, opened 2026-02-24
[go1.21] Updated reflect, reflectlite, and abi

grantnelson-wf:
There was a large refactor in `reflect` and `reflectlite` that moved a lot of the duplicate code out to `abi`. For example, they consolidated the `reflect.name` and `reflectlite.name` object into `abi.Name`. I updated the native overrides to adjust for these changes. I also removed anything in `reflectlite` that was no longer needed.

The main issues I ran into was that I didn't remember or wasn't aware of the `kindType` casts. I added comments into the code to help devs with future updates. With many of the type methods moving to `abi`, there was a new type of object created, extensions on kind types, e.g. `type structType { abi.StructType }`, that needed another way to cast. The code now would cast between `*rtype`, `*abi.Type`, `*abi.StructType`, and `reflect.structType`. I added a function into reflect to handle this complication. `toKindTypeExt` will perform the additional types of casting that couldn't be handled with `kindType`. `kindType` is still being used for `reflectlite`, `abi`, and most of the `reflect` casts.

I tried to minimize the diffs for this PR but with things moving around, it was difficult. Most of the stuff in `abi` is very similar to what was in `reflect` and `reflectlite` except with exported identifier names. However, there were some notable changes:

- `abi.Name`, formerly `reflect.name`, used to use a `*byte` to lookup name information in a map. `NameOff` uses an index to lookup a name in a slice of names. To keep from using a global map or slice, I changed `abi.Name` to store the name information directly and `NameOff` to store a pointer to that `Name`.
This does mean we'll need to update `TestLinknameReflectName` since it was linking to `reflect.name` using a `*byte`, but since we have to do that anyway for the move to `abi.Name`, it has to be updated anyway. I left that test update for a later PR.
- `TypeOff` and `TextOff` had similar updates to `Name` and I followed the same pattern as I used for `Name`.
- `reflect.Value.Clear` is new to go1.21 so I implemented it as well.

If #1416 is in merged into go1.21, then the `reflect`, `reflectlite`, and `abi` packages should pass when tests are run on them. However, since go1.21 is an integration branch, it may be difficult to test and CI will continue to fail.

Related to #1415


nevkontakte (approved):
I will be 100% honest and admit that I can't fully wrap my mind around these changes, and so I will put my trust into tests for validating correctness.

It does look reasonable to me from readability/maintainability perspective. The overlap between reflect and reflectlite has been bugging me and moving some commonalities out into `internal/abi` is nice.

PR #1419, opened 2026-03-03
[go1.21] Adding max, min, and clear builtin methods

grantnelson-wf:
Adding [max](https://pkg.go.dev/builtin@go1.21.0#max), [min](https://pkg.go.dev/builtin@go1.21.0#min), and [clear](https://pkg.go.dev/builtin@go1.21.0#clear) builtin methods that are new to go1.21

The tests will run if we comment out `TestLinknameReflectName` and the links it creates since those links broke in the reflect / reflectlite / ABI update. I've put up a PR to remove this test and why.


flimzy (approved):
Taking GopherJS to the MAX one PR at a time!

PR #1424, opened 2026-03-06
[go1.21] Added unsafe.Slice and other fixes

grantnelson-wf:
Changes:

1. adding `unsafe.Slice` since there are some methods in crypto that need it. The overrides that existed around `unsafe.Slice` are still overridden since they also are casting the type of slice which we don't support yet. To make this work I set `$target` when creating an `$indexPtr` to the array, and I add an additional field `$index` that captures the index used in the get/set functions. When creating a slice using an `indexPtr` the index is the offset into the target array

2. added `unsafe.String` and `unsafe.StringData`. Both of these are used with the `unsafe.Slice` and `unsafe.SliceData` in the log/slog package and now that we have both `unsafe.Slice` and `unsafe.SliceData` working, implementing those for strings was way simpler.

3. When casting the returned `*btye` from an `unsafe.StringData` into a `type bp *byte` would drop the `$index` needed to later call `unsafe.String` on that pointer, so I fixed casting to handle casting pointers with `$index` values.

4. When taking a `&s[i]` pointer of an element in a slice, I changed the expression handler to make an `indexPtr` even if the element type is a struct or array. I also added range tests into that code so that it fails with the Go type of out-of-bounds instead of the TypedArray error or just simply returns an undefined for an out-of-bounds. This update to `&s[i]` expressions allows us to use the resulting pointer in `unsafe.Slice` to recover `s`.

5. fixing problem in reflect when minimized causing `toKindTypeExt` to not be found. I forgot to get the minimized name for that call, so it failed to find the unminimized name when minimized

6. disable a race test, `TestOutputRace`. This has the same issue that `go/token.TestFileSetRace` has. Not sure what causes this but since another test is being skipped for the same reason, I'm skipping this one for now too

7. no-op'ed `registerMetric` so that linking it to runtime stops failing. As far as I could tell, we don't collect those metrics and some of them, like the byte size ones, wouldn't work well in JS

Most of CI will fail since this is part of the go1.21 integration work.

### DIFF
--- a/compiler/natives/src/log/log_test.go
+++ b/compiler/natives/src/log/log_test.go
@@ -1,0 +1,9 @@
+//go:build js
+
+package log
+
+import "testing"
+
+func TestOutputRace(t *testing.T) {
+	t.Skip("Fails with: WaitGroup counter not zero")
+}
--- a/compiler/prelude/types.js
+++ b/compiler/prelude/types.js
@@ -210,10 +210,11 @@ var $newType = (size, kind, string, named, pkg, exported, constructor) => {
             break;
 
         case $kindPtr:
-            typ = constructor || function (getter, setter, target) {
+            typ = constructor || function (getter, setter, target, index) {
                 this.$get = getter;
                 this.$set = setter;
                 this.$target = target;
+                if (index !== undefined) this.$index = index;
                 this.$val = this;
             };
             typ.keyFor = $idKey;
@@ -649,17 +650,35 @@ var $newDataPointer = (data, constructor) => {
     return new constructor(() => { return data; }, v => { data = v; });
 };
 
+var $indexPtrGet = function () { return this.$target[this.$index]; };
+var $indexPtrSet = function (v) { this.$target[this.$index] = v; };
 var $indexPtr = (array, index, constructor) => {
+    var makeIndexPtr = () => {
+        if (constructor.elem.kind === $kindStruct) {
+            var ptr = array[index];
+            if (ptr === undefined) {
+                  ptr = array[index] = constructor.elem.zero();
+            }
+            ptr.$val = ptr;
+            ptr.$target = array;
+            ptr.$index = index;
+            ptr.$get = $indexPtrGet;
+            ptr.$set = $indexPtrSet;
+            return ptr
+        }
+        return new constructor($indexPtrGet, $indexPtrSet, array, index);
+    };
+
     if (array.buffer) {
         // Pointers to the same underlying ArrayBuffer share cache.
         var cache = array.buffer.$ptr = array.buffer.$ptr || {};
         // Pointers of different primitive types are non-comparable and stored in different caches.
         var typeCache = cache[array.name] = cache[array.name] || {};
         var cacheIdx = array.BYTES_PER_ELEMENT * index + array.byteOffset;
-        return typeCache[cacheIdx] || (typeCache[cacheIdx] = new constructor(() => { return array[index]; }, v => { array[index] = v; }));
+        return typeCache[cacheIdx] || (typeCache[cacheIdx] = makeIndexPtr());
     } else {
         array.$ptr = array.$ptr || {};
-        return array.$ptr[index] || (array.$ptr[index] = new constructor(() => { return array[index]; }, v => { array[index] = v; }));
+        return array.$ptr[index] || (array.$ptr[index] = makeIndexPtr());
     }
 };
 
@@ -672,6 +691,7 @@ var $sliceType = elem => {
     }
     return typ;
 };
+
 var $makeSlice = (typ, length, capacity = length) => {
     if (length < 0 || length > 2147483647) {
         $throwRuntimeError("makeslice: len out of range");
--- a/compiler/prelude/types.js
+++ b/compiler/prelude/types.js
@@ -654,7 +654,8 @@ var $indexPtrGet = function () { return this.$target[this.$index]; };
 var $indexPtrSet = function (v) { this.$target[this.$index] = v; };
 var $indexPtr = (array, index, constructor) => {
     var makeIndexPtr = () => {
-        if (constructor.elem.kind === $kindStruct) {
+        if (constructor.elem.kind === $kindStruct ||
+            constructor.elem.kind === $kindArray) {
             var ptr = array[index];
             if (ptr === undefined) {
                   ptr = array[index] = constructor.elem.zero();
@@ -663,7 +664,7 @@ var $indexPtr = (array, index, constructor) => {
             ptr.$target = array;
             ptr.$index = index;
             ptr.$get = $indexPtrGet;
-            ptr.$set = $indexPtrSet;
+            ptr.$set = (v) => { constructor.elem.copy(array[index], v); };
             return ptr
         }
         return new constructor($indexPtrGet, $indexPtrSet, array, index);
--- a/tests/js_test.go
+++ b/tests/js_test.go
@@ -1101,6 +1101,12 @@ func TestSliceDataFromPointer(t *testing.T) {
 	})
 }
 
+// TODO(grantnelson-wf): When calling into `unsafe.Slice` the type checker does
+// not understand that `[]T` or `~[]T` is slice. This is a known issue,
+// https://github.com/golang/go/issues/64406 in go1.21. The type checker was
+// fixed in go1.22. Until then the following will only work if gopherjs was
+// built with a go version above go1.21. Once on go1.22, this test can be restored.
+/*
 func TestSliceDataWithDifferentTypes(t *testing.T) {
 	checkSliceDataRoundTrip(t, []bool(nil))
 	checkSliceDataRoundTrip(t, []bool{})
@@ -1166,6 +1172,7 @@ func checkSliceDataRoundTrip[T comparable, S ~[]T](t *testing.T, s S) {
 		})
 	}
 }
+*/
 
 func TestSliceDataEdgeCases(t *testing.T) {
 	catch := func(h func()) (r string) {
--- a/tests/js_test.go
+++ b/tests/js_test.go
@@ -978,39 +978,306 @@ func TestStructWithNonIdentifierJSTag(t *testing.T) {
 }
 
 func TestSliceData(t *testing.T) {
-	var (
-		s0 = []int(nil)
-		s1 = []int{}
-		s2 = []int{1, 2, 3}
-		s3 = s2[1:]
-		s4 = []int{4, 5, 6}
+	t.Run(`nil slice`, func(t *testing.T) {
+		s0 := []int(nil)
+		sd0 := unsafe.SliceData(s0)
+		if sd0 != nil {
+			t.Error(`slice data for nil slice was not nil`)
+		}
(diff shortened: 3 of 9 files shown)


tests/js_test.go
```diff
@@ -978,39 +978,313 @@ func TestStructWithNonIdentifierJSTag(t *testing.T) {
...
+			}
+		}
+	})
+}
+
+// TODO(grantnelson-wf): When calling into `unsafe.Slice` the type checker does
+// not understand that `[]T` or `~[]T` is slice. This is a known issue,
+// https://github.com/golang/go/issues/64406 in go1.21. The type checker was
+// fixed in go1.22. Until then the following will only work if gopherjs was
+// built with a go version above go1.21. Once on go1.22, this test can be restored.
```
grantnelson-wf, line 1108: I wasn't hitting this problem locally because I've been building my gopherjs with `go version go1.26.1 darwin/arm64` since building with go1.21.13 hits the `dyld[90481]: missing LC_UUID load command` on Macs, but CI was hitting this issue in our Smoke tests because they build with go1.21.13.

compiler/prelude/types.js
```diff
@@ -649,17 +650,36 @@ var $newDataPointer = (data, constructor) => {
...
+            }
+            ptr.$val = ptr;
+            ptr.$target = array;
+            ptr.$index = index;
+            ptr.$get = $indexPtrGet;
+            ptr.$set = (v) => { constructor.elem.copy(array[index], v); };
+            return ptr
+        }
+        return new constructor($indexPtrGet, $indexPtrSet, array, index);
+    };
```
grantnelson-wf, line 671: Here's my thought process on this so that you can double check I'm not missing something:

- If we have `s := []int{}` then `p := &s[i]` will be a pointer into `s` at index `i`. We use the default get/set and we add the target `s` and the index `i`. The target and index will allow us to use `unsafe.Slice(p, j)` to  get the slice since the underlying array is target and the index is the offset to start the slice at with `j` being the length of the slice.

- If we have `s := []X` where `type X struct{...}` then `p := &s[i]` is also a `*X` and can access the methods and fields on `X`. So we create a pointer`*X` with the values from `s[i]` then decorate that pointer by adding get/set/target/index. Since the only pointer to that specific `X` in `x[i]` only exists in the slice, the decorations will not conflict with any other pointer. Since the struct in the slice is not a pointer, we want to the set to make a copy. That way we don't have a `var y X`, do a `*p = y`, and have the slice point via JS reference to the same object as `y`, meaning we can safely change `y` without the entry in the slice being modified.

- If we have `s := []*X`, like above but the slice is of pointers then `p := &s[i]` is a `**X`, so we use the default constructor for a pointer with get/set/target/index just like the `[]int` case. The pointer `*X` will not be decorated since we're creating a new pointer to that pointer.

- If we have `s := [][3]X` then `p := &s[i]` we also have a `*[3]X`. Therefore we need to be able to do indexers, `a := p[j]` or `p[j] = a`, and `len(p)`, so we need to create the pointer to `*[3]X` and decorate it with get/set/target/index. With the target/index we can use `unsafe.Slice` to get the slice back still, but we will also be able call the indexers and have them work. Since the `*[3]X` only exists in the slice, the decorations should not affect anything else and we should copy any assignment to the whole slice, e.g. `*p = [3]X{...}`.

compiler/natives/src/log/log_test.go
```diff
@@ -0,0 +1,9 @@
+//go:build js
+
+package log
+
+import "testing"
+
+func TestOutputRace(t *testing.T) {
+	t.Skip("Fails with: WaitGroup counter not zero")
```
nevkontakte, line 8: Maybe add a comment if this is WAI for GopherJS and why?

compiler/prelude/types.js
```diff
@@ -210,10 +210,11 @@ var $newType = (size, kind, string, named, pkg, exported, constructor) => {
             break;
 
         case $kindPtr:
-            typ = constructor || function (getter, setter, target) {
+            typ = constructor || function (getter, setter, target, index) {
                 this.$get = getter;
                 this.$set = setter;
                 this.$target = target;
+                if (index !== undefined) this.$index = index;
```
nevkontakte, line 217: nit: The if check is somewhat redundant. Without the assignment `this.$indes === undefined` anyway. On the other hand, I suppose, this check acts as documentation for optionality of the index.

compiler/prelude/types.js
nevkontakte: I think this logic is correct.

I convinced myself in it by thinking that logically, each Go object belongs to a single, specific place in memory, and thus has a unique pointer. A quirk of GopherJS is that for Go structs, the underlying JS object simultaneously plays a role of the object-value and the pointer to it. I don't remember if this is the case of Go arrays. For an object-pointer to need to have two different indexes, the same logical region of memory must belong to two different slices, which is impossible for a valid Go program state.

However, the said quirk means that the following code would override the fields populated at the time of object creation, at least for a struct:

```js
            ptr.$val = ptr;
            ptr.$target = array;
            ptr.$index = index;
            ptr.$get = $indexPtrGet;
            ptr.$set = (v) => { constructor.elem.copy(array[index], v); };
```

The closes I to a counter-example could come up with is a program like this:

```go
var escapedPtr *Foo

type Foo struct { x int }

func (f *Foo) escape() { escapedPtr = f }

func main() {
    s := make([]Foo, 3)
    s[0].escape() // Receiver pointer is unaware it exists in a slice at the moment, and it escapes to escapedPtr. However, in reality the pointer is the same underlying JS object that holds the struct value.
    println(escapedPtr.x) // Access through the pointer "directly".
    _ := &s[0] // Here the pointer gets "patched" to be aware it exists within a slice. But because it is still the same object, escapedPtr will now become index-aware too.
    println(escapedPtr.x) // And now we access through the index.
}
```

I would ask you to double-check this, but I think this program would still work correctly. The only counter-intuitive aspect of it is that the two print statements will use a different pointer dereferencing methods between them, simply because pointer to a slice element was taken between them. Taking a pointer is usually thought of as a read-only operation, but in this case is changes downstream program behavior.

I can't think of a better way to implement though, any alternatives I can imagine would open a can of pointer equality issues, which is strictly worse :)

nevkontakte (approved):

compiler/prelude/types.js
grantnelson-wf: So, since I don't know JS as well as I wish, I ran some experiments and asked an AI model about it (still doesn't mean I'm right). What I was thinking is that even though `$index` is undefined it would still show up if/when iterating over the object. That should be fine since we don't iterate over the indexPtr like that, but it means I would be adding additional memory into objects that don't need an `$index`.

```JS
const o = { };
o.cat = 9;
o.dog = undefined;
for (const [key, value] of Object.entries(o))
  console.log(key, value);
// Outputs:
// "cat", 9
// "dog", undefined
```

But TBH, I'm not sure if that's the same when in a constructor function using `this`. That's just my sad attempt at not adding too much overhead whilst adding a little overhead where we need it. Its probably negligible in the end. And, you have a good point, it does also work like documentation to show that it is optional (I hadn't thought of that).

That's also why I added the default get/set functions, since I think that may reduce the amount of closures created and therefore the overall runtime memory. But again, I'm horrible at JS so I'm not sure if it's worth it to try to minimize memory usage like that.

compiler/prelude/types.js
grantnelson-wf: Wow, that broke my brain. I'm going to have to play with that for a bit so I can make sure I'm getting what's happening in the Go memory and in the JS representation. What you said sounds correct and like it might be an issue. Fortunately this is going into go1.21 branch so we can continue to tinker with it before it goes into master. (Which is also good since I found a few other things that need tinkering such as there are two patterns using `%e` instead of `%f` for the index so until I get that PR up, some of these won't work for `int64`.)

I'll keep thinking about this since you bring up a good point, but right now I'm not sure what we could do.

compiler/natives/src/log/log_test.go
grantnelson-wf: Can do. There is another test `TestFileSetRace` in go/tokens that is skipped for the same reason so I'll update both.

I had to refigure this one out just now. It has to do with both "go/token" and "log" being updated in [`augmentOriginalImports`](https://github.com/gopherjs/gopherjs/blob/master/build/build.go#L331-L344) to use [`nosync`](https://github.com/gopherjs/gopherjs/blob/master/nosync/mutex.go#L90-L95) instead of [`sync`](https://github.com/gopherjs/gopherjs/blob/master/compiler/natives/src/sync/waitgroup.go#L27-L31). Since those two tests are in those packages, they fail with the nosync "Wait" error.

[Here is where the nosync code was originally added](https://github.com/gopherjs/gopherjs/commit/ac5f8a84b1bb5392d20a6bf1d671a7174fb539fe#diff-7fdd319c657fbde54b83a5dfa233c5c2c417b291739ff75c05dc5d83ee7a9d24R157-R160) but it didn't have "go/token" and "log". "go/token" was added [here](https://github.com/gopherjs/gopherjs/commit/2e90d8878e451ca810dda047c2ae2e223cc8ae5e) and "log" was added [here](https://github.com/gopherjs/gopherjs/commit/08a172ac0e41797a71b499e9bed728be3162d26d). However, there isn't any explanation why those needed to use `nosync`, I just assume it's how those packages are implemented by GopherJS.

I'll fix this in a follow PR and I'll resolve this conversation for now so that it doesn't block this PR.

compiler/prelude/types.js
grantnelson-wf: I'm going to resolve this conversation so it doesn't block this PR but we can totally rework this anyway that you think would be best.

compiler/natives/src/log/log_test.go
grantnelson-wf: Comments were added to https://github.com/gopherjs/gopherjs/pull/1433

compiler/prelude/types.js
grantnelson-wf: Since we can continue to work on this potential issue, I'm going to resolve this conversation to unblock this PR, and add this comment to my backlog so it isn't forgotten.

As far as I could figure out so far, it seems like the decorations to the pointer should still be stable since that pointer can't exist in two arrays as a non-pointer struct. So we shouldn't have an issue with one index and target overriding another. However, that does make some quirkiness around when a pointer can be used to create a slice or not in GopherJS, but right now I'm not sure how we could work around that.

compiler/natives/src/log/log_test.go
nevkontakte: `nosync` has been a mystery for a while for me too. I think eventually I managed to dig out that it was a performance optimization: due to JS's single-threaded nature and the fact that GopherJS goroutines yield only at pre-determined points, the synchronization in those packages was unnecessary overhead and `nosync` was supposed to fix that. I haven't experimented on how much performance it actually buys us these days.

compiler/prelude/types.js
nevkontakte: You are right, it would show up if one was to iterage over the keys. I don't think that would particularly matter in this case, but it's not a big deal either way.

Like I said, it was just a nitpick on my part, and so you are at liberty to disregard it :)

PR #1429, opened 2026-04-20
Cache imports to improve performance

cadencefinch-wf:
We have been looking into allowing our project to exist outside of a GOPATH and we've noticed that GopherJS compile times are much slower in that case.

After some investigation, it appears that resolving imports simply takes a longer time when using Go modules. While there's not a lot that can be done about that root issue, we can establish a caching system that can help avoid resolving imports for the same thing multiple times.

In our case, which is admittedly a sizable code base, compiling our GopherJS code takes around 40 seconds in GOPATH, a little bit over 8 minutes outside of GOPATH without this change and roughly a minute outside of GOPATH with this change. This change does not seem to make much of an impact on compile times in GoPath because those imports are already fast enough that the caching doesn't provide a significant boost.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -772,6 +772,10 @@ type Session struct {
 	// is the unresolved import path and source directory.
 	importPaths map[string]map[string]string
 
+	// packages caches imported package metadata by resolved import path.
+	// This avoids repeated xctx.Import calls for the same package.
+	packages map[string]*PackageData
+
 	// sources is a map of parsed packages that have been built and augmented.
 	// This is keyed using resolved import paths. This is used to avoid
 	// rebuilding and augmenting packages that are imported by several packages.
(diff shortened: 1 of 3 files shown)


grantnelson-wf (approved):

nevkontakte (approved):
The slowdown in the Modules model is very likely because go/build (and .../go/packages) invoke `go list` under the hood to deal with module resolution. It's an unfortunate, in my opinion, choice on the part of the Go team, but I believe their goal was to couple module resolution logic to the installed toolchain version rather than whatever toolchain the calling library was compiled with.

Cacing package imports would spare a fair bunch of subprocessing and parsing, although 8x slowdown is quite surprising. Perhaps there's some non-linear cost dependent on the codebase size.

PR #1433, opened 2026-04-23
[go1.21] Some more updates to native overrides for go1.21

grantnelson-wf:
This contains some more changes to native overrides:
- Adding some of the updates to fix or skip tests in the native overrides
- Fixing some unsupported `//go:linkname` links
- Removing cast to an `unsafeheader.String`

This also contains the update to add comments to `TestOutputRace` as [discussed here](https://github.com/gopherjs/gopherjs/pull/1424#discussion_r3120424258)

### DIFF
--- a/compiler/natives/src/bytes/bytes_js_wasm_test.go
+++ b/compiler/natives/src/bytes/bytes_js_wasm_test.go
@@ -1,0 +1,15 @@
+//go:build js && wasm
+
+package bytes_test
+
+import (
+	"testing"
+)
+
+//gopherjs:replace
+func TestIssue65571(t *testing.T) {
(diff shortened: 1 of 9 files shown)


compiler/natives/src/bytes/bytes_js_wasm_test.go
```diff
@@ -0,0 +1,15 @@
+//go:build js && wasm
```
grantnelson-wf, line 1: This build constraint has `&& wasm` because the original test is behind the same constraints.

nevkontakte: We also build standard library with the GOARCH=wasm, so that constraint is satisfied.

nevkontakte (approved):

PR #1434, opened 2026-04-23
[go1.21] Fixed int64/uint64 indexes when creating index pointers

grantnelson-wf:
When I was working on the update to indexPtrs for unsafe.Slice, I used a `%e` instead of a `%f` by mistake. This fixes that mistake and adds a test to check the fix.

### DIFF
--- a/tests/js_test.go
+++ b/tests/js_test.go
@@ -1288,3 +1288,46 @@ func TestStringData(t *testing.T) {
 		}
 	})
 }
+
+func TestPointerToSliceIndex64(t *testing.T) {
+	// This test repeated an issue when using an int64 index in an pointer to a
+	// slice or array element causing the element to be resolved to the zero value
+	// of the structure. Since int64 and uint64 are stored as an object with a
+	// high/low value, we should use `%f` when creating the pattern for the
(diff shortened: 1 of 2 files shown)


tests/js_test.go
```diff
@@ -1288,3 +1288,46 @@ func TestStringData(t *testing.T) {
 		}
 	})
 }
+
+func TestPointerToSliceIndex64(t *testing.T) {
+	// This test repeated an issue when using an int64 index in an pointer to a
+	// slice or array element causing the element to be resolved to the zero value
+	// of the structure. Since int64 and uint64 are stored as an object with a
+	// high/low value, we should use `%f` when creating the pattern for the
```
nevkontakte, line 1296: This gave me a momentary pause because %f coverts int64 into a JS number, which loses precision for large values. However, in this context this is fine, because we don't allow arrays that large, and therefor there shouldn't be an index that is more than 52 bits. Maybe add that to the comment for the posterity?

nevkontakte (approved):

tests/js_test.go
grantnelson-wf: Makes sense, I'll add that to a comment in a following PR.

Issue #843, opened 2018-07-17
Typecasting nil results in non-nil


seanburke-wf:
If a variable has value `nil` and is typecast, the result will have a non-nil value which appears to be an invalid pointer. This differs from the behavior of Go.

The following illustrate the difference in behaviors between Go and compiled gopherjs:
https://play.golang.org/p/_feur6S2EjC
https://gopherjs.github.io/playground/#/y785Hlp5q0

Additionally, the non-nil value appears to be an invalid reference:
https://play.golang.org/p/ugzWXe4x2HE
https://gopherjs.github.io/playground/#/HkVf0Bqs1m

flimzy:
Nice find.

theclapp:
Quick-and-possibly-dirty solution that may be a real fix, or at least may help point the way towards one:

```diff
diff --git a/compiler/prelude/types.go b/compiler/prelude/types.go
index a8c29d5..7e81b22 100644
--- a/compiler/prelude/types.go
+++ b/compiler/prelude/types.go
@@ -273,6 +273,7 @@ var $newType = function(size, kind, string, named, pkg, exported, constructor) {
       });
       typ.ptr.nil = Object.create(constructor.prototype, properties);
       typ.ptr.nil.$val = typ.ptr.nil;
+      typ.ptr.nil.isNil = true;
       /* methods for embedded fields */
       $addMethodSynthesizer(function() {
         var synthesizeMethod = function(target, m, f) {

diff --git a/compiler/prelude/prelude.go b/compiler/prelude/prelude.go
index 1eed6c4..2c754ec 100644
--- a/compiler/prelude/prelude.go
+++ b/compiler/prelude/prelude.go
@@ -317,6 +317,9 @@ var $pointerOfStructConversion = function(obj, type) {
     obj.$proxies[type.string] = proxy;
     proxy.$proxies = obj.$proxies;
   }
+  if (obj.isNil) {
+    return type.nil;
+  }
   return proxy;
 };
```

Test case:

```go
var foo *FooType = nil
fmt.Println("foo is nil?", foo == nil)
var bar *BarType = (*BarType)(foo)
fmt.Println("bar is nil?", bar == nil)

foo = nil
fmt.Printf("foo is %v\n", foo)

bar = (*BarType)(foo)
fmt.Printf("bar is %v\n", bar)
```
=>
```
foo is nil? true
bar is nil? true
foo is <nil>
bar is <nil>
```

seanburke-wf:
[...]

PR #1435, opened 2026-04-23
[go1.21] Adding more nil-aware checks

grantnelson-wf:
Changes:
- Fixes the issue https://github.com/gopherjs/gopherjs/issues/843 about nil casts
- Fixes several places where a nil dereference should panic
- Adds the go1.21 update for `panic(nil)` to return [`PanicNilError`](https://pkg.go.dev/runtime#PanicNilError) from recover. (This does not add the disable using `GODEBUG=panicnil=1` since I wasn't sure if we wanted to support that nor how to work that into the prelude smoothly without simply adding a flag that gets set in JS to skip the `if (value === $ifaceNil) {` statement.)
- Several of the gorepo tests got fixed with this change so I updated the knownFails for gorepo
- The change to `goroutine` offset the expected result in the jsSourceMap test, so I updated that too
- Lastly, I realized that tests for `internal/abi` weren't being run so I fixed that (even though that isn't directly related to nil aware things).

I tried to think of all the different ways through the expressions part of the compiler that would need a nil check, but if anyone notices any that I missed, we definitely should add them to the tests and fix them too.

### DIFF
--- a/compiler/expressions.go
+++ b/compiler/expressions.go
@@ -1329,7 +1329,7 @@ func (fc *funcContext) translateConversion(expr ast.Expr, desiredType types.Type
 			//
 			// TODO(nevkontakte): Should this only apply when exprType is a pointer to a
 			// struct as well?
-			return fc.formatExpr("(%1e === %2s.nil ? %3s.nil : $pointerOfStructConversion(%1e, %3s))", expr, fc.typeName(exprType), fc.typeName(desiredType))
+			return fc.formatExpr("$pointerOfStructConversion(%e, %s)", expr, fc.typeName(desiredType))
 		}
 
 		if types.Identical(exprType, types.Typ[types.UnsafePointer]) {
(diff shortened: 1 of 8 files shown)


compiler/expressions.go
```diff
@@ -1320,7 +1329,7 @@ func (fc *funcContext) translateConversion(expr ast.Expr, desiredType types.Type
 			//
 			// TODO(nevkontakte): Should this only apply when exprType is a pointer to a
 			// struct as well?
-			return fc.formatExpr("$pointerOfStructConversion(%e, %s)", expr, fc.typeName(desiredType))
+			return fc.formatExpr("(%1e === %2s.nil ? %3s.nil : $pointerOfStructConversion(%1e, %3s))", expr, fc.typeName(exprType), fc.typeName(desiredType))
```
nevkontakte: nit: To make generated code to be a little more compact, the nil check could be moved into `$pointerOfStructConversion` itself.

nevkontakte (approved):

compiler/expressions.go
grantnelson-wf: I moved that check. I tried to do it without passing the type into it since `$pointerOfStructConversion` is also called from `$unsafeMethodToFunction`. You probably should check that I'm doing the nil check correctly that way. All the tests that were passing seem to still be passing so it seems to work.

nevkontakte (approved):

compiler/expressions.go
nevkontakte: This looks correct to me.

PR #1436, opened 2026-05-01
[go1.21] Setting testing.Testing to true when testing

grantnelson-wf:
Go1.21 added [`testing.Testing`](https://pkg.go.dev/testing#Testing) that returns true if the code was built with `go build`. To make this work for `gopherjs build` I added a native overlay that is hardcoded in `build/build.go` that will be added into the overlay files when building the `testing` package iff the program is being built for a test.

I also bumped timeout-minutes to 45 mins since 30 mins sometimes failed when other stuff was failing and I'd rather see all the failures not just a few.

I also added a gitignore for the AI agent files that the AI agent I'm using wrote. If we want to add agent files in we can always remove these ignores, I just didn't want to flood our repo with a bunch of files.

related to https://github.com/gopherjs/gopherjs/issues/1415

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -183,7 +183,7 @@ jobs:
   gopherjs_tests:
     name: GopherJS Tests (${{ matrix.filter.name }})
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 45
     strategy:
       fail-fast: false
       matrix:
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,7 @@
 /node-syscall/build
 /node_modules
+
+# Folders and files used by AI agents
+.agents/
+.cursor/
+AGENTS.md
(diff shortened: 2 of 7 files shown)


nevkontakte:
I have a slight apprehension about a "magic" overlay that is added to the build in some cases, but not others.

An alternative approach would be to replicate what we've done with [runtime.Version()](https://github.com/gopherjs/gopherjs/blob/204ca8811823075300cb4ad4f2b9643a8f94d558/compiler/natives/src/runtime/runtime.go#L461): when compiler generates the code it creates a [top-level $goVersion variable](https://github.com/gopherjs/gopherjs/blob/204ca8811823075300cb4ad4f2b9643a8f94d558/compiler/compiler.go#L160), and runtime package [reads its value](https://github.com/gopherjs/gopherjs/blob/204ca8811823075300cb4ad4f2b9643a8f94d558/compiler/natives/src/runtime/runtime.go#L73) at init time.

I think we could generalize this approach to serve as a more versatile way to pass build-time labels to the program.

WDYT?

grantnelson-wf:
That sounds like a great idea. I had looked for something similar and didn't notice `runtime.Version()` so just came up with my own way. It would be much better to follow the existing pattern. I will update this PR

Edit: I updated this PR to follow the pattern used by `runtime.Version()`.

flimzy (approved):

PR #1440, opened 2026-05-01
[Closed] Fixed identifier collisions in minified code

grantnelson-wf:
⚠️  It just occurred to me that the prior "fix" I made to master since this is more related to a bug in generics than go1.21, even though it wasn't found until a test in go1.21 hit the issue. I'll recreate this PR to target master next week.

------

The bug found in generics that I fixed recently missed an edge case that could cause an anonymous pointer type to get the same name as a local variable when the output is minified. This was found in the tests in the `slices` package.

The problem was that when the first generics instantiation was created it will also create the package level pointer types. As a side effect of creating the pointer types, the `allVars` is updated. Then in the second instantiation, the pointer type is already found at the package level and so was not being added into the `allVars` meaning the pointer type's identifier could then be used by a local variable causing a collision.

I undid the prior "fix" since that wouldn't fix the whole issue. Now we have a package level `varPtrNames` and a function level `varPtrNames`. The pointer type is then not always treated as package level when it is only scoped to a function. This means when it is package level, it will not collide because of other naming rules, but when the pointer type is local to a generic function, each instantiation will have a new name given to it each time, thus avoiding the collision.

### DIFF
--- a/compiler/functions.go
+++ b/compiler/functions.go
@@ -38,6 +38,7 @@ func (fc *funcContext) nestedFunctionContext(info *analysis.FuncInfo, inst typep
 		parent:       fc,
 		allVars:      make(map[string]int, len(fc.allVars)),
 		localVars:    []string{},
+		varPtrNames:  make(map[*types.Var]string),
 		flowDatas:    map[*types.Label]*flowData{nil: {}},
 		caseCounter:  1,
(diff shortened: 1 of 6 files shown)


grantnelson-wf:
This was moved to https://github.com/gopherjs/gopherjs/pull/1442 where it could properly target master.

PR #1446, opened 2026-05-13
[go1.21] Optimizing call frames

grantnelson-wf:
One of the developers at Workiva, @cadencefinch-wf, reported that while she was improving the CI run time for one of our repos that she noticed `t.Helper()` was slow once GopherJS'ed. This was amplified when using `b.Helper()` in a benchmark. I investigated and determined that there was a way to improve the speed of getting the call frames.

(some what) related to https://github.com/gopherjs/gopherjs/issues/1415. This isn't specific to go1.21 but it will conflict with several changes already put into go1.21 if it was to be added to master first.

### Summary

See [comment below](https://github.com/gopherjs/gopherjs/pull/1446#issuecomment-4614799522) for more detail

The biggest speed boost was setting [`Error.stackTraceLimit`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stackTraceLimit) before reading a limited amount of the call frames. If `Error.stackTraceLimit` isn't being used (since it is "non-standard") setting it has no positive nor negative affect.

Less of a boost was to move `parseCallFrame` into JS. This allowed me to create the regexp in JS the first time they are used and it allowed the tests for `parseCallFrame` to access it without having to export `ParseCallFrame` from the runtime package. I also updated the regexp to allow for unspecified column and line numbers.

I was able to improve the time of `registerPosition` by having it key off of the comparable `basicFrame` without having to create a key as a string.

### Results

I added 2 benchmarks. One for testing `t.Helper()` and one for `Callers`. Someone else on some other machine at sometime should duplicate the benchmark test to check my work.

As can be seen by the benchmarks:
- `t.Helper()` now takes about 1/2 to 1/3 the amount of time that it used to. Some of this is contributed to the improvements to `registerPosition`.
- `Callers` roughly no change (100%) to 30% of the original time but it is very dependent on how many lines are parsed. `Callers` also checked how "skip" affected the results and it was negligible.


compiler/natives/src/runtime/runtime.go
```diff
@@ -164,9 +156,33 @@ type basicFrame struct {
...
 
 func callstack(skip, limit int) []basicFrame {
-	skip = skip + 1 /*skip error message*/ + 1 /*skip callstack's own frame*/
-	lines := js.Global.Get("Error").New().Get("stack").Call("split", "\n").Call("slice", skip, skip+limit)
-	return parseCallstack(lines)
+	// $callstack in prelude.js returns an iterator function.
+	// The returned stack reader may be able to return more than limit number
+	// of frames so only read as many as needed.
+	// skip+1 skips callstack's own call frame.
+	lines := js.Global.Call("$callstack", skip+1, limit)
```
nevkontakte: Assuming the runtime supports the limit, what happens if some of the raw frames are hidden? Unless I'm missing something, `callstack()` would return fewer frames than expected, which probably isn't correct.

compiler/prelude/prelude.js
```diff
@@ -89,6 +89,122 @@ if (($global.process !== undefined) && $global.require) {
 }
 var $println = console.log
 
+// $callstack captures a stack trace and returns a list of lines, typically
+// there will be "limit" numbers of lines returned.
+// skip=0 means "the direct caller of $callstack"; limit caps frames captured.
+var $callstack = (() => {
```
nevkontakte: I'm curious, is the performance worse if we impement exactly the same logic, but in Go using JS APIs? What makes the code slow?

My only guess is that the function is assumed to be blocking and has to deal with the added inefficiency of handling that, but I don't see anything in here that should make it all blocking.

nevkontakte (commented):
I'm not opposed to this change, but I would like to understand why the same peformance can't be achieved with the code output by the GopherJS compiler. For example, if recompiling regexps is expensive in a tight loop (which I can believe), we could move the initialization into `init()`.

compiler/natives/src/runtime/runtime.go
grantnelson-wf: In the original code the lines in `callstack` were reduced with `.Call("slice", skip, skip+limit)` that would cause the output to be less than limit if the hidden frame was in that stack. I added a test to `TestCallers` and confirmed this was the original behavor.

We can improve this by adding more frames but it slows down the stack trace as seen in the [comment below](https://github.com/gopherjs/gopherjs/pull/1446#issuecomment-4614799522). For right now I'm going to keep it as it was originally but we can change it if we want to.

compiler/prelude/prelude.js
grantnelson-wf: It felt faster when written in JS but it was not very noticeable with the other changes. Things like `stack.Call("indexOf", "\n").Int()` becomes `$parseInt(stack.indexOf($externalize("\n", $String))) >> 0;` instead of simply `stack.indexOf("\n")`. But since most of the improvement in speed was from limiting the number of frames, this was hard to measure (without me purposefully measuring just the parse code like did below in [this comment](https://github.com/gopherjs/gopherjs/pull/1446#issuecomment-4614799522)).

The main reason I made the change was that there was nothing in those methods that wasn't *js.Object calls (or at least something that could be JS) so it felt like it could have been moved over to JS and was hoping for improved speed (even though it didn't seem like there was much). Also, being in JS allows the tests package to access `parseCallFrame` without it having to be exported from runtime.

After testing just the parse code in JS and in Go, it was slightly faster in JS, so it feels like it would be better to keep it in JS.

grantnelson-wf:
In my response, I found myself saying that something was hard to measure since it was small compared to the overall change. That's not great, so I'll go back and break out most of the parts into a benchmark and test each separately. I'll rework this

### Error.stack vs Error.captureStackTrace

I tested `new Error().stack` against `Error.captureStackTrace(target)` and `Error.captureStackTrace(target, func)`. I found that they were all equivalent with `captureStackTrace(target, func)` insignificantly better. I ran a 2-way ANOVA on the results and failed to reject the $H_0$ that they were the same. On average, when requesting X number of frames, each frame takes about 2,300 ns (2.3 $\micro{s}$ ) to get from JS. I assume this is because each frame has to be looked up with the source map.

Using `func` in `captureStackTrace` allows us to cut off a frame (saving us almost as much as a frame per call). That's kind of useful but not enough and there is no easy way to determine which `func` to pick when we are skipping.

Since none were significantly faster, I'll go with the most supported one `new Error().stack`.

### ParseCallFrame in Go vs JS

Having ParseCallFrame written in JS is always slightly faster. JS takes 86.3% of the time of Go on average. This is likely due to strings and integers having to be externalized (e.g. `$externalize("\n", $String)`, `skip = skip + (1) >> 0;`). So I'm going to move ParseCallFrame into JS.

### Preprocessing Stack in Go vs JS

The preprocessing is getting the `new Error().stack`, checking for the `Error` header line, then splitting into lines.
There was no significant difference (+/-1% variance). Leaving in JS.

### Split/Join, Regex, vs LastIndex for parsing positions

The code for parsing positions into `[file, line, col]` tuples could be broken up in different ways. The Regex is fastest, the other two are slightly slower (+1%) than regex. Since the regex also protects against malformed positions with no lines or columns, I'm going with the Regex version.

### Exit parse for hidden frames vs checking after parse

The function name is parsed out of the frame first, so we could check if it is the hidden frame before parsing the position information and return an empty array. Or wait until the parse is done, then check function name. Both of these are negligible. This is probably because the hidden frame isn't in the benchmark and it isn't hit often so this is really only checking the overhead. I assume checking during parsing is faster when parsing the actual hidden frame
but that isn't very common (without a bunch of defers inside each other). I'm going to leave it until after parsing since that is closer to what we had originally.

### Setting limit, setting higher limit, vs not setting limit

When a limit is requested we can set the `Error.stackTraceLimit` to either that limit, to the limit plus a constant, or just not set the limit. Setting the limit was the biggest help in improving the speed of `t.Helper()`. Here are the results from setting the limit (tight of limit possible), the limit + 1, the limit + 2, and not setting the limit. When comparing against setting the tight limit: limit +1 takes 10% more time, limit + 2 takes 20% more time, and no limit takes 150% more time (limited by the depth of the stack in the test). This uses the `Benchmark_TestingHelper` benchmark for testing.

image

We can see similar results using `Benchmark_Callers`, however, the larger the request limit gets the parse time starts becoming relatively significant so the difference between setting the +c becomes less pronounced however not setting the limit is always worse. This makes sense with what we found earlier - each frame takes about 2,300 ns (2.3 $\micro{s}$ ) to get from JS.

This shows that we should set as tight of limit (`Error.stackTraceLimit`) as possible to the requested limit.

compiler/natives/src/runtime/runtime.go
nevkontakte: Sounds reasonable.

compiler/prelude/prelude.js
nevkontakte: Ok, that makes sense. I feel like there is a compiler improvement opportunity in there somewhere, but it would be far out of scope for this PR.

nevkontakte (approved):

Issue #1003, opened 2021-03-21
Type assertions don't respect method shadowing with multiple embedded types


nevkontakte:
Reproduction example:

```
package main

type IOne interface {
	conflict()
}

type ITwo interface {
	conflict()
}

type Struct struct {
	IOne
	ITwo
}

func main() {
	s := &Struct{}
	// _ = IOne(s) // Compile-time: cannot convert s (variable of type *Struct) to IOne.
	iface := interface{}(s)
	_, ok := iface.(IOne)
	println("Type assertion successful:", ok)
}
```

Execution:
```
 $ gopherjs run main.go
Type assertion successful: true
$ go run main.go
Type assertion successful: false
```

The relevant part of the [spec](https://golang.org/ref/spec#Selectors):

> For a value x of type T or *T where T is not a pointer or interface type, x.f denotes the field or method at the shallowest depth in T where there is such an f. If there is not exactly one f with shallowest depth, the selector expression is illegal.

There is a related issue https://github.com/gopherjs/gopherjs/issues/757, which concerns struct's field shadowing an embedded method. Technically that case is different, since the conflicting field and method have different depths.

grantnelson-wf:
This should be fixed in master now

nevkontakte:
Thanks for the fix! Closing this as resolved.

Issue #757, opened 2018-02-19
Broken type assertion with structs embedding


damz:
The `$assertType` helper only considers the method set when asserting if a type implement an interface. But with struct embedding, it is possible for a struct field to shadow a method.

This fails `gopherjs run`:

```
package main

type Doer interface {
	Do()
}

type Embedded struct{}

func (e *Embedded) Do() {}

type A struct {
	Embedded
	Do string
}

func main() {
	var a interface{}
	a = &A{}
	if aa, ok := a.(Doer); ok {
		aa.Do()
	}
}
```

with:

```
TypeError: aa.Do is not a function
```

theclapp:
If anyone else was confused (as I was :), the point is not that aa.Do() should call A.Embedded.Do(), the point is that the type assertion should've failed: `a` is not actually a Doer:

```go
        // In the regular Go playground
	if aa, ok := a.(Doer); ok {
		aa.Do()
	} else {
		println("not a doer")
	}
// => not a doer
```

damz:
@theclapp That's correct, I updated to the description in an effort to make that clearer.

damz:
So, is `$assertType` wrong, and needs to look into properties of the type too, or should this method not be in the methodset of the type to begin with?

grantnelson-wf:
This should be fixed in master now

nevkontakte:
Closing as resolved, thanks for the fix!

PR #1448, opened 2026-05-18
Fixing shadowing and ambiguity in method set

grantnelson-wf:
Fixing https://github.com/gopherjs/gopherjs/issues/757 and https://github.com/gopherjs/gopherjs/issues/1003

The `$methodSet` function in `compiler/prelude/types.js` had some missing edge cases:
 - Not promoting methods if more than one method or a field and method in embedded types have the same identifier, i.e. the method is ambiguous so can not be promoted.
 As seen in https://github.com/gopherjs/gopherjs/issues/1003
 - Shadowing embedded types so that they don't add methods for fields that exist in the structures that are embedding that type.
 As seen in https://github.com/gopherjs/gopherjs/issues/757

The embedded sets were already grouped together such that all the embedded type of one type are checked together. That is great and helped make the fix easier. To fix these, I added a `shadow` set of names that are shadowed when going into the next embedded set. I also made `mset` entries get set to `null` when there is an ambiguity or a field. Then when setting methods to the type, the `null` ones are filtered out.

I found this while working on go1.21, but since it was not directly related to go1.21, I got this ready for master.

### DIFF
--- a/compiler/prelude/types.js
+++ b/compiler/prelude/types.js
@@ -418,13 +418,13 @@ var $methodSet = typ => {
         return [];
     }
 
-    var current = [{ typ: isPtr ? typ.elem : typ, indirect: isPtr }];
+    var current = [{ typ: isPtr ? typ.elem : typ, indirect: isPtr, shadow: undefined }];
 
     var seen = {};
 
     while (current.length > 0) {
         var next = [];
-        var mset = [];
+        var mset = {};
 
         current.forEach(e => {
             if (seen[e.typ.string]) {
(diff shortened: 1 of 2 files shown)


grantnelson-wf:
~~The tests passed on my machine.~~ I'll figure out why they aren't passing in CI.

I think it's because those specific failures are fixed by one of the updates to casting in go1.21. I'm going to check it out and may recreate this PR for the go1.21 integration branch.

grantnelson-wf:
I'm not sure what's going on with this. I'm going to rework this an create a new PR.

Sorry for the noise in your inbox.

Issue #1328, opened 2024-07-20
Proposal: remove support for the node_syscall extension


nevkontakte:
The extension has been deprecated for about 3 years (since https://github.com/gopherjs/gopherjs/pull/1111), and has been disabled by default since then. So far I haven't heard of anyone asking to continue supporting it. This extension has known issues in the memory management department and doesn't support all types of arguments (e.g. https://github.com/gopherjs/gopherjs/issues/993). I am not even sure it builds under the modern node versions.

I would like to do the following:

 1. Move the source code into a separate repository under the gopherjs organization and mark it archived. This will keep sources available in case anyone is using it for some other purpose, and it would clearly indicate its status an unmaintained.
 2. Remove `legacy_syscall` build tag from the code base.
 3. Remove the option of using different GOOS/GOARCH for user code which was added as a compatibility option in #1111 and fix them to GOOS=js GOARCH=ecmascript.

Inviting @flimzy and anyone else to comment.

flimzy:
I'm in principle in favor of the proposal. I'm only concerned that the fact that we haven't done a complete release in quite some time may mean that there may be those using it, who aren't aware yet of the changes.

As such, I wonder if for the next major release (when generics are complete) we could move it to its own repo first, which will make the deprecation status more obvious, and likely bring out of the woodwork anyone who really needs it.  Then assuming no major objections, do the rest of the removal for the following major release.

nevkontakte:
Yeah, that's a fair argument. I'll admit I am a bit overeager to get rid of it because [debugging mystery sigfaults](https://github.com/gopherjs/gopherjs/pull/1329/commits/b579c8f75db49a5668877659ef34e1a72c5df798) was not fun 😅 But making a full release before deprecating it entirely does seem like a responsible way of doing things.

PR #1450, opened 2026-05-20
[go1.21] Expunge node_syscall

grantnelson-wf:
node_syscall has been deprecated for a while. I'm adding this work to the go1.21 integration branch so that when GopherJS supports go1.21, it will no longer support legacy node_syscall.

CI may not pass because of fixes in other PRs that haven't been added to the go1.21 integration branch yet.

### Issues related to this work

- https://github.com/gopherjs/gopherjs/issues/993
- https://github.com/gopherjs/gopherjs/issues/1328
- https://github.com/gopherjs/gopherjs/issues/1415

### Slack discussion

image


package.json
nevkontakte, line 1: I think at this point we can remove package.json entirely, as well as the `npm install` step from CI.

nevkontakte (approved):

package.json
grantnelson-wf: Sounds good. I will create a follow up for this work

Issue #1334, opened 2024-07-25
Add Go 1.22 Support?


rcoreilly:
What are the current plans for implementing the rest of the features to get up to go 1.22? 

We are now appreciating in our [Cogent Core](https://github.com/cogentcore/core) framework the significant limitations of the wasm platform (e.g., https://github.com/WebAssembly/design/issues/796#issuecomment-401310366) and performance on mobile web in particular is unacceptable for any kind of more complex app (e.g., https://github.com/golang/go/issues/68156 and other issues linked therein).  Therefore, we're wondering if gopherjs might provide a more performant web experience overall.  However, our code depends on all the 1.22 goodies, including loop var and range over int, min/max, slog, and generics.  We did some quick tests and it looks like the current generics branch could handle the generics aspects.

I have reasonable experience with the go compiler code and recently implemented these same features in [yaegi](https://github.com/traefik/yaegi) as well, so I could probably do them relatively quickly if that would be a welcome PR?

flimzy:
Go 1.20 support is currently [under active development](https://github.com/gopherjs/gopherjs/issues/1270), and nearing completion, with the exception of Generics support, which is simultaneously [under active development](https://github.com/gopherjs/gopherjs/issues/1013).

Once these two objectives are complete, we'll be ready to work on Go 1.21, followed by 1.20.

If you have the knowledge and time to contribute to either of these efforts, that would be welcome.

rcoreilly:
Quick update here: I took a look at the code and your nice documentation about what is required to update to a new Go version, and quickly realized that it is quite a big task to get to a new version, and the updates for the loopvar syntax are a relatively minor aspect of that task. I did manage to implement a quick "hack" implementation of loopvar in the existing codebase and it seemed to work, so I'd be happy to contribute that once the rest of the heavy lifting happens to actually get to the 1.22 version.

If I somehow end up finishing all my other todo items before you get to 1.22, maybe I'll try to help out with wherever you are, but I suspect the collective expertise here makes you all much more efficient at that than I would be.  And wow, javascript is kind of a nightmare! :)

grantnelson-wf:
Currently we are finishing getting go1.21.13 support into the go1.21 integration branch, then tested, then merged to master, and tagged. Since some of that work requires me to wait for reviews, I'm going to start working on go1.22.12 support. However, until go1.21 support is finished, go1.21 support has higher priority than go1.22 support.

grantnelson-wf:
[...]

PR #1459, opened 2026-07-21
[go1.22] Initial version bump for go1.22 integration branch

grantnelson-wf:
Starting the go1.22 integration branch by bumping the go versions. This branch's CI will fail since we don't support go1.22 yet.

related to https://github.com/gopherjs/gopherjs/issues/1334

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ GopherJS compiles Go code ([go.dev](https://go.dev/)) to pure JavaScript code. I
 
 ### What's new?
 
+- 2026-06-26: Go 1.21 support is [released](https://github.com/gopherjs/gopherjs/releases/tag/v1.21.0)!
+- 2026-01-02: Go 1.20 support is [released](https://github.com/gopherjs/gopherjs/releases/tag/v1.20.0)!
 - 2025-08-19: Go 1.19 beta2 is [released](https://github.com/gopherjs/gopherjs/releases/tag/v1.19.0-beta2), with full generics support!
 - 2024-02-24: Go 1.19 support is [available](https://github.com/gopherjs/gopherjs/releases/tag/v1.19.0-beta1)!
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -15,6 +15,8 @@ GopherJS compiles Go code ([go.dev](https://go.dev/)) to pure JavaScript code. I
 
 ### What's new?
 
+- 2026-06-26: Go 1.21 support is [released](https://github.com/gopherjs/gopherjs/releases/tag/v1.21.0)!
+- 2026-01-02: Go 1.20 support is [released](https://github.com/gopherjs/gopherjs/releases/tag/v1.20.0)!
```
flimzy, line 19: should these be merged directly to main (as  well | instead)?

README.md
```diff
@@ -56,6 +58,22 @@ Now you can use `gopherjs build [package]`, `gopherjs build [files]` or `gopherj
...
+dyld[60166]: missing LC_UUID load command in ...
+```
+
+This is a Go + Mac issued [fixed in go1.24](https://tip.golang.org/doc/go1.24#linker).
+This is only an issue because gopherjs hasn't reached go1.24 support yet.
+
+If you get this error, build the `gopherjs` app with go1.24+. However, because of how
+we target specific STL versions, you must still use the correct STL for the version
+of gopherjs that you have. See how to set `GOPHERJS_GOROOT` above for how to select
+the correct STL that gopherjs will use.
```
flimzy, line 75: does this note also belong on main?

flimzy (approved):

README.md
grantnelson-wf: Yep, I noticed it was wrong while bumping the versions, but you're right, we should those in master. Same with the comment you called out below.

grantnelson-wf: These were added to https://github.com/gopherjs/gopherjs/pull/1460 that will put these changes into master

PR #1463, opened 2026-07-30
[go1.22] Fix reflectlite

grantnelson-wf:
The `reflectlite` package had some minor changes, mostly `typ` -> `typ_`. This change will unblock several other packages that have failures hidden by this change.

CI will not pass yet since this is part of the go1.22 update, however this will make the following pass
```bash
go install . && gopherjs test -v internal/reflectlite
```

related to #1334

### DIFF
--- a/compiler/natives/src/internal/reflectlite/value.go
+++ b/compiler/natives/src/internal/reflectlite/value.go
@@ -3,32 +3,35 @@
 package reflectlite
 
 import (
-	"unsafe"
-
 	"internal/abi"
+	"unsafe"
 
 	"github.com/gopherjs/gopherjs/js"
 )
 
+func (v Value) typ() *abi.Type {
+	return v.typ_
+}
+
 //gopherjs:new
 func (v Value) object() *js.Object {
-	if v.typ.Kind() == abi.Array || v.typ.Kind() == abi.Struct {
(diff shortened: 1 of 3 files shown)


compiler/natives/src/internal/reflectlite/value.go
```diff
@@ -3,32 +3,35 @@
...
 import (
-	"unsafe"
-
 	"internal/abi"
+	"unsafe"
 
 	"github.com/gopherjs/gopherjs/js"
 )
 
+func (v Value) typ() *abi.Type {
```
flimzy, line 12: Is  this func used anywhere? I see  only `typ` -> `typ_` changes inline? Was this an early considered approach that was abandoned, or something I'm not seeing?

grantnelson-wf: It's defined in the STL for go1.22 [reflectlite/value.go](https://cs.opensource.google/go/go/+/refs/tags/go1.22.12:src/internal/reflectlite/value.go;l=91) and used by some other methods that aren't overloaded. The original has some pointer casting that this overrides. (I probably should have put a `//gopherjs:replace` directive on that to help make it clearer. I also didn't look into the places where it changed to `typ_` to see if the natives changed to use `typ()`, which might have helped too.)

flimzy (approved):

